PR #27, opened 2024-09-11
WIP: reuse graphics devices over multiple tests

jkwak-work:
This change reuses the graphics devices so that we don't keep creating and destroying them for every test case.

This improves the runtime of slang-test by 40%. And it matches to the initial expectation because when profiled, _createDevice was taking 40% of the total execution time.

TODO: when the testing is done, the cached devices need to be explicitly released in TestContext::~TestContext().

### DIFF
--- a/src/render.cpp
+++ b/src/render.cpp
@@ -279,59 +279,430 @@ extern "C"
         return SLANG_OK;
     }
 
+    struct CachedIDevice
+    {
+        IDevice::Desc desc = {};
+        IDevice* device = nullptr;
+
+        AdapterLUID adapterLUID;
+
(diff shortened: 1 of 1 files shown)


skallweitNV:
Device caching is definitely a good idea for testing. I also use it for the local slang-rhi unit tests. However, this should be implemented in the slang test runner and **not** in the RHI. As an example how this is implemented for the local unit tests, see here: https://github.com/shader-slang/slang-rhi/blob/main/tests/testing.cpp#L344

jkwak-work:
Thank you for the feedback.
I will move my code to somewhere inside of slang-test application.

PR #40, opened 2024-09-17
Fix DX on VK

expipiplus1:
This could probably be a little neater, as at the moment it conflates dxvk with being not on windows, but I think that if anyone wants to use dxvk on Windows they can cope with improving this


skallweitNV:
so does this assume that dxvk is installed system wide? otherwise, where would we get the D3D headers from?

PR #93, opened 2024-11-01
Make sure WGPU buffers are created with initial data if available

aleino-nv:


src/rhi-shared.h
```diff
@@ -437,7 +437,7 @@ class SimpleShaderObjectData
 
     Index getCount() { return m_ordinaryData.size(); }
     void setCount(Index count) { m_ordinaryData.resize(count); }
-    uint8_t* getBuffer() { return m_ordinaryData.data(); }
+    uint8_t* getBuffer() { return m_ordinaryData.empty() ? nullptr : m_ordinaryData.data(); }
```
skallweitNV, line 440: probably not needed as we shouldn't access data if the buffer is empty .. but ok to keep too

src/wgpu/wgpu-shader-object.cpp
```diff
@@ -402,7 +402,8 @@ Result ShaderObjectImpl::_ensureOrdinaryDataBufferCreatedIfNeeded(
         bufferDesc.defaultState = ResourceState::ConstantBuffer;
         bufferDesc.memoryType = MemoryType::DeviceLocal;
         ComPtr<IBuffer> buffer;
-        SLANG_RETURN_ON_FAIL(encoder->m_device->createBuffer(bufferDesc, nullptr, buffer.writeRef()));
+        void * initData = m_data.getBuffer();
+        SLANG_RETURN_ON_FAIL(encoder->m_device->createBuffer(bufferDesc, initData, buffer.writeRef()));
```
skallweitNV:
```suggestion
        SLANG_RETURN_ON_FAIL(encoder->m_device->createBuffer(bufferDesc, m_constantBufferData.data(), buffer.writeRef()));
```

aleino-nv: Done except I check for empty as well.

skallweitNV: that's not necessary as we exit early if there is no constant data .. see at the start of the function
```c++
    if (m_constantBufferSize == 0)
    {
        return SLANG_OK;
    }
```

aleino-nv: Done.

PR #95, opened 2024-11-04
Propagate wsgl compilation errors

cheneym2:
wgpuDeviceCreateShaderModule reports
errors via a callback. Previously, the callback
only printed the error, but now, the error
can be read back and used to report SLANG_FAIL from createShaderModule.

Work on Issue 5291


skallweitNV:
I think the proper way to do this would be using error scopes. But I also just realized that I'm probably not doing good error handling in all of the WGPU backend. We can merge this to fix the current issue, will open a ticket to revisit that later and use error scopes on all resource creation calls.

PR #112, opened 2024-11-22
Remove specialization logic for interfaces in structured buffers.

csyonghe:
Delete some unnecessary specialization logic around handling of `StructuredBuffer<IFoo>`, since we no longer treat a shader parameter of `StructuredBuffer<IFoo>` type as a specializable parameter. Such parameters will always treat `IFoo` as a dynamically dispatched element instead of something specializable, because it is not very meaningful to specialize StructuredBuffer<IFoo> to StructuredBuffer<X>. In practice the StructuredBuffer<IFoo> usually represent a heterogeneous array.


skallweitNV (commented):
Is this specific to WGPU only or to other backends as well?

csyonghe:
It applies to all backends.

PR #117, opened 2024-12-19
Select vulkan device that supports most extensions.

csyonghe:
This makes it easy to run slang-test on devices with both an integrated GPU and a dedicated GPU.


skallweitNV:
We could also check for `deviceProperties.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU` to select a discrete GPU. Not sure what an ideal heuristic is for "selecting the best GPU".

PR #208, opened 2025-03-03
D3D11 support working

ccummingsNV:
D3D11 upload buffers now will use a 'STAGING' type for upload heaps in read/write mode if no other D3D11 usage flags are provided (i.e the user has said the buffer won't be used for any shader stages). 

Map will now selectively choose WRITE_DISCARD or READ_WRITE depending on whether buffer is dynamic or staging.

Staging heap specifies the buffers are copy only, so they can now safely use the heap like other targets.


skallweitNV (commented):
LGTM

can you rename `test-upload.cpp` to `test-staging-heap.cpp`? that would be the better name IMO

PR #211, opened 2025-03-04
Staging heap thread safe

ccummingsNV:
- Delete buffer pool
- Staging heap page size configurable
- Staging heap is thread safe

note: this adds thread safety to the staging  heap, however doesn't yet solve the problem of truly thread safe upload, as mapping/unmapping is still an issue. I'll address one after texture upload is done - instinct is it'll involve shifting the map/unmap/copy logic inside the heap, and platforms that don't need map/unmap will become naturally thread safe. Older platforms that do will then just wrap the copy in a mutex.


src/staging-heap.cpp
```diff
@@ -5,25 +5,28 @@
 
 namespace rhi {
 
-void StagingHeap::initialize(Device* device)
+void StagingHeap::initialize(Device* device, Size pageSize)
 {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
skallweitNV, line 12: probably not needed, `initialize` is called once only

src/staging-heap.cpp
```diff
@@ -5,25 +5,28 @@
...
     m_device = device;
+    m_pageSize = pageSize;
 }
 
 void StagingHeap::releaseAllFreePages()
```
skallweitNV, line 22: should this be a `public` function?

src/staging-heap.cpp
```diff
@@ -5,25 +5,28 @@
...
     }
-    for (auto page_id : pagesToRemove)
-        m_pages.erase(page_id);
+    for (auto page : pagesToRemove)
+        freePage(page);
 }
 
 void StagingHeap::release()
 {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
skallweitNV, line 36: again, not sure this needs a lock .. is `StagingHeap` supposed to be initialized/released from multiple threads?

src/staging-heap.h
```diff
@@ -149,8 +159,13 @@ class StagingHeap : public RefObject
     Size m_alignment = 1024;
     Size m_pageSize = 16 * 1024 * 1024;
     std::unordered_map<int, RefPtr<Page>> m_pages;
+    mutable std::mutex m_mutex;
 
     Result allocPage(size_t size, StagingHeap::Page** outPage);
+
+    void freePage(StagingHeap::Page* page);
+
+    void releaseAllFreePages();
```
skallweitNV, line 205: ah, looks like this is private now!

skallweitNV (approved):
LGTM

PR #217, opened 2025-03-05
Initial version of texture upload working for d3d12, webgpu and vulkan

ccummingsNV:


src/command-buffer.cpp
```diff
@@ -494,7 +494,7 @@ void CommandEncoder::copyTextureToBuffer(
     m_commandList->write(std::move(cmd));
 }
 
-void CommandEncoder::uploadTextureData(
+Result CommandEncoder::uploadTextureData(
```
skallweitNV, line 497: generally command encoder calls shouldn't have return values, but I guess for this call it makes sense to have one

src/command-list.h
```diff
@@ -438,11 +440,7 @@ class CommandList : public RefObject
         }
     }
 
-private:
-    ArenaAllocator& m_allocator;
-    std::set<RefPtr<RefObject>>& m_trackedObjects;
-    CommandSlot* m_commandSlots = nullptr;
-    CommandSlot* m_lastCommandSlot = nullptr;
+    const void* allocData(size_t size) { return m_allocator.allocate(size); }
```
skallweitNV: this should return `void*` not `const void*` .. users want to write into the allocated data

src/command-buffer.cpp
```diff
@@ -503,20 +503,98 @@ void CommandEncoder::uploadTextureData(
     uint32_t subresourceDataCount
 )
 {
+    // Allocate space in command list memory for layout information on each subresource.
+    SubresourceLayout* layouts =
+        (SubresourceLayout*)m_commandList->allocData(sizeof(SubresourceLayout) * subresourceDataCount);
```
skallweitNV: you're doing a const cast here .. maybe replace with `static_cast<SubresourceLayout*>(...)`

src/command-buffer.cpp
```diff
@@ -503,20 +503,98 @@ void CommandEncoder::uploadTextureData(
...
+    {
+        SubresourceLayout* srLayout = layouts;
+        for (uint32_t layerOffset = 0; layerOffset < subresourceRange.layerCount; layerOffset++)
+        {
+            uint32_t layerIndex = subresourceRange.baseArrayLayer + layerOffset;
+            for (uint32_t mipOffset = 0; mipOffset < subresourceRange.mipLevelCount; mipOffset++)
+            {
+                uint32_t mipLevel = subresourceRange.mipLevel + mipOffset;
+
+                ((Texture*)dst)->getSubresourceRegionLayout(mipLevel, layerIndex, offset, extent, srLayout);
```
skallweitNV: what I usually do is up in the function have
```c++
TextureImpl* textureImpl = checked_cast<TextureImpl*>(texture);
```
this will check the type in debug builds and just be a cast in release builds.

src/command-buffer.cpp
```diff
@@ -503,20 +503,98 @@ void CommandEncoder::uploadTextureData(
...
+        SubresourceData* srSrcData = subresourceData;
+        uint8_t* srDestData = dstData;
+        for (uint32_t layerOffset = 0; layerOffset < subresourceRange.layerCount; layerOffset++)
+        {
+            for (uint32_t mipOffset = 0; mipOffset < subresourceRange.mipLevelCount; mipOffset++)
+            {
+                // Source and dest rows may have different alignments, so its valid for strides to be
+                // different (even if data itself isn't). We copy the minimum of the two to avoid
+                // reading/writing out of bounds.
+                Size rowCopyStride = srSrcData->strideY < srLayout->strideY ? srSrcData->strideY : srLayout->strideY;
```
skallweitNV:
```suggestion
                Size rowCopyStride = min(srSrcData->strideY, srLayout->strideY);
```

src/command-buffer.cpp
```diff
@@ -503,20 +503,98 @@ void CommandEncoder::uploadTextureData(
...
+
+                // Iterate over slices and rows, copying a row at a time
+                uint8_t* sliceSrcData = (uint8_t*)srSrcData->data;
+                uint8_t* sliceDestData = srDestData;
+                for (uint32_t z = 0; z < srLayout->size.depth; z++)
```
skallweitNV: `z` -> `slice`
`y` -> `row`
?

src/d3d12/d3d12-command.cpp
```diff
@@ -402,7 +402,56 @@ void CommandRecorder::cmdClearTexture(const commands::ClearTexture& cmd)
...
+            footprint.Footprint.RowPitch = srLayout->strideY;
+
+            m_cmdList->CopyTextureRegion(&dstRegion, cmd.offset.x, cmd.offset.y, cmd.offset.z, &srcRegion, nullptr);
+
+            bufferOffset += srLayout->sizeInBytes;
+            srLayout++;
+        }
+    }
+
 #if 0
```
skallweitNV: should this be removed?

src/debug-layer/debug-command-encoder.cpp
```diff
@@ -352,7 +352,30 @@ void DebugCommandEncoder::uploadTextureData(
...
     requireOpen();
     requireNoPass();
-    baseObject->uploadTextureData(dst, subresourceRange, offset, extent, subresourceData, subresourceDataCount);
+
+    if (subresourceRange.mipLevelCount != 1)
+    {
+        if (offset.x != 0 || offset.y != 0 || offset.z != 0)
+        {
+            RHI_VALIDATION_ERROR("Uploading multiple mip levels at once requires offset to be 0");
+            return SLANG_FAIL;
```
skallweitNV: we should return `SLANG_E_INVALID_ARG` for these errors

src/staging-heap.h
```diff
@@ -112,13 +112,17 @@ class StagingHeap : public RefObject
...
 
+        const Allocation& getAllocation() const { return m_allocation; }
         Offset getOffset() const { return m_allocation.getOffset(); }
         Size getSize() const { return m_allocation.getSize(); }
         Page* getPage() const { return m_allocation.page; }
         int getPageId() const { return m_allocation.getPageId(); }
         Buffer* getBuffer() const { return m_allocation.getBuffer(); }
         const MetaData& getMetaData() const { return m_allocation.getMetaData(); }
 
+        void* map() { return m_heap->map(m_allocation); }
```
skallweitNV: should this return a `Result` or can it never fail?

include/slang-rhi.h
```diff
@@ -689,6 +689,24 @@ struct Extents
     int32_t height = 0;
     /// Depth (if 3d).
     int32_t depth = 0;
+
+    static Extents WholeTexture;
```
skallweitNV: other constants we define inline with `k` prefix, maybe we should change this?

include/slang-rhi.h
```diff
@@ -689,6 +689,24 @@ struct Extents
...
+};
+
+/// Layout of a single subresource in a texture. (see also SubresourceData)
```
skallweitNV, line 696: is this tightly packed or respecting the alignment requirements of the backend?

src/d3d12/d3d12-texture.cpp
```diff
@@ -80,12 +81,27 @@ Result TextureImpl::getSharedHandle(NativeHandle* outHandle)
 #endif
 }
 
-D3D12_CPU_DESCRIPTOR_HANDLE TextureImpl::getSRV(
-    Format format,
-    TextureType type,
-    TextureAspect aspect,
-    const SubresourceRange& range
+Result TextureImpl::getSubresourceRegionLayout(
```
skallweitNV: I wonder if this should be moved to the `Texture` base class .. all we probably need is a function to return the alignment for a given format.

src/rhi-shared.cpp
```diff
@@ -61,6 +67,50 @@ Result Buffer::getSharedHandle(NativeHandle* outHandle)
...
+    Offset3D offset,
+    Extents extents,
+    Size rowAlignment,
+    SubresourceLayout* outLayout
+)
+{
+    Extents textureSize = desc.size;
+    const FormatInfo& formatInfo = getFormatInfo(desc.format);
+
+    if (extents.width == kRemainingTextureSize)
```
skallweitNV, line 87: would it make sense to first compute the full extent of the texture, then check that offset is in range, and then shrink the extent? this will overflow if any(offset > extent)

src/d3d12/d3d12-texture.cpp
ccummingsNV: In revised version for next PR, this uses the getTextureRowAlignment device function (which I hadn't found previously), so is entirely implemented in the Texture base class.

include/slang-rhi.h
ccummingsNV: It would be required to respect the alignment requirements of the backend - this is the layout the subresource would need to be in memory. In next checkin it'll implicitly rely on getTextureRowAlignment

include/slang-rhi.h
ccummingsNV: Can do - I didn't because it's not technically const, but I'm not fussy

src/d3d12/d3d12-texture.cpp
ccummingsNV: (in fact, I'll add it for this PR)

src/rhi-shared.cpp
ccummingsNV: Yeah good shout.

PR #224, opened 2025-03-07
Read texture support for subresources

ccummingsNV:
- Added 2nd heap for read back
- Switched read texture to default to using copyTextureToBuffer
- Tests for texture layout
- Minimal tests for texture read
- WGPU still using bespoke readTexture - not worked out why staging heap won't map yet
- D3D11 will always need bespoke readTexture as it has no copy to buffer
- CUDA should work once copyTextureToBuffer works


CMakeLists.txt
```diff
@@ -651,6 +651,7 @@ if(SLANG_RHI_BUILD_TESTS)
         tests/test-clear-texture.cpp
         tests/test-cmd-clear-buffer.cpp
         tests/test-cmd-copy-buffer.cpp
+
```
skallweitNV: stray newline

include/slang-rhi.h
```diff
@@ -2588,8 +2592,21 @@ class IDevice : public ISlangUnknown
...
+        Size* outPixelSize = nullptr
+    ) = 0;
+
+    /// Helper overload that reads the entire texture (layer 0, mip level 0)
     virtual SLANG_NO_THROW Result SLANG_MCALL
-    readTexture(ITexture* texture, ISlangBlob** outBlob, Size* outRowPitch, Size* outPixelSize) = 0;
+    readTexture(ITexture* texture, ISlangBlob** outBlob, Size* outRowPitch, Size* outPixelSize = nullptr)
+    {
+        return readTexture(texture, 0, 0, outBlob, outRowPitch, outPixelSize);
+    };
```
skallweitNV: this should be an `inline` function:
```c++
inline Result readTexture(ITexture* texture, ISlangBlob** outBlob, Size* outRowPitch, Size* outPixelSize = nullptr)
```

src/cpu/cpu-device.h
```diff
@@ -56,9 +56,6 @@ class DeviceImpl : public Device
 
     virtual SLANG_NO_THROW Result SLANG_MCALL getQueue(QueueType type, ICommandQueue** outQueue) override;
 
-    virtual SLANG_NO_THROW Result SLANG_MCALL
```
skallweitNV, line 59: why is this removed? I think for CPU readback we will have a custom implementation, no?

src/cuda/cuda-device.h
```diff
@@ -132,6 +138,9 @@ class DeviceImpl : public Device
         override;
 
     void customizeShaderObject(ShaderObject* shaderObject) override;
+
+    // Provides a default implementation that returns SLANG_E_NOT_AVAILABLE.
```
skallweitNV: remove that comment

include/slang-rhi.h
ccummingsNV: I can see it shouldn't be virtual, but we shouldn't actively mark it as inline right? Isn't that the compiler's call these days?

skallweitNV: it's just a hint, compiler can ignore it. but we use inline for all other of these "helper" functions .. i agree it's not necessary though.

src/cpu/cpu-device.h
ccummingsNV: The default version calls copyTextureToBuffer, which will on CPU currently return that the function isn't implemented - i.e. result in the same behaviour. Having the extra override is redundant until we actually write a custom implementation

skallweitNV: sure, i'm just saying we probably will write a custom implementation so then we have to add it back :)

ccummingsNV: If we're going to do work to make the CPU version work, I'd rather go to the trouble of implementing 'copyTextureToBuffer' on the CPU side, then the readTexture will work without more platform specific implementations.

skallweitNV: fine with me .. it just feels silly do copy to staging buffers on the CPU :) but it's not perf critical, so having less code is probably better.

PR #225, opened 2025-03-07
DRAFT: Web gpu + metal read texture

ccummingsNV:


src/metal/metal-device.cpp
```diff
@@ -287,10 +289,13 @@ Result DeviceImpl::getTextureAllocationInfo(const TextureDesc& descIn, Size* out
     return SLANG_OK;
 }
 
-Result DeviceImpl::getTextureRowAlignment(Size* outAlignment)
+Result DeviceImpl::getTextureRowAlignment(Format format, Size* outAlignment)
 {
     AUTORELEASEPOOL
-    *outAlignment = 256;
+    if (format == Format::Undefined)
```
skallweitNV, line 293: should the `Format::Undefined` return the maximum of all supported formats?

ccummingsNV: If we can find one - I struggled to extract that info for metal.

ccummingsNV: I guess we could do that up front in device init - spin over all formats and get the max

skallweitNV: yep, that's what I was thinking too

PR #246, opened 2025-03-12
First version of texture test bed

ccummingsNV:
- New texture test bed that automatically tests lots of variations etc
- Limitted to sensible set of formats, with a lot of 'TODO' exclusions atm due to missing platform functionality
- Added format support checks for platforms missing it
- Test currently has special name to separate it from the existing texture-create tests due to be deleted


include/slang-rhi.h
```diff
@@ -732,6 +739,9 @@ struct SubresourceLayout
 
     /// Overall size required to fit the subresource data (typically size.z*strideZ).
     Size sizeInBytes;
+
+    /// Number of rows (will match size.height for uncompressed formats).
+    Size rowCount;
```
tomas-davidovic, line 744: What is it gonna be for compressed formats?

src/cuda/cuda-device.cpp
```diff
@@ -397,6 +397,48 @@ Result DeviceImpl::getCUDAFormat(Format format, CUarray_format* outFormat)
...
+    case Format::R16_SINT:
+    case Format::R8G8B8A8_SINT:
+    case Format::R8G8_SINT:
+    case Format::R8_SINT:
+        break;
+    default:
+        *outFormatSupport = (FormatSupport)0;
+        break;
+    }
+    return SLANG_OK;
```
tomas-davidovic, line 439: Looking at this, I don't understand why do you override the `Device::getFormatSupport` result for some of the formats. Maybe add a comment?

src/cuda/cuda-device.cpp
```diff
@@ -548,8 +590,10 @@ Result DeviceImpl::readTexture(
     copyParam.Depth = mipSize.depth;
     SLANG_CUDA_RETURN_ON_FAIL(cuMemcpy3D(&copyParam));
 
-    *outRowPitch = rowPitch;
-    *outPixelSize = pixelSize;
+    if (outRowPitch)
+        *outRowPitch = rowPitch;
```
tomas-davidovic, line 594: Here you have one line `then` statement without braces, while in `if (!_getFormatInfo(format))` (in the same PR) you have it with braces. Just flagging in case of guidelines ruling one way or another, if they leave it open, then it is totally fine.

include/slang-rhi.h
```diff
@@ -2113,6 +2123,8 @@ class ICommandQueue : public ISlangUnknown
 
     inline Result submit(ICommandBuffer* commandBuffer)
     {
+        if (!commandBuffer)
```
skallweitNV, line 2126: that's probably a condition for the validation layer, we don't generally check the inputs in other places

src/d3d11/d3d11-device.cpp
```diff
@@ -368,6 +368,7 @@ Result DeviceImpl::readTexture(
         *outPixelSize = bytesPerPixel;
 
     D3D11_TEXTURE2D_DESC textureDesc;
```
skallweitNV, line 370: I mostly just do zero initialization using `D3D11_TEXTURE2D_DESC textureDesc = {};`

src/device.h
```diff
@@ -211,7 +211,7 @@ class Device : public IDevice, public ComObject
         uint32_t mipLevel,
         ISlangBlob** outBlob,
         Size* outRowPitch,
-        Size* outPixelSize
+        Size* outPixelSize = nullptr
```
tomas-davidovic, line 214: Is there a reason why is `outPixelSize` optional but `outRowPitch` is not?

tomas-davidovic (approved):
Just couple small things, mostly for comments or clarifications.

src/d3d12/d3d12-command.cpp
```diff
@@ -292,43 +294,62 @@ void CommandRecorder::cmdCopyTextureToBuffer(const commands::CopyTextureToBuffer
...
             footprint.Footprint.Depth = extent.depth;
         }
         else
         {
-            footprint.Footprint.Depth = max(1, (textureSize.depth >> mipLevel)) - srcOffset.z;
+            footprint.Footprint.Depth = mipSize.depth - srcOffset.z;
         }
 
+        footprint.Footprint.Width =
+            (footprint.Footprint.Width + formatInfo.blockWidth - 1) & ~(formatInfo.blockWidth - 1);
```
skallweitNV, line 327: we could add some `divRoundUp(x, alignment)` if this shows up a lot

src/device.h
ccummingsNV: Goal is to remove outPixelSize, but as a first step making it optional so we don't need to keep writing new code that will break when its removed

include/slang-rhi.h
ccummingsNV: height/blockHeight  (aligned). will add comment

src/device.h
tomas-davidovic: Makes total sense.. maaaaaybe add `// deprecated` or something, so others (me) don't fall into that pitfall? :-D

src/cuda/cuda-device.cpp
ccummingsNV: Good point. I'll fix this in next PR as I've had to do similar for D3D11

tests/texture-test.cpp
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+}
+
+Result TextureData::createTexture(IDevice* device, ITexture** texture) const
+{
+    return device->createTexture(desc, subresourceData.data(), texture);
+}
+
+void TextureData::checkEqual(ComPtr<ITexture> texture) const
```
skallweitNV, line 136: You could just take `ITexture*` here

tests/texture-test.h
```diff
@@ -0,0 +1,424 @@
...
+
+/// Description of a given variant to test.
+struct TextureTestVariant
+{
+    std::vector<TestTextureDesc> descriptors;
+};
+
+enum class TTShape
+{
+    D1 = 1 << 0,
```
skallweitNV, line 67: if find `D1` etc. a little confusing .. alternatives might be `Shape1D` or `Texture1D` ..

skallweitNV: it could also be `TTDimension` ..

src/cuda/cuda-device.cpp
skallweitNV: most of the code uses braces for if blocks, I try to stick with that as much as possible

src/d3d12/d3d12-command.cpp
skallweitNV: well, that name makes no sense .. maybe `alignTo(value, alignment)` ?

tomas-davidovic: This only works for power-of-2 alignments, doesn't it?

skallweitNV: yep, but block sizes should always be power of 2 ..

tests/texture-test.h
ccummingsNV: Aim is to get it less wordy for writing tests (so it all fits on 1 line), and avoiding redundancy. i.e. TTShape::Shape1D is pointless. Ideally it'd be TTShape (or TTDimension) :: 1D but c++ doesn't like it. I'll try and think of something better for next PR though

src/d3d12/d3d12-command.cpp
ccummingsNV: Yeah I'm going to add a helper for this - it already exists in several places

tomas-davidovic: Sorry, my point was that a generic `alignTo` should align to other alignments than just power of 2.
It is more of an input for the helpers name/function than argument about correctness of this specific code.

include/slang-rhi.h
ccummingsNV: Cool - I'll move to debug layer

src/d3d12/d3d12-command.cpp
ccummingsNV: heheh - already exists, but most of code doesn't use it: calcAligned

PR #249, opened 2025-03-12
Fixes for new texture tests

ccummingsNV:
- Add/use helpers for alignment calculations across the code
- Route slang includes through a single file so SLANG_HANDLE_RESULT_FAIL can be overriden
- Fix WebGPU compressed textures


tests/texture-test.cpp
```diff
@@ -80,15 +80,24 @@ bool getMultisampleType(TextureType type, TextureType& outArrayType)
 // TextureData
 //----------------------------------------------------------
 
-void TextureData::init(const TextureDesc& _desc, TextureInitMode _initMode, int _initSeed)
+void TextureData::init(IDevice* _device, const TextureDesc& _desc, TextureInitMode _initMode, int _initSeed)
```
skallweitNV, line 83: `_` prefix is reserved, so I usually do `device_` instead

ccummingsNV: Yeah I'll probably switch TextureData to be a class next anyway - the need for the _ is only because I've got something as a struct that should really be a fully fledged class.

PR #251, opened 2025-03-12
Read texture cpu and d3d11

ccummingsNV:
Fixed up readTexture implementations for cpu + d3d11, and fixed issue with cuda:
- CPU is basically just a better memcpy
- D3D11 now supports creation of textures in staging mode, and uses this to create a single layer staging texture to copy from of the correct size for the subresource
- Removed 3-component formats from CUDA as cuArrayCreate requires 1/2/4 channels.

### DIFF
--- a/src/d3d12/d3d12-command.cpp
+++ b/src/d3d12/d3d12-command.cpp
@@ -270,7 +270,8 @@ void CommandRecorder::cmdCopyTextureToBuffer(const commands::CopyTextureToBuffer
     if (srcSubresource.mipLevelCount == 0)
         srcSubresource.mipLevelCount = src->m_desc.mipLevelCount;
     if (srcSubresource.layerCount == 0)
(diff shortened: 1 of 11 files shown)


src/d3d12/d3d12-command.cpp
```diff
@@ -270,7 +270,8 @@ void CommandRecorder::cmdCopyTextureToBuffer(const commands::CopyTextureToBuffer
     if (srcSubresource.mipLevelCount == 0)
         srcSubresource.mipLevelCount = src->m_desc.mipLevelCount;
     if (srcSubresource.layerCount == 0)
-        srcSubresource.layerCount = src->m_desc.arrayLength;
+        srcSubresource.layerCount = src->m_desc.arrayLength; // TODO: This is wrong - should we be allowing
```
skallweitNV, line 273: Good point, I think this is from the past. We need to document the API and then revisit the implementations.

PR #253, opened 2025-03-13
Clear texture test

skallweitNV:
- add clear texture tests (float/uint for d3d11,d3d12 and vulkan) using new texture test framework
- various fixes in d3d11/d3d12

### DIFF
--- a/tests/texture-test.cpp
+++ b/tests/texture-test.cpp
@@ -98,7 +98,7 @@ void TextureData::init(IDevice* device_, const TextureDesc& desc_, TextureInitMo
 
     REQUIRE(is_set(formatSupport, FormatSupport::Texture));
 
-    desc.usage = TextureUsage::CopySource | TextureUsage::CopyDestination;
+    desc.usage |= TextureUsage::CopySource | TextureUsage::CopyDestination;
 
     // D3D12 needs multisampled textures to be render targets.
     if (isMultisamplingType(desc_.type))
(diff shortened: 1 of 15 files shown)


tests/texture-test.h
```diff
@@ -43,7 +45,7 @@ struct TextureData
     std::vector<Subresource> subresources;
     std::vector<SubresourceData> subresourceData;
 
-    void init(IDevice* device, const TextureDesc& desc, TextureInitMode initMode, int initSeed = 0);
+    void init(const TextureTestOptions& options, const TextureDesc& desc, TextureInitMode initMode, int initSeed = 0);
```
ccummingsNV: Not sure this is valid for some cases - the TextureData is designed for you to be able to create new ones, in addition to relying on the testing system to do it for you. This means that can only be done if the texture test options have been plumbed through. Maybe make it an overload?

tests/texture-test.h
```diff
@@ -165,6 +176,11 @@ class TextureTestOptions
         addGenerator([this](int state, TextureTestVariant variant) { filterFormat(state, variant); });
     }
 
+    /// Set additional texture usage flags.
```
ccummingsNV: As discussed, if it were me this would be an additional scalar option in the variant generation.

tests/texture-test.cpp
```diff
@@ -98,6 +104,7 @@ void TextureData::init(IDevice* _device, const TextureDesc& _desc, TextureInitMo
     REQUIRE(is_set(formatSupport, FormatSupport::Texture));
 
     desc.usage = TextureUsage::CopySource | TextureUsage::CopyDestination;
```
ccummingsNV: This should probably be tweaked so the usage isn't initialized here, but just added to (even for the copy source/dest stuff). Even in the absence of your new usage stuff, you can specify up front a 'base texture descriptor' when generating variants, which could have default usage flags that we should add to rather than replace.

ccummingsNV (changes requested):
Mostly fine - the only thing I'd definitely say wasn't right is the switch to initializing texture data with test options. Fine to have it as an option, but it shouldn't be a requirement.

PR #260, opened 2025-03-14
CUDA clear texture support

skallweitNV:
- add support for clearing textures in CUDA
- implemented a utility class `ClearEngine` that has a set of kernels to clear various surface dimensions/bit widths
- enable all clear tests for CUDA
- fix texture upload and enable all tests


ccummingsNV (approved):
LGTM

PR #263, opened 2025-03-17
metal texture upload

skallweitNV:
- add support for metal texture upload (including compressed formats)


ccummingsNV (approved):
Suggest sorting out alignment issue and testing with BC textures as I can't tell whether the sizing calculations are correct without knowing more about the api. Otherwise looks good though.

PR #266, opened 2025-03-18
Test none-power-of-2 textures

ccummingsNV:
Add support for none-power-of-2 tests, along with format info saying whether a format supports it, and some test fixes to handle alignments.


include/slang-rhi.h
```diff
@@ -294,6 +294,7 @@ struct FormatInfo
     bool isSigned : 1;
     bool isSrgb : 1;
     bool isCompressed : 1;
+    bool supportsNonePowerOf2 : 1;
```
skallweitNV: shouldn't that be spelled `supportsNonPowerOf2` ? In addition I would probably prefer `supportsNonPowerOfTwo`

tests/test-cmd-copy-texture.cpp
```diff
@@ -13,7 +13,13 @@ using namespace rhi::testing;
 GPU_TEST_CASE("cmd-copy-texture-full", D3D12 | Vulkan | WGPU)
 {
     TextureTestOptions options(device);
-    options.addVariants(TTShape::All, TTArray::Both, TTMip::Both, TTMS::Both);
+    options.addVariants(
+        TTShape::All,    // all shapes
+        TTArray::Both,   // array and none-array
```
skallweitNV: `non-array`

_none_ means "nothing" or "zero" which I don't think is correct here .. I think I've seen this non/none swap in other places before, maybe worth a quick text search.

skallweitNV (commented):
LGTM other than the non/none spelling

PR #267, opened 2025-03-18
Fix vulkan validation

westlicht:
- disable debug printf unless specifically enabled using the new `VulkanDeviceExtendedDesc` (this gets rid of the annoying validation message)
- fix semaphore wait stages in `submit`
- fix surface <-> submit synchronization
- fix initial swapchain image barrier

### DIFF
--- a/src/d3d12/d3d12-device.cpp
+++ b/src/d3d12/d3d12-device.cpp
@@ -367,7 +367,7 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         switch (header->type)
         {
         case StructType::D3D12DeviceExtendedDesc:
-            memcpy((void*)&m_extendedDesc, header, sizeof(m_extendedDesc));
+            memcpy(static_cast<void*>(&m_extendedDesc), header, sizeof(m_extendedDesc));
             break;
         case StructType::D3D12ExperimentalFeaturesDesc:
             processExperimentalFeaturesDesc(d3dModule, header);
--- a/src/d3d12/d3d12-device.cpp
+++ b/src/d3d12/d3d12-device.cpp
@@ -367,7 +367,7 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         switch (header->type)
         {
         case StructType::D3D12DeviceExtendedDesc:
-            memcpy(&m_extendedDesc, header, sizeof(m_extendedDesc));
+            memcpy((void*)&m_extendedDesc, header, sizeof(m_extendedDesc));
             break;
         case StructType::D3D12ExperimentalFeaturesDesc:
             processExperimentalFeaturesDesc(d3dModule, header);
--- a/src/metal/metal-texture.cpp
+++ b/src/metal/metal-texture.cpp
@@ -106,8 +106,6 @@ Result DeviceImpl::createTexture(const TextureDesc& desc_, const SubresourceData
         {
         case Format::R32Uint:
         case Format::R32Sint:
-            // case Format::RG32Uint:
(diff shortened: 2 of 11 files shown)


src/d3d12/d3d12-device.cpp
```diff
@@ -367,7 +367,7 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         switch (header->type)
         {
         case StructType::D3D12DeviceExtendedDesc:
-            memcpy(&m_extendedDesc, header, sizeof(m_extendedDesc));
+            memcpy((void*)&m_extendedDesc, header, sizeof(m_extendedDesc));
```
ccummingsNV: Is this here because it's a const cast? Maybe use a const_cast if so?

src/metal/metal-texture.cpp
```diff
@@ -106,8 +106,8 @@ Result DeviceImpl::createTexture(const TextureDesc& desc_, const SubresourceData
         {
         case Format::R32Uint:
         case Format::R32Sint:
-        // case Format::RG32Uint:
-        // case Format::RG32Sint:
+            // case Format::RG32Uint:
+            // case Format::RG32Sint:
             textureUsage |= MTL::TextureUsageShaderAtomic;
```
ccummingsNV, line 109: Shall we just delete these comments?

src/vulkan/vk-device.cpp
```diff
@@ -1074,6 +1077,19 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
 
     m_desc = desc;
 
+    // Process chained descs
+    for (DescStructHeader* header = static_cast<DescStructHeader*>(desc.next); header; header = header->next)
+    {
+        switch (header->type)
+        {
+        case StructType::VulkanDeviceExtendedDesc:
+            memcpy((void*)&m_extendedDesc, header, sizeof(m_extendedDesc));
```
ccummingsNV: Another const cast?

src/vulkan/vk-surface.cpp
```diff
@@ -229,6 +254,13 @@ void SurfaceImpl::destroySwapchain()
     auto& api = m_device->m_api;
     api.vkQueueWaitIdle(m_device->m_queue->m_queue);
     m_textures.clear();
+    for (FrameData& frameData : m_frameData)
```
ccummingsNV, line 249: You check if swapchain is null, but not the fences. Should they have checks too?

ccummingsNV (approved):
A few comments worth you double checking, but otherwise looks good.

src/d3d12/d3d12-device.cpp
skallweitNV: No, it's here because of a gcc warning (that might be incorrect) where it complains because the ExtendedDesc has a a constructor (because it has default values). I could actually remove it here because at least for now we're not compiling D3D12 on gcc, but it was necessary in the vulkan device code.

src/metal/metal-texture.cpp
skallweitNV: we will need to properly set those flags depending on some feature tables in the future .. I can remove the commented out once if you prefer.

ccummingsNV: Your call - just highlighting it

src/d3d12/d3d12-device.cpp
skallweitNV: replaced with `static_cast` now

src/metal/metal-texture.cpp
skallweitNV: removed

PR #268, opened 2025-03-18
Surface improvements

westlicht:
- rename `getCurrentTexture` to `acquireNextImage` which is what that functions does
- allow presenting an image that has not been rendered to
- add validation for the `acquireNextImage` -> `present` dependency

### DIFF
--- a/src/d3d11/d3d11-surface.h
+++ b/src/d3d11/d3d11-surface.h
@@ -18,8 +18,6 @@ public:
     virtual IUnknown* getOwningDevice() override { return m_d3dDevice; }
 
     virtual SLANG_NO_THROW Result SLANG_MCALL configure(const SurfaceConfig& config) override;
-    // virtual SLANG_NO_THROW Result SLANG_MCALL acquireNextImage(ITexture** outTexture) override;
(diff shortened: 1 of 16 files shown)


src/d3d11/d3d11-surface.h
```diff
@@ -18,7 +18,7 @@ class SurfaceImpl : public D3DSurface
     virtual IUnknown* getOwningDevice() override { return m_d3dDevice; }
 
     virtual SLANG_NO_THROW Result SLANG_MCALL configure(const SurfaceConfig& config) override;
-    // virtual SLANG_NO_THROW Result SLANG_MCALL getCurrentTexture(ITexture** outTexture) override;
+    // virtual SLANG_NO_THROW Result SLANG_MCALL acquireNextImage(ITexture** outTexture) override;
```
ccummingsNV: Just remove these rather than leave them commented out?

ccummingsNV (approved):
Looks good (I ignored changes that were from the previous unmerged PR)

src/d3d11/d3d11-surface.h
skallweitNV: done

PR #271, opened 2025-03-19
Remove test noise

ccummingsNV:
- Shifted a load of 'MESSAGE" calls to 'INFO'
- Removed logging of skipped tests
- Removed logging of skipped formats
- Hide Vulkan warning about coop vec structure

### DIFF
--- a/tests/test-benchmark-command.cpp
+++ b/tests/test-benchmark-command.cpp
@@ -89,7 +89,7 @@ GPU_TEST_CASE("benchmark-command", ALL)
 
     end = std::chrono::high_resolution_clock::now();
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
-    INFO("Duration: " << duration << " ms");
+    fprintf(stderr, ": Duration: %lld ms)", duration);
 
     queue->waitOnHost();
 
--- a/tests/test-benchmark-command.cpp
+++ b/tests/test-benchmark-command.cpp
@@ -89,7 +89,7 @@ GPU_TEST_CASE("benchmark-command", ALL)
 
     end = std::chrono::high_resolution_clock::now();
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
-    MESSAGE("Duration: " << duration << " ms");
+    INFO("Duration: " << duration << " ms");
 
     queue->waitOnHost();
 
--- a/tests/testing.cpp
+++ b/tests/testing.cpp
@@ -92,7 +92,28 @@ public:
     {
         if (!doctest::is_running_in_test)
             return;
-        MESSAGE(doctest::String(message));
+        if (type == DebugMessageType::Info)
+        {
+            INFO(doctest::String(message));
+        }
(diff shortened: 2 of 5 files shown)


tests/testing.cpp
```diff
@@ -92,7 +92,28 @@ class DebugCallback : public IDebugCallback
...
         if (!doctest::is_running_in_test)
             return;
-        MESSAGE(doctest::String(message));
+        if (type == DebugMessageType::Info)
+        {
+            INFO(doctest::String(message));
+        }
+        else if (type == DebugMessageType::Warning)
+        {
+            INFO(doctest::String(message));
```
westlicht: do we only want to see these when a test fails? i'd argue we want to always see validation warnings and try to fix them.

tests/test-benchmark-command.cpp
```diff
@@ -89,7 +89,7 @@ GPU_TEST_CASE("benchmark-command", ALL)
 
     end = std::chrono::high_resolution_clock::now();
     auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
-    MESSAGE("Duration: " << duration << " ms");
+    INFO("Duration: " << duration << " ms");
```
westlicht: i'd keep this a message .. although we should eventually move to some proper performance testing.
or is there an easy option in doctest to enable `INFO` logging without logging all the asserts too?

tests/testing.h
```diff
@@ -207,7 +207,6 @@ enum TestFlags
 #define SKIP(msg)                                                                                                      \
```
westlicht: `msg` doesn't make much sense now

tests/test-benchmark-command.cpp
ccummingsNV: I'd argue that unless you're looking at them, it's noise :) Do you check them regularly? Maybe I capture them and log them at the end so they don't clog up the test results?

tests/testing.cpp
ccummingsNV: I'll check out when warnings come through and treat as message or failure if it's not too crazy.

tests/test-benchmark-command.cpp
westlicht: i've looked at them a lot when optimizing command submission

tests/testing.h
ccummingsNV: I might leave it in for now - forcing people to put a reason into the code for why something is skipped doesn't hurt, even if we don't actively remove it.

tests/test-benchmark-command.cpp
ccummingsNV: Ok - found a way to just use fprintf to sneakily put them on the same line as the test name so they don't spam the log

tests/testing.cpp
ccummingsNV: Had to filter a couple of extra things, but warnings will now get treated as an error too.

tests/testing.h
ccummingsNV: changed arg name so it doesn't imply it'll actually log something though

PR #273, opened 2025-03-20
Sampler improvements

skallweitNV:
- set default border color to transparent black
- add `custom-border-color` device feature (d3d11, d3d12, vulkan)
- add basic sampler tests
- add basic sampler validation
- add support for predefined and custom border color in vulkan
- add support for predefined border color in metal


ccummingsNV (approved):
Looks good

PR #275, opened 2025-03-21
Copy texture to buffer

ccummingsNV:
- Implemented copy texture to buffer for d3d12, vulkan, wgpu and cuda
- Adjusted some of the testing api to avoid confusion over which parameters were src vs dst


include/slang-rhi.h
```diff
@@ -2021,7 +2023,8 @@ class ICommandEncoder : public ISlangUnknown
         Size dstSize,
         Size dstRowStride,
         ITexture* src,
-        SubresourceRange srcSubresource,
+        uint32_t layerIndex,
+        uint32_t mipLevel,
```
tdavidovicNV, line 2027: Isn't this removing the texture arrayness that was in the SubresourceRange?

tdavidovicNV (commented):
I think this is fine, but the change from `SubresourceRange` to `mipLevel` and `layerIndex` could use some comment in the PR, I think.
At a first glance it feels like the API now allows fewer things than it used to.

include/slang-rhi.h
ccummingsNV: Not quite - this was a concious decision by Simon and I to make it that the buffer copies only did 1 subresource at a time - most of the existing code did that anyway, it was just undocumented or checked at runtime. Arrays still work the same. Though the SubresourceRange naming should be updated if it refers to arrays, as it isn't an array index - it's a layer index, which isn't the same thing for cube maps.

PR #279, opened 2025-03-26
Output layout from readTexture

ccummingsNV:
Increasing number of places that used readTexture then proceeded to attempt to calculate mip sizes, explicitly get layout or look at formats to figure out extra info about the data it reads. Standardized this by just letting readTexture return the full layout it used for the read back.


tdavidovicNV (approved):
LGTM

PR #281, opened 2025-03-26
fix(d3d11): use correct cbvCount when binding constant buffers

szihs:
The D3D11 backend was incorrectly using uavCount instead of cbvCount when binding constant buffers in VSSetConstantBuffers1. This caused rendering tests to fail as constant buffers weren't properly provided to the pipeline.

Fixes shader-slang/slang#6531

#262 Already has the fix !. So either we update the submodule for Top of the tree shader-slang/slang or pick this.


skallweitNV:
this has long been fixed, we need to update `slang-rhi` in the slang testing infrastructure instead

PR #299, opened 2025-04-07
Add reporting for D3D11 double support

aidanfnv:
For https://github.com/shader-slang/slang/issues/6171

This adds logic for reporting double support to the d3d11 backend, to allow for checking for that support when running slang-test tests on GPUs that do not support D3D11_FEATURE_DOUBLES.


jkwak-work (approved):
Looks good to me

Issue #220, opened 2025-03-07
Texture3D WSize not properly set in TextureImpl::getUAV


Hyiker:
## Problem

I'm having a demand creating Texture3D with mipmap, which is quite rare. When I tried to create a UAV from a 256x256x256 texture3d with mipCount = 7, dx12 debug layer reported:

> D3D12 ERROR: ID3D12Device::CreateUnorderedAccessView: The Dimensions of the View are invalid. MipSlice (value = 2) must be between 0 and MipLevels-1 of the Texture Resource, 7, inclusively. FirstWSlice (value = 0) must be between 0 and the Depth size of the Mip Level, 63, inclusively. With the current FirstWSlice, WSize (value = 256) must be between 1 and 64, or -1 to default to all slices from MipSlice, inclusively, in order that the View fit on the Texture. [ STATE_CREATION ERROR #345: CREATEUNORDEREDACCESSVIEW_INVALIDDIMENSIONS]

## Cause

I was running my algorithm on [Falcor](https://github.com/NVIDIAGameWorks/Falcor), with slang version 2023.5.8, keyword search showed that WSize is set in slang-rhi/src/d3d12/d3d12-texture.cpp:

```cpp
case TextureType::Texture3D:
        viewDesc.ViewDimension = D3D12_UAV_DIMENSION_TEXTURE3D;
        viewDesc.Texture3D.MipSlice = range.mipLevel;
        viewDesc.Texture3D.FirstWSlice = range.baseArrayLayer;
        viewDesc.Texture3D.WSize = m_desc.size.depth;
        break;
```

where WSize is set to texture depth, as a constant in this function(256 in my case), which should be the depth corresponding to the texture mipmap level(64, in my case). In the latest version of slang-rhi, [src/d3d12/d3d12-texture.cpp](https://github.com/shader-slang/slang-rhi/blob/ee449ba6a400aa12b23e960b93e8abc6c80bff4f/src/d3d12/d3d12-texture.cpp#L233) hasn't changed yet.

## Possible solution

[Microsoft document](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_tex3d_uav) indicates that:

> WSize
> The number of depth slices.
> Set to -1 to indicate all the depth slices from FirstWSlice to the last slice.

Otherwise, API should calculate the WSize for the specific mipLevel or provides an option for user to manually set this parameter. Anyway, I prefer the -1 solution.

[...]

PR #300, opened 2025-04-08
Fix issues with d3d12 textureViews (#220)

sricker-nvidia:
Change makes a fix to how 3D texture UAVs and RTVs were handled that resulted in d3d12 errors.

Change also adds a new unit test to exercise the problematic pathway. This unit test was used to repro the issue in #220 and also used to verify the fixes in this change.

The change also makes some additional small fix ups discovered while investigating the original issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -730,6 +730,7 @@ if(SLANG_RHI_BUILD_TESTS)
         tests/test-texture-create.cpp
         tests/test-texture-layout.cpp
         tests/test-texture-shared.cpp
+        tests/test-texture-view.cpp
         tests/test-texture-types.cpp
         tests/test-uint16-structured-buffer.cpp
         tests/testing.cpp
--- a/cmake/FetchPackage.cmake
+++ b/cmake/FetchPackage.cmake
@@ -9,6 +9,7 @@ macro(FetchPackage name)
     FetchContent_Declare(
         ${name}
         URL ${FETCH_URL}
+        HTTP_HEADER "Authorization: token ${SLANG_GITHUB_TOKEN}"
     )
     FetchContent_GetProperties(${name})
     if(NOT ${name}_POPULATED)
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -656,7 +656,7 @@ struct SubresourceRange
 
     // TODO: Check this comment - many areas explicitly specify a 3D offset / extents,
     // and this is expected to be 0 for 3D texture.
-    uint32_t baseArrayLayer; // For Texture3D, this is WSlice.
+    uint32_t baseArrayLayer;
 
     uint32_t layerCount; // For cube maps, this is a multiple of 6.
 
@@ -686,7 +686,7 @@ static const size_t kDefaultAlignment = 0xffffffff;
 /// "layers" of texels.
 ///
 /// For a texture with multiple mip levels or array elements,
-/// each mip level and array element is stores as a distinct
+/// each mip level and array element is stored as a distinct
 /// subresource. When indexing into an array of subresources,
 /// the index of a subresoruce for mip level `m` and array
 /// index `a` is `m + a*mipLevelCount`.
--- a/src/d3d12/d3d12-texture.cpp
+++ b/src/d3d12/d3d12-texture.cpp
@@ -215,8 +215,11 @@ D3D12_CPU_DESCRIPTOR_HANDLE TextureImpl::getUAV(
     case TextureType::Texture3D:
         viewDesc.ViewDimension = D3D12_UAV_DIMENSION_TEXTURE3D;
         viewDesc.Texture3D.MipSlice = range.mipLevel;
-        viewDesc.Texture3D.FirstWSlice = range.baseArrayLayer;
-        viewDesc.Texture3D.WSize = max(m_desc.size.depth >> range.mipLevel, 1);
+        // Set these to 0 and -1 for now to select all depth
+        // slices by default, as selecting a subset of depth slices
+        // is a concept currently only supported by d3d12.
+        viewDesc.Texture3D.FirstWSlice = 0;
+        viewDesc.Texture3D.WSize = -1;
         break;
     }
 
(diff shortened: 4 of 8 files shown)


jkwak-work:
CI are failing.

CMakeLists.txt
```diff
@@ -727,6 +727,7 @@ if(SLANG_RHI_BUILD_TESTS)
         tests/test-texture-create.cpp
         tests/test-texture-layout.cpp
         tests/test-texture-shared.cpp
+        tests/test-texture-textureView.cpp
```
skallweitNV: Rename this to `test-texture-view.cpp`. We use all lower-case file names.

cmake/FetchPackage.cmake
```diff
@@ -9,6 +9,7 @@ macro(FetchPackage name)
     FetchContent_Declare(
         ${name}
         URL ${FETCH_URL}
+        HTTP_HEADER "Authorization: token ${SLANG_GITHUB_TOKEN}"
```
skallweitNV, line 12: Is this necessary?

tests/textureView-test-shaders.slang
```diff
@@ -0,0 +1,25 @@
+// textureView-test-shaders.slang - Shaders used by the unit tests in test-texture-textureView.cpp
```
skallweitNV: Rename to `test-texture-view.slang` to match with the test cpp unit.

include/slang-rhi.h
```diff
@@ -656,19 +656,24 @@ struct SubresourceRange
 
     // TODO: Check this comment - many areas explicitly specify a 3D offset / extents,
     // and this is expected to be 0 for 3D texture.
-    uint32_t baseArrayLayer; // For Texture3D, this is WSlice.
+    uint32_t baseArrayLayer;
 
     uint32_t layerCount; // For cube maps, this is a multiple of 6.
 
+    // Only relevant to 3D textures.
+    uint32_t firstSlice; // For 3D textures, this is the first slice index.
```
skallweitNV: I don't like introducing these two fields just for 3D textures. If we really want to select slices (not sure we need that in the first place) we should probably just repurpose `baseArrayLayer` and `layerCount`, because there is no `Texture3DArray` resource type.

src/d3d12/d3d12-texture.cpp
```diff
@@ -215,8 +215,8 @@ D3D12_CPU_DESCRIPTOR_HANDLE TextureImpl::getUAV(
         case TextureType::Texture3D:
         viewDesc.ViewDimension = D3D12_UAV_DIMENSION_TEXTURE3D;
         viewDesc.Texture3D.MipSlice = range.mipLevel;
-        viewDesc.Texture3D.FirstWSlice = range.baseArrayLayer;
-        viewDesc.Texture3D.WSize = max(m_desc.size.depth >> range.mipLevel, 1);
+        viewDesc.Texture3D.FirstWSlice = range.firstSlice;
```
skallweitNV: The easiest solution here would be to just set `FirstWSlice` to `0` and `WSize` to `-1` to always select all depth slices.

CMakeLists.txt
sricker-nvidia: Sure, but is this worth keeping around if we don't want this change? It's specifically testing selectable depth slices.

cmake/FetchPackage.cmake
sricker-nvidia: No, I can break this out into a separate change. This was to match https://github.com/shader-slang/slang/blob/master/cmake/GitHubRelease.cmake#L154 and avoid github rest api rate limiting when using cmake to setup and build the project.

tests/textureView-test-shaders.slang
sricker-nvidia: ack

include/slang-rhi.h
sricker-nvidia: I initially attempted to use baseArrayLayer and layerCount, but it proved hard to implement correctly. I found that to use them, I needed to edit how they were updated in Texture::resolveSubresourceRange, but this broke a bunch of tests for Vulkan because the texture view would have a layerCount greater than the count that the texture was initially created with. I can't change the texture's baseArrayLayer and layerCount becauase that breaks CommandEncoder::uploadTextureData. I might have been able to make a WAR in TextureImpl::getView, but at the point things seemed too hacky.

src/d3d12/d3d12-texture.cpp
sricker-nvidia: I'm fine with that if that's easier. I came at this with the assumption that it should be possible to select a subset of the depth slices, but if there isn't an assumption that slang-rhi completely wraps all d3d12 features, then I don't think my changes as is are strictly necessary.

Is there any documentation that should be updated to indicate that selecting a subset of depth slices for 3D textures is not supported?

sricker-nvidia: Ultimately I'm fine with the FirstWSlice = 0 and WSize = -1 approach. I think that's needed at the very least as currently I can repro the issue in #220, by using a non zero baseArrayLayer which will reduce the max allowed size of Wsize.

skallweitNV: Ok, lets go with the simple approach then. If we ever need selecting depth slices I'd like to repurpose layer/layerCount for that ideally. But I don't think we really have a use case for that at this point. Hope this is ok.

cmake/FetchPackage.cmake
skallweitNV: sounds useful then!

sricker-nvidia: Alright, I've kept it then.

src/rhi-shared.cpp
```diff
@@ -139,7 +139,7 @@ SubresourceRange Texture::resolveSubresourceRange(const SubresourceRange& range)
     SubresourceRange resolved = range;
     resolved.mipLevel = min(resolved.mipLevel, m_desc.mipLevelCount);
     resolved.mipLevelCount = min(resolved.mipLevelCount, m_desc.mipLevelCount - resolved.mipLevel);
-    resolved.baseArrayLayer = min(resolved.baseArrayLayer, m_desc.getLayerCount());
+    resolved.baseArrayLayer = min(resolved.baseArrayLayer, (m_desc.getLayerCount()-1));
```
sricker-nvidia, line 142: Noting an additional edit here. I found that it was possible to set baseArrayLayer == 1 and layerCount == 1 and this generated vulkan errors because baseArrayLayer was technically referencing an index that didn't exist.

PR #303, opened 2025-04-08
metal: parameter block fix

kaizhangNV:
- When set element type layout for Metal target, if the bindingType is parameter block, we will add the container type, `ParameterBlock`, to indicate that.

- When creating ShaderObject on Metal target, we have to use 'MetalArgumentBufferTier2' when the type layout is parameter block to get the uniformSize, otherwise, it will always be 0.

- Fix the issue in calculating the offset of argument buffer. We should use "MetalArgumentBufferTier2" consistently.

- Simplify the ordinary data writing. We can write the orinary data at beginning, the offsets where resource types located will just be 0. And following up process will fill up those offsets with device address (for buffers) and resource IDs (for texture/sampler).

### DIFF
--- a/src/metal/metal-shader-object.cpp
+++ b/src/metal/metal-shader-object.cpp
@@ -399,9 +399,6 @@ Result BindingDataBuilder::writeArgumentBuffer(
 )
 {
     auto argumentBufferTypeLayout = specializedLayout->getParameterBlockTypeLayout();
-    // TODO(shaderobject) for some reason, using the specialized layout here always returns zero uniform data size
-    // auto elementTypeLayout = specializedLayout->getElementTypeLayout();
-    auto elementTypeLayout = shaderObject->getElementTypeLayout();
 
     ComPtr<IBuffer> argumentBuffer;
     BufferDesc argumentBufferDesc = {};
@@ -411,6 +408,8 @@ Result BindingDataBuilder::writeArgumentBuffer(
     argumentBufferDesc.memoryType = MemoryType::Upload;
     SLANG_RETURN_ON_FAIL(m_device->createBuffer(argumentBufferDesc, nullptr, argumentBuffer.writeRef()));
     auto argumentBufferImpl = checked_cast<BufferImpl*>(argumentBuffer.get());
+
+    memcpy(argumentBufferImpl->m_buffer->contents(), shaderObject->m_data.data(), shaderObject->m_data.size());
 
     // Once the buffer is allocated, we can fill it in with the uniform data
     // and resource bindings we have tracked, using `argumentBufferTypeLayout` to obtain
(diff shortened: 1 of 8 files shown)


jarcherNV (approved):
LGTM

src/metal/metal-shader-object.cpp
```diff
@@ -444,15 +443,18 @@ Result BindingDataBuilder::writeArgumentBuffer(
...
                 const ResourceSlot& slot = shaderObject->m_slots[slotIndex + i];
                 TextureViewImpl* textureView = checked_cast<TextureViewImpl*>(slot.resource.get());
-                auto resourceId = textureView->m_textureView->gpuResourceID();
-                memcpy(argumentPtr + i * sizeof(uint64_t), &resourceId, sizeof(resourceId));
-                if (bindingRangeInfo.bindingType == slang::BindingType::MutableTexture)
-                {
-                    SLANG_RETURN_ON_FAIL(addUsedRWResource(m_bindingData, textureView->m_textureView.get()));
-                }
-                else
+                if (textureView)
```
sricker-nvidia, line 446: Do we need to throw some kind of error here if textureView is NULL? Seems like the expectation is that it should be valid.

src/metal/metal-shader-object.cpp
```diff
@@ -535,13 +543,6 @@ Result BindingDataBuilder::writeArgumentBuffer(
         }
     }
 
-    SLANG_RETURN_ON_FAIL(writeOrdinaryDataIntoArgumentBuffer(
```
sricker-nvidia, line 538: Is this call handled by the newly added, "memcpy(argumentBufferImpl->m_buffer->contents(), shaderObject->m_data.data(), shaderObject->m_data.size());" above now?

src/shader-object.cpp
```diff
@@ -108,7 +108,8 @@ Result ShaderObject::setObject(const ShaderOffset& offset, IShaderObject* object
     //    field, a constant buffer or a parameter block.
```
sricker-nvidia, line 108: is this comment still valid if we are skipping over the ParameterBlock type below?

sricker-nvidia: There's a similar instances below at L144

tests/test-nested-parameter-block.cpp
```diff
@@ -116,3 +116,63 @@ GPU_TEST_CASE("nested-parameter-block", ALL)
 
     compareComputeResult(device, resultBuffer, makeArray<uint32_t>(1123u, 1123u, 1123u, 1123u));
 }
+
+GPU_TEST_CASE("nested-parameter-block-2", ALL)
```
sricker-nvidia, line 126: I might suggest adding a comment of some sort explaining how the two tests are different. This new test seems like a more direct test of nested parameter blocks and the original test may have been working around a bug?

skallweitNV:
sorry this took so long to take a look at @kaizhangNV 

does this change imply that whenever we want to create a `ShaderObject` on metal that is later assigned to a root object, we need to create it with the `ParameterBlock` container type? because now, users didn't have to know if a shader object is later used as a parameter block or not ...

that's the main concern I have with this change

kaizhangNV:
> does this change imply that whenever we want to create a ShaderObject on metal that is later assigned to a parameter block variable in a root object, we need to create it with the ParameterBlock container type? 

That's correct, the fundamental reason behind this is that layout of ParameterBlock on metal is different from all the other targets, and Slang reflection currently is not able to automatically apply the argument buffer tier2 on the parameter block.

src/metal/metal-shader-object.cpp
kaizhangNV: No, it doesn't need to.
Because all of the slots are preserved, and it is possible that user doesn't setup for the textures inside a parameter.

src/metal/metal-shader-object.cpp
kaizhangNV: yes.

src/shader-object.cpp
kaizhangNV: The comment is still valid, we still only need to handle StructuredBuffer in this `if ` condition.
Note, `ParameterBlock` is added in this PR, so we don't want to change the behavior here at all.

tests/test-nested-parameter-block.cpp
kaizhangNV: done.

sricker-nvidia (approved):
LGTM, thanks for responding to comments. Probably good to still get a +1 from Simon before merging.

PR #325, opened 2025-04-22
Support sm 6 8 and 6 9

jkwak-work:
sm6.9 is needed for cooperative vector support.


jkwak-work:
It seems like I need to add sm6.8 and 6.9 in slang repo first.

jkwak-work:
I have a PR on slang repo that adds SM6.8 and SM6.9.

jkwak-work:
This change is required by
- https://github.com/shader-slang/slang/pull/6763

PR #333, opened 2025-05-02
Metal: Handle empty struct correctly

kaizhangNV:
There could be empty struct in AST and that is also visible to slang reflection API though such type will be eliminated in IR level.

When creating argument buffer for such type, we should check whether it's an empty struct, if it is we don't need to create argument buffer.


jkwak-work (approved):
Looks good to me.

PR #345, opened 2025-05-07
Add render-feature for cooperative-matrix-2

jkwak-work:


skallweitNV:
@jkwak-work is this still needed? I wonder if we should have that level of granularity in the features enum to begin with. I'm closing this for now as there has not been any updates to this in 2 months, feel free to reopen if necessary.

jkwak-work:
I will work on it later.
The trouble I was having is that the CI machines don't support cooperative-matrix-2 yet.
I will be able to properly test it once we upgrade the driver again.

PR #362, opened 2025-05-26
Fix the handling of combined texture sampler for dx12

jkwak-work:
This PR implements the missing handling of the combined texture sampler for dx12.

It is related to
- https://github.com/shader-slang/slang/pull/7235
- https://github.com/shader-slang/slang/issues/6982


csyonghe:
Why cant slang generate HLSL code without the sampler argument in the call?

PR #370, opened 2025-05-30
Enable GCP runner for RHI workflow

aasgaonkar:
- enable gcp for msvc debug build config
- ignore raytracing failure for GCP runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
           # Builds running on self-hosted runners (build + tests)
           - { os: windows, platform: "x86_64", compiler: clang, preset: clang, config: "Debug", flags: "unit-test,coverage", runs-on: { group: nvrgfx, labels: [Windows, X64] } }
           - { os: windows, platform: "x86_64", compiler: msvc, preset: default, config: "Release", flags: "unit-test", runs-on: { group: nvrgfx, labels: [Windows, X64] } }
+          - { os: windows, platform: "x86_64", compiler: msvc, preset: default, config: "Debug", flags: "unit-test", runs-on: { group: gcp, labels: [Windows, X64] } }
           - { os: linux, platform: "x86_64", compiler: clang, preset: clang, config: "Debug", flags: "unit-test,coverage", runs-on: { group: nvrgfx, labels: [Linux, X64] } }
           - { os: linux, platform: "x86_64", compiler: clang, preset: clang, config: "Release", flags: "unit-test", runs-on: { group: nvrgfx, labels: [Linux, X64] } }
           # Builds running on GitHub hosted runners (build + tests)
@@ -36,7 +37,6 @@ jobs:
           - { os: macos, platform: "aarch64", compiler: clang, preset: default, config: "Release", flags: "unit-test", runs-on: macos-latest }
           # Builds running on GitHub hosted runners (build only)
           - { os: windows, platform: "x86_64", compiler: clang, preset: clang, config: "Release", runs-on: windows-latest }
-          - { os: windows, platform: "x86_64", compiler: msvc, preset: default, config: "Debug", runs-on: windows-latest }
           - { os: linux, platform: "x86_64", compiler: gcc, preset: gcc, runs-on: ubuntu-latest }
           - { os: windows, platform: "aarch64", compiler: msvc, preset: default, runs-on: windows-latest }
           - { os: linux, platform: "aarch64", compiler: clang, preset: clang, runs-on: ubuntu-24.04-arm }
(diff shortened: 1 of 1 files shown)


tests/test-ray-tracing.cpp
```diff
@@ -384,6 +384,10 @@ GPU_TEST_CASE("ray-tracing-a", ALL)
     if (!device->hasFeature(Feature::RayTracing))
         SKIP("ray tracing not supported");
 
+    // Skip ray tracing tests on GCP runners
+    if (getenv("SKIP_RAY_TRACING"))
```
skallweitNV: We shouldn't skip tests like that. You can call `slang-rhi-tests` with an argument `-tce=ray-tracing*` (test case exclude) to skip the ray-tracing tests. See https://github.com/doctest/doctest/blob/master/doc/markdown/commandline.md for more details.

.github/workflows/ci.yml
```diff
@@ -29,14 +29,14 @@ jobs:
           # Builds running on self-hosted runners (build + tests)
           - { os: windows, platform: "x86_64", compiler: clang, preset: clang, config: "Debug", flags: "unit-test,coverage", runs-on: { group: nvrgfx, labels: [Windows, X64] } }
           - { os: windows, platform: "x86_64", compiler: msvc, preset: default, config: "Release", flags: "unit-test", runs-on: { group: nvrgfx, labels: [Windows, X64] } }
+          - { os: windows, platform: "x86_64", compiler: msvc, preset: default, config: "Debug", flags: "unit-test", runs-on: { group: gcp, labels: [Windows, X64] } }
```
skallweitNV, line 32: I think we prefer running on our machines for the time being. So if we want to additionally run on GCP machines, we should probably do that in a separate workflow.

skallweitNV (changes requested):

.github/workflows/ci.yml
aasgaonkar: this is taking the build only windows workflow and using GCP runner, so it should be a parallel workflow to existing test coverage.

tests/test-ray-tracing.cpp
aasgaonkar: ah ok. I used ctest exclude-regex but that didn't work. so I thought test-case level ignore needs to happen in this way. always for simpler, less code. will make this change and try out. should we do a separate unit test

skallweitNV: I should remove the ctest stuff. It's just to allow running it from IDEs but it seems to be confusing people.

.github/workflows/ci.yml
skallweitNV: That's true. I was under the impression that we are limited with GCP resources, so I'd rather see these resources spent on testing slang-rhi with latest slang on all platforms/build configs in the slang CI. We have good coverage already with the build machines here. Also we could enable release build testing, should be OK with the amount of build machines we have available now. So, I see this GCP workflow more like a way to test that slang-rhi tests run on GCP. So this could be a dispatch only workflow even.

tests/test-ray-tracing.cpp
aasgaonkar: fixed this in latest patch with tce option

.github/workflows/ci.yml
aasgaonkar: @dryu-nv Thoughts on this? Adding GCP support will add strain on the resources. we have 4 windows runners currently.

.github/workflows/ci.yml
```diff
@@ -68,8 +68,21 @@ jobs:
...
+        if: contains(matrix.flags, 'unit-test') && matrix.runs-on.group != 'gcp'
+        run: |
+          ctest --test-dir build --build-config ${{matrix.config}} --verbose
+
+      - name: Unit Tests (Windows GCP)
+        if: contains(matrix.flags, 'unit-test') && matrix.os == 'windows' && matrix.runs-on.group == 'gcp'
+        run: |
+          $testExe = Get-ChildItem -Path . -Recurse -Filter "slang-rhi-tests.exe" | Select-Object -First 1
+          if ($testExe) {
+              & $testExe.FullName -tce=ray-tracing*
```
jkwak-work, line 80: What does the option `-tce` do?
Does it exclude the ray-tracing tests?

jkwak-work (commented):
Looks good to me.

.github/workflows/ci.yml
aasgaonkar: yes, it excludes ray-tracing tests. it's a doc-test framework

.github/workflows/ci.yml
aasgaonkar: we spoke about this in the slang scrum meeting - let's merge this and enable GCP platform testing. if there are issues, we can add more windows runeers and make sure there is enough capacity for all.

aasgaonkar:
@jkwak-work can you help merge the PR?

PR #373, opened 2025-06-02
Fix slang-rhi build for slang

szihs:
Also fix build issue when building as part of slang


tests/test-ray-tracing.slang
```diff
@@ -1,8 +1,7 @@
 // ray-tracing-test-shaders.slang
-
-struct RayPayload
+struct [raypayload] RayPayload
```
skallweitNV: This is and the PAQs are already in ToT

PR #380, opened 2025-06-03
Fix several VVL violations detected in slang-test

gtong-nv:
1. `VUID-vkCmdPipelineBarrier-pBufferMemoryBarriers-02818` Removed VK_ACCESS_TRANSFER_READ_BIT as it's not supported in ray tracing stage
2. `VUID-vkCmdPipelineBarrier-dstAccessMask-06257` Check rayquery feature first before setting stages to the AS barriers
3. `VUID-RuntimeSpirv-shaderSubgroupRotateClustered-09566` Query 1.4 features and turn on ShaderSubgroupRotate* if available

Fixes part of https://github.com/shader-slang/slang/issues/4798

### DIFF
--- a/src/vulkan/vk-command.cpp
+++ b/src/vulkan/vk-command.cpp
@@ -1519,13 +1519,23 @@ void CommandRecorder::accelerationStructureBarrier(
         memBarriers[i].offset = 0;
         memBarriers[i].size = asImpl->m_buffer->m_desc.size;
     }
+
+    VkPipelineStageFlagBits dstStageMask = VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR;
+    if (m_device->m_api.m_extendedFeatures.rayQueryFeatures.rayQuery)
+    {
+        // for VUID-vkCmdPipelineBarrier-dstAccessMask-06257
+        // If the rayQuery feature is not enabled and a memory barrier dstAccessMask includes VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR,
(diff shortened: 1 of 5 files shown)


src/vulkan/vk-command.cpp
```diff
@@ -1519,13 +1519,23 @@ void CommandRecorder::accelerationStructureBarrier(
...
         memBarriers[i].size = asImpl->m_buffer->m_desc.size;
     }
+
+    VkPipelineStageFlagBits dstStageMask = VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR;
+    if (m_device->m_api.m_extendedFeatures.rayQueryFeatures.rayQuery)
+    {
+        // for VUID-vkCmdPipelineBarrier-dstAccessMask-06257
+        // If the rayQuery feature is not enabled and a memory barrier dstAccessMask includes VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR,
+        // dstStageMask must not include any of the VK_PIPELINE_STAGE_*_SHADER_BIT stages except VK_PIPELINE_STAGE_RAY_TRACING_SHADER_BIT_KHR
+        dstStageMask = (VkPipelineStageFlagBits)(VK_PIPELINE_STAGE_ACCELERATION_STRUCTURE_BUILD_BIT_KHR |
```
skallweitNV, line 1529: Ideally we would track the pipeline stage needing access to a resource and set the barriers accordingly. I remember seeing validation errors in graphics pipelines as well, because right now we always insert barriers with stage masks containing "all possible stages". But this requires some more thought and careful implementation. For now, this should do to silence the validation error.

PR #392, opened 2025-06-05
fix test failures

kaizhangNV:
In issue: https://github.com/shader-slang/slang/issues/7315 There are precompiled-module and precompiled-module-mixed test failed.

The reason is that we treat warning message to error in the test. We should just ignore that.


csyonghe:
I merged the slang PR, does this PR need to be updated?

PR #400, opened 2025-06-12
Fix linking errros when buildling slang-rhi examples w/ slang

szihs:
Add optix-dev submodule

target_link_libraries(example-surface PRIVATE slang-rhi slang glfw) By listing slang-rhi first, the linker identifies the symbols it needs. Then, when it processes slang (libslang.so), it can find and link the required spReflection... functions, resolving the "undefined reference" errors.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "external/metal-cpp"]
 	path = external/metal-cpp
 	url = https://github.com/bkaradzic/metal-cpp.git
+[submodule "external/optix-dev"]
+	path = external/optix-dev
+	url = https://github.com/NVIDIA/optix-dev.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "external/metal-cpp"]
 	path = external/metal-cpp
 	url = https://github.com/bkaradzic/metal-cpp.git
+[submodule "external/optix-dev"]
```
skallweitNV: Let's not add this. We fetch optix headers through fetch content already.

cmake/FindOptiX.cmake
```diff
@@ -2,15 +2,27 @@ include(FindPackageHandleStandardArgs)
```
skallweitNV: We should probably just remove `FindOptiX.cmake` as it's not used.

skallweitNV (changes requested):

PR #402, opened 2025-06-12
Add sphere tests

ksavoie-nv:
The new tests check that we can intersect with sphere primitives, and that both the plain and hit object sphere intrinsics return valid results.

To use spheres, OptiX requires the application to set OPTIX_PRIMITIVE_TYPE_FLAGS_SPHERE in its pipeline compile options and link in the builtin sphere intersection module. To allow this, we add a new flag named `EnableSpheres` to `RayTracingPipelineFlags`. This flag instructs slang-rhi to set the necessary pipeline compile option and make the builtin sphere intersection module available via the entry point name `__builtin_intersection__sphere`.


ksavoie-nv:
The unit tests failed, but that's expected. This PR fixes a blocking issue in Slang: https://github.com/shader-slang/slang/pull/7427. I'll re-trigger the unit tests after that PR is merged.

skallweitNV:
If this test depends on new slang features, you also have to update `SLANG_RHI_FETCH_SLANG_VERSION` in `CMakeLists.txt` to use the new slang version that has support for the new intrinsics.

src/cuda/cuda-pipeline.cpp
```diff
@@ -223,11 +244,17 @@ Result DeviceImpl::createRayTracingPipeline2(const RayTracingPipelineDesc& desc,
         }
         if (hitGroupDesc.intersectionEntryPoint)
         {
-            optixProgramGroupDesc.hitgroup.moduleIS =
-                optixModules[entryPointNameToModuleIndex[hitGroupDesc.intersectionEntryPoint]];
-            entryFunctionNameIS = std::string("__intersection__") + hitGroupDesc.intersectionEntryPoint;
-            optixProgramGroupDesc.hitgroup.entryFunctionNameIS = entryFunctionNameIS.data();
+            if( std::strcmp( hitGroupDesc.intersectionEntryPoint, "__builtin_intersection__sphere" ) == 0 )
```
skallweitNV: Can you run `pre-commit run --all-files` to apply formatting rules? We really should add that to the CI workflow.

PR #404, opened 2025-06-15
filter d3d12 validation messages

westlicht:
- filter d3d12 validation messages in callback
- filter `D3D12_MESSAGE_ID_FENCE_ZERO_WAIT`


src/d3d12/d3d12-device.cpp
```diff
@@ -26,6 +26,13 @@ namespace rhi::d3d12 {
 
 static const uint32_t D3D_FEATURE_LEVEL_12_2 = 0xc200;
 
+// List of validation messages that are filtered by default.
```
tdavidovicNV: "filtered out" maybe?

PR #436, opened 2025-06-30
Add linear swept sphere tests

ksavoie-nv:
Add tests that check that we can intersect with linear swept spheres and that the associated plain and hit object intrinsics work as expected.

Also add the plumbing necessary to use LSS with OptiX:
- Add the LSS feature to OptiX devices
- Add `EnableLinearSweptSpheres` flag to `RayTracingPipelineFlags`
- Update OptiX accel structure validation and conversion to handle LSS
- Add `__builtin_intersection__linear_swept_spheres` intersection program to OptiX pipelines and make it available when  `EnableLinearSweptSpheres` is set

### DIFF
--- a/src/cuda/cuda-acceleration-structure.cpp
+++ b/src/cuda/cuda-acceleration-structure.cpp
@@ -196,7 +196,49 @@ Result AccelerationStructureBuildDescConverter::convert(
     }
     case AccelerationStructureBuildInputType::LinearSweptSpheres:
     {
-        return SLANG_E_NOT_AVAILABLE;
+        for (uint32_t i = 0; i < buildDesc.inputCount; ++i)
+        {
+            const AccelerationStructureBuildInputLinearSweptSpheres& linearSweptSpheres =
+                buildDesc.inputs[i].linearSweptSpheres;
+            if (linearSweptSpheres.vertexBufferCount != 1)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+            if (linearSweptSpheres.vertexPositionFormat != Format::RGB32Float)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+            if (linearSweptSpheres.vertexRadiusFormat != Format::R32Float)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+            if (!linearSweptSpheres.indexBuffer)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+
+            OptixBuildInput& buildInput = buildInputs[i];
+            buildInput = {};
+            buildInput.type = OPTIX_BUILD_INPUT_TYPE_CURVES;
+            buildInput.curveArray.curveType = OPTIX_PRIMITIVE_TYPE_ROUND_LINEAR;
+            buildInput.curveArray.numPrimitives = linearSweptSpheres.primitiveCount;
+
+            pointerList.push_back(linearSweptSpheres.vertexPositionBuffers[0].getDeviceAddress());
+            buildInput.curveArray.numVertices = linearSweptSpheres.vertexCount;
+            buildInput.curveArray.vertexBuffers = &pointerList.back();
+            buildInput.curveArray.vertexStrideInBytes = linearSweptSpheres.vertexPositionStride;
+
+            pointerList.push_back(linearSweptSpheres.vertexRadiusBuffers[0].getDeviceAddress());
+            buildInput.curveArray.widthBuffers = &pointerList.back();
+            buildInput.curveArray.widthStrideInBytes = linearSweptSpheres.vertexRadiusStride;
+
+            buildInput.curveArray.indexBuffer = linearSweptSpheres.indexBuffer.getDeviceAddress();
+
+            buildInput.curveArray.flag = translateGeometryFlags(linearSweptSpheres.flags);
+
+            // TODO: Add support for end caps
+        }
+        break;
     }
     default:
         return SLANG_E_INVALID_ARG;
--- a/src/cuda/cuda-acceleration-structure.cpp
+++ b/src/cuda/cuda-acceleration-structure.cpp
@@ -216,6 +216,10 @@ Result AccelerationStructureBuildDescConverter::convert(
             {
                 return SLANG_E_INVALID_ARG;
             }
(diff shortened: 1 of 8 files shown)


src/cuda/cuda-acceleration-structure.cpp
```diff
@@ -196,8 +196,52 @@ Result AccelerationStructureBuildDescConverter::convert(
...
+            }
+            if (linearSweptSpheres.vertexRadiusFormat != Format::R32Float)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+            if (!linearSweptSpheres.indexBuffer)
+            {
+                return SLANG_E_INVALID_ARG;
+            }
+            if (linearSweptSpheres.endCapsMode == LinearSweptSpheresEndCapsMode::None)
```
skallweitNV, line 219: we should probably also reject if `indexingMode` is not set to `List` as OptiX expects indices that way (also add to validation)

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -289,7 +289,58 @@ void validateAccelerationStructureBuildDesc(DebugContext* ctx, const Acceleratio
...
+            {
+                if (!linearSweptSpheres.vertexPositionBuffers[j].buffer)
+                {
+                    RHI_VALIDATION_ERROR("vertexBuffers cannot be null.");
+                }
+            }
+
+            if (ctx->deviceType == DeviceType::CUDA)
+            {
+                if (linearSweptSpheres.endCapsMode == LinearSweptSpheresEndCapsMode::None)
```
skallweitNV: As mentioned above, add a check for `indexingMode` to be `List` here. Also, we could check that the length of the index buffer is at least `primitiveCount + 1` here.

ksavoie-nv:
@skallweitNV I read the OptiX and VK LSS documentation more carefully and realized that OptiX uses successive and not list indexing. I updated validation to reflect that, added the extra checks that you mentioned, reorganized the validation layer's checks.

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -289,7 +289,78 @@ void validateAccelerationStructureBuildDesc(DebugContext* ctx, const Acceleratio
...
+            if (linearSweptSpheres.indexBuffer)
+            {
+                if (linearSweptSpheres.indexingMode == LinearSweptSpheresIndexingMode::List)
+                {
+                    if (linearSweptSpheres.indexCount < linearSweptSpheres.primitiveCount + 1)
+                    {
+                        RHI_VALIDATION_ERROR("indexCount must be >= primitiveCount + 1 when indexingMode is List.");
+                    }
+                }
+                else
```
skallweitNV: maybe do `else if (linearSweptSpheres.indexingMode == LinearSweptSpheresIndexingMode::Successive)` instead

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -289,7 +289,78 @@ void validateAccelerationStructureBuildDesc(DebugContext* ctx, const Acceleratio
...
+                break;
+            default:
+                RHI_VALIDATION_ERROR("Unsupported vertexRadiusFormat. Valid values are R32Float.");
+            }
+
+            if (linearSweptSpheres.indexBuffer)
+            {
+                if (linearSweptSpheres.indexingMode == LinearSweptSpheresIndexingMode::List)
+                {
+                    if (linearSweptSpheres.indexCount < linearSweptSpheres.primitiveCount + 1)
```
skallweitNV: This seems wrong to me, `List` is where the index buffer contains pairs of indices, specifying the start/end vertex of each segment. In this mode we should verify that `indexCount >= 2 * primitiveCount`.

In `Successive` mode, each index `k` specifies a segment that uses vertices `(k, k+1)` for the segment. In this mode we should check for `indexCount >= primitiveCount`.

Does this sound right or did I misunderstand?

From the VK docs:
> VK_RAY_TRACING_LSS_INDEXING_MODE_LIST_NV specifies that a list of indices is provided where each consecutive pair of indices define a LSS primitive.
>
> VK_RAY_TRACING_LSS_INDEXING_MODE_SUCCESSIVE_NV specifies a successive implicit indexing format, in which each LSS primitive is defined by two successive positions and radii, (k, k + 1), where k is a single index provided in the index buffer. In this indexing scheme, there is a 1:1 mapping between the index buffer and primitive index within the geometry.

ksavoie-nv: Good catch, thanks! And yes, I think your interpretation of the docs is correct - I'll update the check accordingly.

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -333,20 +333,24 @@ void validateAccelerationStructureBuildDesc(DebugContext* ctx, const Acceleratio
             {
                 if (linearSweptSpheres.indexingMode == LinearSweptSpheresIndexingMode::List)
                 {
-                    if (linearSweptSpheres.indexCount < linearSweptSpheres.primitiveCount + 1)
+                    if (linearSweptSpheres.indexCount < linearSweptSpheres.primitiveCount * 2)
                     {
                         RHI_VALIDATION_ERROR("indexCount must be >= primitiveCount + 1 when indexingMode is List.");
```
skallweitNV:
```suggestion
                        RHI_VALIDATION_ERROR("indexCount must be >= primitiveCount when indexingMode is List.");
```

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -333,20 +333,24 @@ void validateAccelerationStructureBuildDesc(DebugContext* ctx, const Acceleratio
...
                     }
                 }
-                else
+                else if (linearSweptSpheres.indexingMode == LinearSweptSpheresIndexingMode::Successive)
                 {
-                    if (linearSweptSpheres.indexCount != linearSweptSpheres.primitiveCount)
+                    if (linearSweptSpheres.indexCount < linearSweptSpheres.primitiveCount)
                     {
                         RHI_VALIDATION_ERROR(
                             "indexCount must be equal to primitiveCount when indexingMode is Successive."
```
skallweitNV:
```suggestion
                            "indexCount must be >= primitiveCount when indexingMode is Successive."
```

skallweitNV (approved):
LGTM, thanks!

PR #439, opened 2025-07-06
External cuda context support

ccummingsNV:
Support for initializing from external cuda devices, and specifying the cuda stream for a submit.

Added tests to verify interaction of different devices/streams + also some general tests for pointer support.

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -2405,7 +2405,7 @@ enum class QueueType
 
 // The NULL CUDA stream is valid (it refers to the default stream), so we
 // use this constant to indicate the absence of one.
-void* const kInvalidCUDAStream = reinterpret_cast<void* const>(~uintptr_t{0});
+void* const kInvalidCUDAStream = reinterpret_cast<void*>(~uintptr_t{0});
 
 struct SubmitDesc
 {
@@ -2689,8 +2689,9 @@ struct DeviceDesc
     DeviceType deviceType = DeviceType::Default;
     // The device's handles (if they exist) and their associated API. For D3D12, this contains a single
     // NativeHandle for the ID3D12Device. For Vulkan, the first NativeHandle is the VkInstance, the second is the
-    // VkPhysicalDevice, and the third is the VkDevice. For CUDA, this contains a handle for the device,
-    // and optionally a handle for the context (if not provided, a context will be created internally).
+    // VkPhysicalDevice, and the third is the VkDevice.
+    // For CUDA, this contains a handle for the device and/or a handle for the context. If context is provided,
+    // device is ignored (and retreived from context). If only device is provided, a context is created for it.
     DeviceNativeHandles existingDeviceHandles;
     // LUID of the adapter to use. Use getGfxAdapters() to get a list of available adapters.
     const AdapterLUID* adapterLUID = nullptr;
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -2418,7 +2418,7 @@ struct SubmitDesc
     const uint64_t* signalFenceValues = nullptr;
     uint32_t signalFenceCount = 0;
 
-    // The CUDA stream to use for the submission. Ignored on none-CUDA backends.
+    // The CUDA stream to use for the submission. Ignored on non-CUDA backends.
     // If set to `kInvalidCUDAStream`, the CUDA stream associated with the device
     // queue is used, which in the default case the default (NULL) stream.
     void* cudaStream = kInvalidCUDAStream;
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -573,6 +573,7 @@ enum class NativeHandleType
     CUmipmappedArray = 0x00050006,
     CUtexObject = 0x00050007,
     CUsurfaceObject = 0x00050008,
+    CUcontext = 0x00050009,
 
     OptixDeviceContext = 0x00060001,
     OptixTraversableHandle = 0x00060002,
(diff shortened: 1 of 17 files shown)


include/slang-rhi.h
```diff
@@ -2402,16 +2403,25 @@ enum class QueueType
...
+    ICommandBuffer** commandBuffers = nullptr;
+    uint32_t commandBufferCount = 0;
+    IFence** waitFences = nullptr;
+    const uint64_t* waitFenceValues = nullptr;
+    uint32_t waitFenceCount = 0;
+    IFence** signalFences = nullptr;
+    const uint64_t* signalFenceValues = nullptr;
+    uint32_t signalFenceCount = 0;
+
+    // The CUDA stream to use for the submission. Ignored on none-CUDA backends.
```
```suggestion
    // The CUDA stream to use for the submission. Ignored on non-CUDA backends.
```

src/core/assert.h
```diff
@@ -2,9 +2,18 @@
 
 namespace rhi {
 
-[[noreturn]] void handleAssert(const char* message, const char* file, int line);
+class ScopedDisableAsset
```
skallweitNV: should probably read `ScopedDisableAssert`

src/core/assert.h
```diff
@@ -2,9 +2,18 @@
...
+    ScopedDisableAsset();
+    ~ScopedDisableAsset();
+};
 
-}
+void handleAssert(const char* message, const char* file, int line);
+
+} // namespace rhi
+
+#define SLANG_RHI_DISABLE_ASSERT_SCOPE() ::rhi::ScopedDisableAsset __disable_assert;
```
skallweitNV: `__` prefix is reserved .. so maybe use `rhi_disable_assert__` or something

src/cuda/cuda-command.cpp
```diff
@@ -16,7 +16,7 @@ class CommandExecutor
 {
 public:
     DeviceImpl* m_device;
-    CUstream m_stream;
+    CUstream m_activeStream;
```
skallweitNV: I would have kept this just `m_stream`, it's the stream we write commands to.

src/cuda/cuda-command.h
```diff
@@ -9,7 +9,23 @@ namespace rhi::cuda {
...
+            : m_queue(queue)
+        {
+            m_queue->m_activeStream = stream;
+        }
+
+        ~StreamScope() { m_queue->m_activeStream = (CUstream)kInvalidCUDAStream; }
+    };
+
+    CUstream m_defaultStream;
+    CUstream m_activeStream;
```
skallweitNV: Why do we need all of this? Shouldn't the "stream overwriting" logic be completely local to the `submit` call?

src/cuda/cuda-command.cpp
```diff
@@ -712,13 +720,28 @@ void CommandExecutor::cmdExecuteCallback(const commands::ExecuteCallback& cmd)
 CommandQueueImpl::CommandQueueImpl(Device* device, QueueType type)
     : CommandQueue(device, type)
 {
-    SLANG_CUDA_ASSERT_ON_FAIL(cuStreamCreate(&m_stream, 0));
+    // On CUDA, treat the graphics stream as the default stream, identified
+    // by a NULL ptr. When we support async compute queues on d3d/vulkan,
+    // they will be equivalent to secondary, none-default streams in cuda.
```
skallweitNV, line 717:
```suggestion
    // they will be equivalent to secondary, non-default streams in cuda.
```

src/cuda/cuda-surface.cpp
```diff
@@ -841,21 +841,25 @@ Result SurfaceImpl::present()
...
+
     // Signal semaphore on CUDA stream.
     // This would be the preferred way, but leads to Vulkan validation errors because
     // the validation layer cannot se the signal sent from the CUDA stream.
 #if 0
     {
         CUDA_EXTERNAL_SEMAPHORE_SIGNAL_PARAMS signalParams = {};
         signalParams.params.fence.value = ++frameData.signalValue;
-        cuSignalExternalSemaphoresAsync(&frameData.cudaSemaphore, &signalParams, 1, m_deviceImpl->m_queue->m_stream);
+        cuSignalExternalSemaphoresAsync(&frameData.cudaSemaphore, &signalParams, 1, nullptr);
```
skallweitNV: we can't we use the queue's stream here anymore?

src/cuda/cuda-surface.cpp
```diff
@@ -841,21 +841,25 @@ Result SurfaceImpl::present()
...
     }
 #endif
 
-    // Synchronize the CUDA stream and manually signal semaphore on Vulkan (avoid validation errors above).
+    // As the semaphore approach doesn't currently work, call cuStreamSynchronize, which blocks
+    // the host until the default CUDA stream is completely drained.
 #if 1
     {
-        cuStreamSynchronize(m_deviceImpl->m_queue->m_stream);
+        cuStreamSynchronize(nullptr);
```
skallweitNV: same here

src/debug-layer/debug-device.h
```diff
@@ -1,3 +1,4 @@
+
```
skallweitNV: extra newline

skallweitNV (changes requested):

src/cuda/cuda-command.h
ccummingsNV: Using scoping to avoid issues where the active stream isn't turned off if submit bails out due to error. Various ways to do it - could have made 'submit' wrap a 'submitinternal'. I've always been a fan of the scopped handler myself.

src/cuda/cuda-surface.cpp
ccummingsNV: In the current design, to avoid accidental errors with stale state, the queue by default does not have an active stream - it is invalid and you have to tell the queue what stream to add to. We could use the queue's default stream instead - neither seemed cleaner.

src/cuda/cuda-command.cpp
ccummingsNV: Can do - I renamed it because 'active' makes it clear there won't always be an active one, but I'm not fussy.

src/cuda/cuda-command.h
skallweitNV: It just feels like this bleeds out to more places than it really needs to IMO. 
Why do we ever have to change `m_stream` in queue in the first place? Can't we just pass the overwritten stream to the `CommandExecutor`?

PR #441, opened 2025-07-07
CUDA paged memallocs + delayed resource freeing

ccummingsNV:
Adds paging system for allocated cuda memory to reduce calls to cuda allocation functions
Implements use of cuda events to track in flight command buffers
Remove explicit sync after cuda submit

### DIFF
--- a/src/cuda/cuda-dual-page-allocator.cpp
+++ b/src/cuda/cuda-dual-page-allocator.cpp
@@ -57,11 +57,10 @@ Result DualPageAllocator::allocate(size_t minSize, DualPageAllocator::Handle** h
 
 Result DualPageAllocator::allocate(size_t minSize, DualPageAllocator::Page& outPage)
 {
-    std::lock_guard<std::mutex> lock(m_mutex);
     if (minSize == 0)
     {
         outPage = {};
-        return SLANG_OK;
+        return SLANG_E_INVALID_ARG;
     }
 
     size_t idx = nextPowerOf2(minSize);
@@ -69,6 +68,8 @@ Result DualPageAllocator::allocate(size_t minSize, DualPageAllocator::Page& outP
     {
         return SLANG_E_OUT_OF_MEMORY;
     }
+
+    std::lock_guard<std::mutex> lock(m_mutex);
 
     // Get existing or create new page.
     if (m_freePages[idx].empty())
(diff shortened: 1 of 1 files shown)


src/cuda/cuda-constant-buffer-pool.cpp
```diff
@@ -65,8 +52,8 @@ Result ConstantBufferPool::allocate(size_t size, Allocation& outAllocation)
         Page& page = m_largePages.back();
         SLANG_RETURN_ON_FAIL(createPage(size, page));
         page.usedSize = size;
-        outAllocation.hostData = page.hostData;
-        outAllocation.deviceData = page.deviceData;
+        outAllocation.hostData = page.handle->getHostPtr();
```
skallweitNV, line 55: Worth thinking about keeping `hostData` and `deviceData` members in `Page`. Makes debugging easier and less indirection in debug builds. But fine to keep as is (not having redundant data is probably better).

src/cuda/cuda-dual-page-allocator.cpp
```diff
@@ -0,0 +1,144 @@
...
+    Page page;
+    SLANG_RETURN_ON_FAIL(allocate(minSize, page));
+    RefPtr<Handle> res = new Handle(this, page);
+    returnRefPtr(handle, res);
+    return SLANG_OK;
+}
+
+Result DualPageAllocator::allocate(size_t minSize, DualPageAllocator::Page& outPage)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
```
skallweitNV: you could move the guard down after the initial checks and size computation

src/cuda/cuda-dual-page-allocator.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return SLANG_OK;
+}
+
+Result DualPageAllocator::free(Page page)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (page.hostData == nullptr && page.deviceData == 0)
+    {
+        // Nothing to free
+        return SLANG_OK;
```
skallweitNV: double frees are typically some logic error, so not sure I would just return OK here

src/cuda/cuda-dual-page-allocator.cpp
```diff
@@ -0,0 +1,144 @@
...
+    m_totalAllocated += outPage.size;
+
+    // Unless code is broken, the page should be big enough to fit 'minSize'
+    SLANG_RHI_ASSERT(outPage.size >= minSize);
+    return SLANG_OK;
+}
+
+Result DualPageAllocator::free(Page page)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
```
skallweitNV: could move down after the page check below

skallweitNV (approved):
LGTM, minor suggestions

src/cuda/cuda-constant-buffer-pool.cpp
ccummingsNV: I'd rather keep it this way until it's demonstrably an issue - redundant state is my least favourite thing in the world

src/cuda/cuda-dual-page-allocator.cpp
ccummingsNV: in allocate I returned 'null' if the user asks for a page of size 0 so this felt symetric. I've made both cases an error instead.

src/cuda/cuda-constant-buffer-pool.cpp
skallweitNV: agreed

PR #447, opened 2025-07-15
Add tests that exercise ReorderThread and Invoke

ksavoie-nv:
The tests check that we get correct results when we create a hit object, reorder, then invoke the hit obj.

Note that we don't check that the reorder call actually has an effect, because whether or not the implementation chooses to reorder the threads is unspecified.

Also, fix a bug where the D3D12 implementation didn't add the UAV slot for NVAPI to the root descriptor table

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -2424,6 +2424,8 @@ public:
         ResourceState state
     ) = 0;
 
+    virtual SLANG_NO_THROW void SLANG_MCALL globalBarrier() = 0;
+
     inline void setTextureState(ITexture* texture, ResourceState state)
     {
         setTextureState(texture, kEntireTexture, state);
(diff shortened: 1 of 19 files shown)


src/d3d12/d3d12-shader-object-layout.cpp
```diff
@@ -963,6 +968,23 @@ Result RootShaderObjectLayoutImpl::createRootSignatureFromSlang(
         builder.addAsValue(entryPoint->getVarLayout(), rootDescriptorSetIndex);
     }
 
+#if SLANG_RHI_ENABLE_NVAPI
+    // If NVAPI is enabled, we need to add a "fake" UAV descriptor range to the root signature.
+    if (device->m_nvapiShaderExtension)
```
ksavoie-nv: @skallweitNV I noticed that the UAV for NVAPI wasn't being added and did so here, but it seems to break a bunch of tests. Do you know the proper way to do this? I'm not very familiar with DX root signatures.

ksavoie-nv: I decided to disable the reorder test for D3D12 and leave the D3D12 + NVAPI fix for someone more experienced with that backend.

PR #448, opened 2025-07-16
Global barriers

ccummingsNV:
- Add global barrier concept to slang-rhi
- Implemented global barrier in vulkan+d3d12 (other platforms don't need it as state tracking is implicit)
- Debug option to break state tracking + tests to verify the global barrier fixes it

### DIFF
--- a/src/d3d12/d3d12-command.cpp
+++ b/src/d3d12/d3d12-command.cpp
@@ -1407,7 +1407,7 @@ void CommandRecorder::requireTextureState(TextureImpl* texture, SubresourceRange
 
 void CommandRecorder::commitBarriers()
 {
-    if (gDebugDisableStateTracking)
+    if (detail::gDebugDisableStateTracking)
         return;
 
     short_vector<D3D12_RESOURCE_BARRIER, 16> barriers;
--- a/src/d3d12/d3d12-command.cpp
+++ b/src/d3d12/d3d12-command.cpp
@@ -104,6 +104,7 @@ public:
     void cmdConvertCooperativeVectorMatrix(const commands::ConvertCooperativeVectorMatrix& cmd);
     void cmdSetBufferState(const commands::SetBufferState& cmd);
     void cmdSetTextureState(const commands::SetTextureState& cmd);
+    void cmdGlobalBarrier(const commands::GlobalBarrier& cmd);
     void cmdPushDebugGroup(const commands::PushDebugGroup& cmd);
     void cmdPopDebugGroup(const commands::PopDebugGroup& cmd);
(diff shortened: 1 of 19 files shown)


src/d3d12/d3d12-command.cpp
```diff
@@ -1271,6 +1272,17 @@ void CommandRecorder::cmdSetTextureState(const commands::SetTextureState& cmd)
     m_stateTracking.setTextureState(checked_cast<TextureImpl*>(cmd.texture), cmd.subresourceRange, cmd.state);
 }
 
+void CommandRecorder::cmdGlobalBarrier(const commands::GlobalBarrier& cmd)
+{
+    // Global barrier on D3D12 is implemented with a UAV barrier pointing at null resource.
+    // https://learn.microsoft.com/en-us/windows/win32/direct3d12/using-resource-barriers-to-synchronize-resource-states-in-direct3d-12
+    // TODO: Look at using D3D12 advanced barriers when available, currently only experimental in agility sdk though.
```
tdavidovicNV, line 1279: Don't we have the pattern where, agility sdk is available use that, otherwise use the default?

src/device.cpp
```diff
@@ -8,6 +8,9 @@
 
 namespace rhi {
 
+// Debug option for tests to turn off state tracking (so we can effectively test explicit barriers)
+bool gDebugDisableStateTracking = false;
```
tdavidovicNV, line 13: I'd push this into some `testing` namespace.. debugging things you're never supposed to touch outside tests shouldn't be in the main namespace.

src/vulkan/vk-command.cpp
```diff
@@ -1357,6 +1383,9 @@ void CommandRecorder::requireTextureState(TextureImpl* texture, SubresourceRange
 
 void CommandRecorder::commitBarriers()
 {
+    if (gDebugDisableStateTracking)
```
tdavidovicNV: Should this be alive in Release? If we want all the performance and this is testing only, I'd imagine it would be compiled in in DEBUG only.

tdavidovicNV (approved):
Generally LGTM, but I'd definitely push the debugging global into `rhi::testing` namespace or something that's very obviously "don't touch this" scope, rather than `rhi`.

And ideally it would be compiled out outside the testing context, even though it can probably be perfectly predicted.

src/d3d12/d3d12-command.cpp
ccummingsNV: Yeah but this is the _preview_ agility sdk - like the normal agility sdk, but less reliable :)

PR #450, opened 2025-07-17
Enable cuda tests

mkeshavaNV:


ccummingsNV:
Both slang-rhi and slangpy specify the version of slang to use in their cmake file. I believe @tomas-davidovic updated it today though.

tdavidovicNV (approved):
LGTM

PR #456, opened 2025-07-23
Switching to 2025.13, for the latest Slang fixes

tdavidovicNV:


samestep:
@jkwak-work Determinism seems like a pretty good reason; that way if anything breaks you can see that it was caused by this PR, instead of randomly breaking on an unrelated PR. For instance, shader-slang/slang-vscode-extension#42 was caused because we just cloned the latest version of Emscripten instead of actually specifying what version to use.

jkwak-work:
I see a few benefits of using the hardcoded number.
Thanks for the examples.

Let's go with how we are doing for now.

PR #458, opened 2025-07-28
Don't request explicit profile name for vk.

csyonghe:
This will allow Slang to automatically upgrade spirv version depending on what feature is used, without producing any warnings.


skallweitNV (approved):
LGTM

PR #459, opened 2025-07-28
Minor clean for gDebugDisableStateTracking option

skallweitNV:
Moved declaration to device.h into testing namespace, as this is only used for testing purposes.


tdavidovicNV (approved):
LGTM

PR #464, opened 2025-07-29
Reduce the number of cmake cache variables

skallweitNV:
This PR removes a lot of the cache variables for slang-rhi's dependencies, in order to make it easier to switch between branches and keep dependencies automatically up-to-date.

The remaining cache variables are:
- SLANG_RHI_ENABLE_D3D11
- SLANG_RHI_ENABLE_D3D12
- SLANG_RHI_ENABLE_AGILITY_SDK
- SLANG_RHI_ENABLE_NVAPI
- SLANG_RHI_ENABLE_VULKAN
- SLANG_RHI_ENABLE_METAL
- SLANG_RHI_ENABLE_CUDA
- SLANG_RHI_ENABLE_OPTIX
- SLANG_RHI_ENABLE_WGPU
- SLANG_RHI_FETCH_SLANG
- SLANG_RHI_FETCH_SLANG_VERSION
- SLANG_RHI_SLANG_INCLUDE_DIR
- SLANG_RHI_SLANG_BINARY_DIR

All the URLs for the dependencies are now normal variables set to sensible default values and generally have the form SLANG_RHI_XXX_URL (removed the _FETCH_ part) so they don't pick up old values during the transition.


src/d3d12/d3d12-texture.cpp
```diff
@@ -207,7 +207,6 @@ D3D12_CPU_DESCRIPTOR_HANDLE TextureImpl::getUAV(
         viewDesc.Texture2DArray.FirstArraySlice = range.layer;
         viewDesc.Texture2DArray.PlaneSlice = D3DUtil::getPlaneSlice(viewDesc.Format, aspect);
         break;
-#if SLANG_RHI_ENABLE_AGILITY_SDK
```
tdavidovicNV, line 210: Does this no longer depend on Agility SDK?

skallweitNV: The reason I removed this is because we depend on newer D3D12 features in other places (bindless for examples) so the minimum required version always supports this.

Btw. minimum version check will follow in a separate PR.

PR #465, opened 2025-07-29
Cleanup utility functions

skallweitNV:
This PR contains some long overdue cleanup of the various utility functions for the different backends and makes things more uniform, remove a bunch of dead code etc.

Specifically the following has been done:
- `backend-helper-functions.h/cpp` and `backend-util.h/cpp` have been merged into a single `backend-utils.h/cpp`
- split up the mix of D3D11/D3D12 utils into `d3d11-utils.h`, `d3d12-utils.h` and `d3d-utils.h`
- removed "class namespaces" like `VulkanUtil` etc. and just use free standing functions
- moved the various `createXXXDevice` and `getXXXAdapters` functions to the respective backend device code
- introduce `backend-api.h` headers across all backends that include the graphics API headers
- make `translateXXX` functions to translate slang-rhi enums to the backend enums more consistent
  - exhaustive case labels to catch all places when adding a new enum value to the slang-rhi enums
  - uniform error handling (assert and return `0`)
- remove a whole bunch of dead code
- cleaned up includes (also, all headers are no self-sufficient, i.e. can be compiled when added into a empty cpp file)
- removed all the `#define CASE(from, to)` macros to just plain switch statements (often less code, more compatible with intellisense)


tdavidovicNV (approved):
LGTM

PR #469, opened 2025-08-05
Enable Claude Code Actions

szihs:
Tag @claude in PR/Issue comments to trigger the Claude Code assistant:
  - Bug fixes and code reviews
  - Feature implementation assistance
  - Code explanations and refactoring
  - Test writing and debugging
  - Documentation updates


tdavidovicNV:
@szihs is this ready for a review, or should this be turned into draft for the time being?

szihs:
> @szihs is this ready for a review, or should this be turned into draft for the time being?

@tdavidovicNV 
The PR depends on common GitHub [Action](https://github.com/shader-slang/slang/pull/7811) :
It is ready for review.

jkwak-work (approved):
Looks good to me.

.claude/hooks/stop.py
```diff
@@ -0,0 +1,120 @@
...
+        # Check if all todos are completed
+        incomplete_todos = []
+        for todo in last_todo_write:
+            if todo.get("status") != "completed":
+                incomplete_todos.append(todo)
+
+        return len(incomplete_todos) == 0, incomplete_todos
+
+    except Exception:
+        # If any error occurs during parsing, assume OK to proceed
```
skallweitNV, line 57: This feels wrong to me, but so does most of this script.
But I guess this is not super important, so not too fussy.

skallweitNV (approved):

.claude/hooks/stop.py
szihs: The hooks are something tool would just ignore if there is error. So there is no much point catching and giving to tool since we tool won't be able to change track.

szihs:
@skallweitNV I don't have necessary permission to merge. (Can you help) ?

.claude/hooks/stop.py
skallweitNV: Yeah, but this whole script reads to me like "do a bunch of stuff, but really, we don't care if it works or not". Given that this runs automated, we will have no idea if the script actually does what it's supposed to do at all. There is no difference between an expected/unexpected error or just the fact that the function no longer works due to different inputs for example. This just feels very wrong to me.

szihs:
@skallweitNV My bad, triggered it with the values set.

PR #470, opened 2025-08-05
Add CUDA synchronization points

skallweitNV:
Add cuStreamSynchronize calls before kernel launches and before reading buffers/textures. This fixes issues with texture writes not being visible in subsequent launches. CUDA driver API does less automatic synchronization compared to CUDA runtime API. This needs more investigation and switching to a finer grained synchronization scheme.


tdavidovicNV (approved):
LGTM

PR #471, opened 2025-08-05
Add reporting of RHI live objects if enabled

skallweitNV:


src/core/smart-pointer.cpp
```diff
@@ -1,7 +1,23 @@
...
+#include <cstdio>
+
 namespace rhi {
 
+#if SLANG_RHI_ENABLE_REF_OBJECT_TRACKING
+void RefObjectTracker::reportLiveObjects()
+{
+    if (!objects.empty())
+    {
+        printf("Found %zu live RHI objects!\n", objects.size());
```
tdavidovicNV, line 12: Slang RHI isn't using `fmt::format`?

Either way, shouldn't this be printed to `stdcerr`?

src/core/smart-pointer.h
```diff
@@ -7,7 +7,7 @@
 #include <atomic>
 #include <type_traits>
 
-#define SLANG_RHI_ENABLE_REF_OBJECT_TRACKING 0
+#define SLANG_RHI_ENABLE_REF_OBJECT_TRACKING 1
```
tdavidovicNV: Should it be on by default? Or is this just for CI and this should be a draft?

skallweitNV: No, that was a mistake. But also revealed an issue with the format string :)

src/core/smart-pointer.cpp
skallweitNV: No, printf is just fine and already used in other places and for string formatting in general.
No, we are reporting diagnostics not an error here.

tdavidovicNV (approved):
LGTM

PR #473, opened 2025-08-07
CUDA sync fixes

skallweitNV:
Fix various issues with CUDA backend wrt. synchronization. See commit comments for details.

### DIFF
--- a/src/cuda/cuda-command.cpp
+++ b/src/cuda/cuda-command.cpp
@@ -471,14 +471,6 @@ void CommandExecutor::cmdDispatchCompute(const commands::DispatchCompute& cmd)
         CU_LAUNCH_PARAM_END,
     };
 
-    // This a big hammer! There are various scenarios where we have to synchronize the stream
-    // between kernel launches. One example is accessing texture memory: writes to texture memory (CUDA surface)
-    // are not guaranteed to be visible to the next launch reading from it (CUDA texture) without synchronization.
-    // We need to investigate if finer grained synchronization makes sense here. But this requires
-    // tracking texture states similarly to how it is done in D3D12/Vulkan.
-    // Note: This *does not* block the host unless the context is set to blocking mode.
-    cuStreamSynchronize(m_stream);
-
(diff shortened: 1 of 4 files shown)


src/cuda/cuda-device.cpp
```diff
@@ -549,7 +549,7 @@ Result DeviceImpl::readTexture(
 {
     SLANG_CUDA_CTX_SCOPE(this);
 
-    cuStreamSynchronize(m_queue->m_stream);
+    SLANG_CUDA_RETURN_ON_FAIL_REPORT(cuStreamSynchronize(m_queue->m_stream), this);
```
ccummingsNV: Is this needed now the command buffer thing is fixed?

src/cuda/cuda-device.cpp
```diff
@@ -581,7 +581,7 @@ Result DeviceImpl::readBuffer(IBuffer* buffer, size_t offset, size_t size, void*
...
```
ccummingsNV: Same question as for texture

src/cuda/cuda-command.cpp
```diff
@@ -834,7 +825,7 @@ Result CommandQueueImpl::submit(const SubmitDesc& desc)
         SLANG_RETURN_ON_FAIL(executor.execute(commandBuffer));
 
         // Only record command buffer if executor succeeds and we correctly add it to the active stream
-        SLANG_CUDA_RETURN_ON_FAIL(cuEventCreate(&commandBuffer->m_completionEvent, 0));
+        SLANG_CUDA_RETURN_ON_FAIL(cuEventCreate(&commandBuffer->m_completionEvent, CU_EVENT_DISABLE_TIMING));
```
ccummingsNV, line 828: Guessing this is just a perf thing?

skallweitNV: yes, we don't need timing, just synchronization

src/cuda/cuda-device.cpp
skallweitNV: I think in theory we don't.

PR #476, opened 2025-08-08
Fix sphere and LSS tests under D3D12 and Vulkan

ksavoie-nv:
I noticed some tests were failing when I ran the test suite on a card that supports LSS and spheres in D3D12 and Vulkan. This PR fixes tests wherever possible and disables the hit object tests under D3D12, because they're blocked by a bug in Slang.

### DIFF
--- a/src/d3d12/d3d12-pipeline.cpp
+++ b/src/d3d12/d3d12-pipeline.cpp
@@ -676,6 +676,22 @@ Result DeviceImpl::createRayTracingPipeline2(const RayTracingPipelineDesc& desc,
             m_nvapiShaderExtension.uavSlot,
             m_nvapiShaderExtension.registerSpace
         ));
+
+        if (is_set(desc.flags, RayTracingPipelineFlags::EnableLinearSweptSpheres) ||
+            is_set(desc.flags, RayTracingPipelineFlags::EnableSpheres))
+        {
+            NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS params = {};
+            params.version = NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS_VER;
+
+            if (is_set(desc.flags, RayTracingPipelineFlags::EnableLinearSweptSpheres))
+                params.flags = NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_LSS_SUPPORT;
+            if (is_set(desc.flags, RayTracingPipelineFlags::EnableSpheres))
+                params.flags = NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_SPHERE_SUPPORT;
+
+            SLANG_RHI_NVAPI_RETURN_ON_FAIL(
+                NvAPI_D3D12_SetCreatePipelineStateOptions((ID3D12Device5*)m_device, &params)
+            );
+        }
     }
 #endif // SLANG_RHI_ENABLE_NVAPI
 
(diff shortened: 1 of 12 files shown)


src/d3d12/d3d12-pipeline.cpp
```diff
@@ -676,6 +676,22 @@ Result DeviceImpl::createRayTracingPipeline2(const RayTracingPipelineDesc& desc,
...
+            NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS params = {};
+            params.version = NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS_VER;
+
+            if (is_set(desc.flags, RayTracingPipelineFlags::EnableLinearSweptSpheres))
+                params.flags = NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_LSS_SUPPORT;
+            if (is_set(desc.flags, RayTracingPipelineFlags::EnableSpheres))
+                params.flags = NVAPI_D3D12_PIPELINE_CREATION_STATE_FLAGS_ENABLE_SPHERE_SUPPORT;
+
+            SLANG_RHI_NVAPI_RETURN_ON_FAIL(
+                NvAPI_D3D12_SetCreatePipelineStateOptions((ID3D12Device5*)m_device, &params)
```
ksavoie-nv, line 694: This setting is global to the device. Does slang-rhi support creating pipelines in multiple threads, and if so will this cause issues? I tried to find an API call to set this per-pipeline, but didn't see a way to do so.

src/d3d12/d3d12-pipeline.cpp
```diff
@@ -689,6 +705,17 @@ Result DeviceImpl::createRayTracingPipeline2(const RayTracingPipelineDesc& desc,
...
         SLANG_RHI_NVAPI_RETURN_ON_FAIL(NvAPI_D3D12_SetNvShaderExtnSlotSpace(m_device, 0xffffffff, 0));
+
+        // LSS support is global and has a perf impact on all pipelines, so we should disable it.
+        if (is_set(desc.flags, RayTracingPipelineFlags::EnableLinearSweptSpheres))
+        {
+            NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS params = {};
+            params.version = NVAPI_D3D12_SET_CREATE_PIPELINE_STATE_OPTIONS_PARAMS_VER;
+
+            SLANG_RHI_NVAPI_RETURN_ON_FAIL(
+                NvAPI_D3D12_SetCreatePipelineStateOptions((ID3D12Device5*)m_device, &params)
```
ksavoie-nv, line 718: This clears the settings globally. If we try to set them somewhere else, it could mess with them. I wanted to query the settings and restore them after the call, but didn't see a way to do so.

jkwak-work:
Is this related to this issue?
- https://github.com/shader-slang/slang/issues/8128

ksavoie-nv:
> Is this related to this issue?
> 
> * [LSS and sphere hit object intrinsics fail to compile slang#8128](https://github.com/shader-slang/slang/issues/8128)

Yes. This PR disables two tests that are affected by that issue.

src/d3d12/d3d12-pipeline.cpp
jkwak-work: For our testing framework, I think this is fine.
We run multiple test-server.exe, each of which loads render-test.dll which uses slang-rhi.
It should be fine when the setting is global to each process.

For any user applications that may run multiple threads of slang-rhi, I am not sure if it will cause a problem.
We may need to add a new API function to slang-rhi so that the application can control the behavior?

skallweitNV: slang-rhi currently does not support multi-threading. We plan to add support down the road, but there are a bunch of areas that need work first.

src/d3d12/d3d12-pipeline.cpp
skallweitNV: Thats fine for now.

ksavoie-nv:
> @ksavoie-nv I took the liberty and rebased this to `main` and added a few changes:
> 
> * Handling of NVAPI descriptor was a bit more intricate and I implemented in two separate PRs including tests (merged to `main`). The code for handling the NVAPI slot in this PR was removed.
> * Fixed a few Vulkan validation layer issues.
> 
> Other than that, this PR looks good. Maybe you can go over my changes and take a look?

LGTM, and the tests all pass for me with the rebase + your changes. Thanks!

PR #478, opened 2025-08-11
Fix vulkan validation errors

skallweitNV:


skallweitNV:
merging as self-reviewed

PR #479, opened 2025-08-12
Support implicit NVAPI use

skallweitNV:


skallweitNV:
merging as self-reviewed

PR #481, opened 2025-08-12
Consistently use "WebGPU" instead of "wgpu"

samestep:
The slang-rhi API and also https://github.com/shader-slang/slang use "wgpu" to refer to WebGPU, which is confusing because others do not use that term, because [**wgpu** is a different thing](https://wgpu.rs/). To get rid of that confusion, this PR changes this repo to consistently use the correct term "WebGPU".


samestep:
@skallweitNV this is just a draft PR to see if it'd be easy to do. For discussion on whether this makes sense at all (which I agree is unclear), probably better to do so in shader-slang/slang#8157.

Out of curiosity, though, what's the issue with renaming files specifically?

skallweitNV:
I just prefer having inline changes and file renaming in separate PRs (at least for big renamings like this). It makes it much easier to see what changed.

samestep:
Ah never mind, I see now: while GitHub does identify some of the file renamings in this PR, it doesn't identify all of them. Apologies for not noticing that earlier.

PR #485, opened 2025-08-15
Dump live objects with class names if RTTI data is available

skallweitNV:
- Improve the information dumped in `reportLifeObjects`
- Minor internal API cleanup


skallweitNV:
merging as self-reviewed

PR #490, opened 2025-08-21
Delayed freeing

ccummingsNV:


src/core/smart-pointer.h
```diff
@@ -124,7 +124,8 @@ class SLANG_RHI_API RefObject
         if (count == 1)
         {
             // Last reference, delete the object
-            delete this;
+            // Default behavior immediately calls delete
+            markForDeletion();
```
skallweitNV: Instead of inheritance we a "deleter" callback, something like:

```c++
// deleter callback
class RefObject
{
    using DeleterFunc = void(*)(RefObject* object);

    DeleterFunc m_deleter = nullptr;

    ...
    if (count == 1)
    {
        if (m_deleter) m_deleter(this) else delete this;
     }
};
```

Might be more flexible, but just an idea ..

ccummingsNV: I guess the issue (which may be good or bad) is this gives every object, rather than every type, the ability to have a bespoke deleter. So we pay the 8 byte overhead for every object, rather than just an extra entry in the vtable which already existed.

skallweitNV: Yep, we might not need it. I don't really like the `markForDeletion` name which default implementation actually deletes the object. Maybe `destroy` or `destroySelf` would be more appropriate?

src/heap.h
```diff
@@ -85,8 +86,10 @@ class Heap : public IHeap, public DeviceChild
 
 public:
     HeapDesc m_desc;
+    StructHolder m_descHolder;
     uint32_t m_nextPageId = 1;
 
+
```
skallweitNV, line 98: extra newline

src/device.cpp
```diff
@@ -956,4 +956,5 @@ Result Device::getShaderObjectLayout(
     return SLANG_OK;
 }
 
+
```
skallweitNV: extra newline

src/debug-layer/debug-helper-functions.cpp
```diff
@@ -147,6 +147,12 @@ std::string createRayTracingPipelineLabel(const RayTracingPipelineDesc& desc)
     return std::string("Unnamed ray tracing pipeline");
 }
 
+std::string createHeapLabel(const HeapDesc& desc)
+{
+    return string::format("Unnamed heap (memoryType=%s)", enumToString(desc.memoryType));
+}
+
+
```
skallweitNV: extra newline

src/cuda/cuda-device.cpp
```diff
@@ -416,9 +418,24 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
...
-    SLANG_RETURN_ON_FAIL(m_clearEngine.initialize(m_debugCallback));
+    // Create 2 heaps. On CUDA both Upload and ReadBack just use host memory,
+    // so we only need one for DeviceLocal and one for Upload/ReadBack.
+    ComPtr<IHeap> heapPtr;
+    HeapDesc heapDesc = {};
+
+    heapDesc.memoryType = MemoryType::Upload;
+    heapDesc.label = "Device local heap";
+    SLANG_RETURN_ON_FAIL(createHeap(heapDesc, heapPtr.writeRef()));
+    m_hostMemHeap = checked_cast<HeapImpl*>(heapPtr.get());
```
skallweitNV, line 429: should we call this `m_uploadHeap` instead?

src/cuda/cuda-device.cpp
```diff
@@ -416,9 +418,24 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
...
+    m_hostMemHeap->breakStrongReferenceToDevice();
 
-    SLANG_RETURN_ON_FAIL(m_dualPageAllocator.init(this));
+    heapDesc.memoryType = MemoryType::DeviceLocal;
+    heapDesc.label = "Device upload heap";
+    SLANG_RETURN_ON_FAIL(createHeap(heapDesc, heapPtr.writeRef()));
+    m_localMemHeap = checked_cast<HeapImpl*>(heapPtr.get());
```
skallweitNV: `m_deviceHeap` or similar?

src/cuda/cuda-constant-buffer-pool.h
```diff
@@ -16,20 +15,25 @@ class ConstantBufferPool
...
+    // the global device heap, which eventually handles small allocations.
     static constexpr size_t kAlignment = 64;
-    static constexpr size_t kPageSize = 4 * 1024 * 1024;
+    static constexpr size_t kPageSize = 128 * 1024;
     static_assert(kPageSize % kAlignment == 0, "Page size must be a multiple of alignment");
 
     struct Page
     {
-        RefPtr<DualPageAllocator::Handle> handle;
+        HeapAlloc deviceMem;
```
skallweitNV, line 35: `deviceMemory`?

src/cuda/cuda-constant-buffer-pool.h
```diff
@@ -16,20 +15,25 @@ class ConstantBufferPool
...
+        HeapAlloc hostMem;
```
skallweitNV, line 36: `uploadMemory`?

src/cuda/cuda-command.cpp
```diff
@@ -755,24 +755,68 @@ CommandQueueImpl::~CommandQueueImpl()
...
             // We aren't recycling, so all we have to do is destroy the event
-            SLANG_CUDA_ASSERT_ON_FAIL(cuEventDestroy(commandBuffer->m_completionEvent));
-            m_submitCompleted = commandBuffer->m_submitIndex;
-
-            // Reset the command buffer for reuse.
-            commandBuffer->reset();
+            SLANG_CUDA_ASSERT_ON_FAIL(cuEventDestroy(submitIt->event));
+            m_lastFinishedID = submitIt->submitID;
 
             // Remove the command buffer from the list.
```
skallweitNV: this is not a command buffer

src/cuda/cuda-command.cpp
```diff
@@ -832,12 +867,17 @@ Result CommandQueueImpl::submit(const SubmitDesc& desc)
...
 
         // Increment submit count
-        m_submitCount++;
+        m_lastSubmittedID++;
 
-        // Only record command buffer if executor succeeds and we correctly add it to the active stream
-        SLANG_CUDA_RETURN_ON_FAIL(cuEventCreate(&commandBuffer->m_completionEvent, CU_EVENT_DISABLE_TIMING));
-        SLANG_CUDA_RETURN_ON_FAIL(cuEventRecord(commandBuffer->m_completionEvent, requestedStream));
-        commandBuffer->m_submitIndex = m_submitCount;
+        // Record submission event so we can detect completion
```
skallweitNV: Can't you reuse the `signalFence` function here?

skallweitNV (commented):
LGTM, some rather minor comments

src/cuda/cuda-device.cpp
ccummingsNV: I think not, because it serves as both upload and readback. The accurate term is host mem.

src/cuda/cuda-constant-buffer-pool.h
ccummingsNV: I think on CUDA I'll go with generally referring to device and host memory, as it's more accurate.

src/cuda/cuda-device.cpp
skallweitNV: Ah I see, you're sharing the heap for both upload and readback. Guess that's fine then. I still feel `MemHeap` is strange when the class is called just `Heap` now. I'd go for `m_deviceHeap` and `m_hostHeap` but no biggie.

skallweitNV (approved):

PR #493, opened 2025-08-22
Propagate range offset from ParameterBlock

jhelferty-nv:
ParameterBlocks have binding offsets associated with them that need to be propagated to their subobjects.

Fixes https://github.com/shader-slang/slang/issues/7611


src/d3d12/d3d12-shader-object-layout.cpp
```diff
@@ -845,6 +845,7 @@ void RootShaderObjectLayoutImpl::RootSignatureDescBuilder::addAsValue(
             offsetForChildrenThatNeedNewSpace += BindingRegisterOffsetPair(elementVarLayout);
             BindingRegisterOffsetPair offsetForOrindaryChildren = subDescriptorSetOffset;
             offsetForOrindaryChildren += BindingRegisterOffsetPair(containerVarLayout);
+            offsetForOrindaryChildren += inElementOffset;
```
jhelferty-nv, line 848: It feels like if this was missing, then maybe there's other stuff that should also be propagated? I'm really not familiar enough with this section of the code, though. My gut feeling is that the `BindingType::ConstantBuffer` case on line 799 and this case should share more DNA.

PR #496, opened 2025-08-25
Fix validateAccelerationStructureBuildDesc

### LINKED ISSUE #495: (Incorrectly) unreachable code in debug helper function
Looks like `break;` on line 270 in `slang-rhi/src/debug-layer/debug-helper-functions.cpp` makes the if-cuda block that follow unreachable. I suspect this break shouldn't be there.

https://github.com/shader-slang/slang-rhi/blob/7eb8bd0f4bbbbe4323badb94e96eb241fca2cc01/src/debug-layer/debug-helper-functions.cpp#L262-L287

skallweitNV:
Return an error if validation failed. Propagate error in getAccelerationStructureSizes

Fixes #495

### DIFF
--- a/src/debug-layer/debug-device.cpp
+++ b/src/debug-layer/debug-device.cpp
@@ -438,7 +438,7 @@ Result DebugDevice::getAccelerationStructureSizes(
 )
 {
     SLANG_RHI_API_FUNC;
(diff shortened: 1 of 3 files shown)


jkwak-work (dismissed):
Looks good to me

PR #499, opened 2025-08-28
Vulkan heap

ccummingsNV:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,56 @@
+Act as a senior C++ developer, with extensive experience in low graphics libraries such as Vulkan, Direct 3D 12 and Cuda.
```
tdavidovicNV: Should this be part of this PR or a separate one (to track things)?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,56 @@
...
+
+Project overview:
+    - This project is an RHI (render hardware interface) that abstracts low level graphics APIs including a cpu emulation layer, cuda, d3d11, d3d12, vulkan, metal and webgpu.
+    - The project is cross platform and supports windows, linux, macos.
+    - The codebase is c++17 and uses cmake as the build system.
+    - Shaders are written in the slang shading language.
```
tdavidovicNV: I don't know how case sensitive these tools are, but isn't it called `Slang` in most of the sources it can find for it?

src/vulkan/vk-heap.cpp
```diff
@@ -0,0 +1,184 @@
...
+    if (api.m_extendedFeatures.vulkan12Features.bufferDeviceAddress)
+    {
+        usage |= VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT;
+    }
+
+    // Determine memory properties based on heap memory type
+    VkMemoryPropertyFlags reqMemoryProperties;
+    HeapImpl* heapImpl = static_cast<HeapImpl*>(heap);
+
+    if (heapImpl->m_desc.memoryType == MemoryType::DeviceLocal)
```
tdavidovicNV: I'd do switch

tdavidovicNV (dismissed):
This looks fine to me, it just does a bit more on the AI support side than I imagined

.github/copilot-instructions.md
skallweitNV: Agreed, this should probably be a separate PR.

tests/test-heap.cpp
```diff
@@ -331,6 +331,317 @@ GPU_TEST_CASE("heap-pointer-stress-test", CUDA)
...
+    uint64_t bStart = b.getDeviceAddress();
+    uint64_t bEnd = bStart + b.size;
+
+    return !(aEnd <= bStart || bEnd <= aStart);
+}
+
+GPU_TEST_CASE("heap-no-overlaps", CUDA | Vulkan)
+{
+    HeapDesc desc;
+    desc.label = "Test Graphics Heap";
```
skallweitNV: I would not add a label here. In debug we get auto-labels. (I think we don't generally set labels in tests, do we?)

src/vulkan/vk-heap.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "vk-base.h"
+#include "vk-buffer.h"
+#include "../heap.h"
+
+#include <mutex>
+
+namespace rhi::vk {
+
+class DeviceImpl;
```
skallweitNV: This fwd declaration is not needed when including `vk-base`

.github/copilot-instructions.md
ccummingsNV: Sure - I'll split it out

src/vulkan/vk-heap.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+HeapImpl::PageImpl::PageImpl(Heap* heap, const PageDesc& desc, DeviceImpl* device)
+    : Heap::Page(heap, desc)
+    , m_device(device)
+{
+    // Initialize buffer similar to BufferImpl::createBuffer logic
+    const VulkanApi& api = device->m_api;
+
+    // Determine buffer usage flags for heap pages
+    VkBufferUsageFlags usage =
```
skallweitNV, line 17: So this means the heap cannot be used for textures etc. I think that can be added later, or do we not want heaps to be usable for other resources?

src/vulkan/vk-heap.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+DeviceAddress HeapImpl::PageImpl::offsetToAddress(Size offset)
+{
+    if (!m_device->m_api.vkGetBufferDeviceAddress)
+        return 0;
+
+    VkBufferDeviceAddressInfo info = {};
+    info.sType = VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO;
+    info.buffer = m_buffer.m_buffer;
+    return (DeviceAddress)(m_device->m_api.vkGetBufferDeviceAddress(m_device->m_api.m_device, &info) + offset);
```
skallweitNV, line 73: Don't we want to call `vkGetBufferDeviceAddress` upfront and cache the address?

src/vulkan/vk-heap.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // Create page - the constructor will handle all buffer creation logic
+    Page* page = new PageImpl(this, desc, deviceImpl);
+
+    // Vulkan memory allocation guarantees an alignment suitable for
+    // any memory type, however if the user asks for an alignment
+    // higher than that, the page may end up misaligned. For now,
+    // treat this as an error, as the user should only ever be asking
+    // for alignments based on correct memory requirements.
+    if (page->offsetToAddress(0) % desc.alignment != 0)
```
skallweitNV, line 146: Do we know what the alignments are? because then we could handle this in `fixUpAllocDesc` where we already check for power-of-two alignment. Then here we could just have an assert.

ccummingsNV: Annoyingly not - I did some digging and it's device dependent. The only guaruntee you have is that the memory allocation will be of a high enough alignment for any buffer or texture configuration. There's no way in vulkan (as far as I can tell) of asking what that is, other than creating buffers/textures with lots of different configurations and asking what the alignment requirements are. If you see a way though I'd rather use it - I might have just not found it in the docs.

src/vulkan/vk-heap.cpp
ccummingsNV: Can do - this is what vk-buffer does but I'm easy either way.

src/vulkan/vk-heap.cpp
ccummingsNV: In this PR, it's buffers only. In the next PR I'm aiming to separate out the concept of memory+buffer, so (hopefully) we can create buffers and textures direct from a heap page. The page will still need a buffer object if it wants  to return arbitrary pointers, but other than that it can be used purely for the memory handle. Need to dig into memory types settings first though.

src/vulkan/vk-heap.cpp
ccummingsNV: (I'll do this in next PR though, as I'm reworking a load of bits here)

skallweitNV (approved):
LGTM

PR #504, opened 2025-09-01
CUDA CB pool has separate pages for global vs ep data

ccummingsNV:


src/cuda/cuda-constant-buffer-pool.cpp
```diff
@@ -28,48 +33,67 @@ void ConstantBufferPool::upload(CUstream stream)
...
 
 void ConstantBufferPool::reset()
+{
+    m_entryPointDataPool.reset(m_device);
+    m_globalDataPool.reset(m_device);
+}
+
+Result ConstantBufferPool::allocate(size_t size, bool global, Allocation& outAllocation)
+{
+    if (global)
```
skallweitNV: `global` could be an enum instead of a boolean .. would make things a bit more readable (especially at the call site)

src/cuda/cuda-constant-buffer-pool.h
```diff
@@ -37,15 +37,24 @@ class ConstantBufferPool
...
+    struct Pool
+    {
+        std::vector<Page> m_pages;
+        std::vector<Page> m_largePages;
 
-    std::vector<Page> m_pages;
-    std::vector<Page> m_largePages;
+        int m_currentPage = -1;
+        size_t m_currentOffset = 0;
+        bool m_host;
```
skallweitNV: You could share the same enum here instead of two bools .. `m_device` is a bad name here IMO

skallweitNV (approved):
LGTM

PR #505, opened 2025-09-01
Command buffer cleanup

skallweitNV:
- Cleanup command buffer reuse in D3D12/Vulkan
- Implement command buffer reuse in CUDA
- Report total runtime in slang-rhi-tests

### DIFF
--- a/src/cuda/cuda-command.cpp
+++ b/src/cuda/cuda-command.cpp
@@ -716,19 +716,6 @@ void CommandExecutor::cmdExecuteCallback(const commands::ExecuteCallback& cmd)
 CommandQueueImpl::CommandQueueImpl(Device* device, QueueType type)
     : CommandQueue(device, type)
 {
-    SLANG_CUDA_CTX_SCOPE(getDevice<DeviceImpl>());
-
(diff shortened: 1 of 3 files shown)


src/vulkan/vk-command.cpp
```diff
@@ -1923,7 +1920,7 @@ Result CommandBufferImpl::reset()
 {
     DeviceImpl* device = getDevice<DeviceImpl>();
     m_commandList.reset();
-    SLANG_VK_RETURN_ON_FAIL(device->m_api.vkResetCommandBuffer(m_commandBuffer, 0));
+    SLANG_VK_RETURN_ON_FAIL(device->m_api.vkResetCommandPool(device->m_device, m_commandPool, 0));
```
ccummingsNV, line 1923: I don't know enough about vulkan command buffers to say if this switch to vkResetCommandPool is correct. I'm guessing it is, but just calling it out in case it needs double checking.

ccummingsNV (approved):

src/vulkan/vk-command.cpp
skallweitNV: It is and should be slightly more performant

PR #506, opened 2025-09-02
Report heaps api

ccummingsNV:


skallweitNV (commented):
I would consider implementing `reportHeaps` on the shared Device base class only and add a "internal heaps" list to the base class that is filled by the device implementations. This way the logic for querying is implemented once and guaranteed to work the same across all backends.

include/slang-rhi.h
```diff
@@ -2642,6 +2642,12 @@ class IHeap : public ISlangUnknown
     virtual SLANG_NO_THROW Result SLANG_MCALL removeEmptyPages() = 0;
 };
 
+struct HeapReport
+{
+    const char* name = nullptr;
+    IHeap::Report report = {};
```
skallweitNV: I think we're mostly not using nested structs, so maybe worth moving `Report` out of `IHeap` and rename to `HeapReport`.

include/slang-rhi.h
```diff
@@ -2642,6 +2642,12 @@ class IHeap : public ISlangUnknown
     virtual SLANG_NO_THROW Result SLANG_MCALL removeEmptyPages() = 0;
 };
 
+struct HeapReport
+{
+    const char* name = nullptr;
```
skallweitNV: Might be worth changing to `char name[128]`

include/slang-rhi.h
```diff
@@ -2642,6 +2642,12 @@ class IHeap : public ISlangUnknown
     virtual SLANG_NO_THROW Result SLANG_MCALL removeEmptyPages() = 0;
 };
 
+struct HeapReport
```
skallweitNV: Ah, just realized we now have `HeapReport` and `IHeap::Report` .. this is a bit confusing IMO

include/slang-rhi.h
```diff
@@ -2620,29 +2620,29 @@ struct HeapAllocDesc
     Size alignment = 0;
 };
 
+struct HeapReport
+{
+    char name[128] = {};
```
skallweitNV: Maybe rename this to `label` because it corresponds to the `label` in the desc.

include/slang-rhi.h
```diff
@@ -3269,6 +3269,18 @@ class IDevice : public ISlangUnknown
...
         uint32_t descCount
     ) = 0;
+
+    /// Report status of internal heaps used by the device.
+    /// If outHeapReports is null, returns the number of heaps in outHeapCount.
+    /// If outHeapReports is provided, fills up to bufferSize heap reports and returns actual count.
+    /// @param outHeapCount [out] Number of heaps available or written
+    /// @param outHeapReports [out] Buffer to write heap reports to (can be null for count query)
+    /// @param bufferSize [in] Size of the outHeapReports buffer (ignored if outHeapReports is null)
+    virtual SLANG_NO_THROW Result SLANG_MCALL reportHeaps(
```
skallweitNV: I would consider removing `bufferSize`.
Usually these APIs work like this:
```c++
uint32_t heapCount;
reportHeaps(&heapCount, nullptr);
std::vector<HeapReport> reports(heapCount);
reportHeaps(&heapCount, reports.data());
```

skallweitNV: There is this for example:
```c++
    virtual SLANG_NO_THROW Result SLANG_MCALL getCooperativeVectorProperties(
        CooperativeVectorProperties* properties,
        uint32_t* propertiesCount
    ) = 0;
```
I think the downside is while it should be `outProperties`, the `propertiesCount` is technically in/out ...

skallweitNV: We should settle on a style and make sure it's the same across the API.

src/device.cpp
```diff
@@ -910,6 +911,35 @@ Result Device::convertCooperativeVectorMatrix(const ConvertCooperativeVectorMatr
...
+    if (!heapReports)
+    {
+        *heapCount = totalHeapCount;
+        return SLANG_OK;
+    }
+
+    // If buffer is provided, it must be large enough
+    uint32_t bufferSize = *heapCount;
+    if (bufferSize < totalHeapCount)
+        return SLANG_E_BUFFER_TOO_SMALL;
```
skallweitNV:
```suggestion
    if (*heapCount < totalHeapCount)
        return SLANG_E_BUFFER_TOO_SMALL;
```

skallweitNV (approved):

PR #508, opened 2025-09-03
Heap fixes

ccummingsNV:
- Fix incorrect names on heaps
- Fix allocation of CB pages for 0 size allocs
- Add code to flush heaps when CB is retired

### DIFF
--- a/src/cuda/cuda-command.cpp
+++ b/src/cuda/cuda-command.cpp
@@ -811,6 +811,11 @@ Result CommandQueueImpl::retireCommandBuffers()
         cbIt = m_commandBuffersInFlight.erase(cbIt);
     }
 
+    // Flush the device heaps
+    DeviceImpl* device = getDevice<DeviceImpl>();
+    device->m_hostMemHeap->flush();
+    device->m_deviceMemHeap->flush();
+
     return SLANG_OK;
 }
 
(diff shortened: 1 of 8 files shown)


src/cuda/cuda-command.cpp
```diff
@@ -811,6 +811,10 @@ Result CommandQueueImpl::retireCommandBuffers()
         cbIt = m_commandBuffersInFlight.erase(cbIt);
     }
 
+    // Flush all device heaps
+    DeviceImpl* device = getDevice<DeviceImpl>();
+    SLANG_RETURN_ON_FAIL(device->flushHeaps());
```
tdavidovicNV: Why not just `SLANG_RETURN_ON_FAIL(getDevice<DeviceImpl>()->flushHeaps());`

src/d3d12/d3d12-command.cpp
```diff
@@ -1578,6 +1578,10 @@ void CommandQueueImpl::retireCommandBuffers()
             m_commandBuffersInFlight.push_back(commandBuffer);
         }
     }
+
+    // Flush all device heaps
+    DeviceImpl* device = getDevice<DeviceImpl>();
+    device->flushHeaps();
```
tdavidovicNV: Same question, the local variable seems totally superfluous

tdavidovicNV (approved):
I just don't see the point for the local `device` variable when you could just call on the `getDevice()` directly

src/cuda/cuda-command.cpp
ccummingsNV: Good point (was from debugging)

ccummingsNV:
[self-approve: minor tweaks for PR]

PR #509, opened 2025-09-04
Use GPU_TEST_CASE for all GPU tests

skallweitNV:
- change existing GPU tests using `TEST_CASE` to use `GPU_TEST_CASE`
- Add `GpuTestFlags::NoDevice` to allow tests to control device creation (with extra properties etc.)
- Pass `ctx` to GPU test functions
- Pass current device type in `GpuTestCtx`
- Remove `runGpuTests`

### DIFF
--- a/tests/test-buffer-shared.cpp
+++ b/tests/test-buffer-shared.cpp
@@ -3,11 +3,14 @@
 using namespace rhi;
 using namespace rhi::testing;
 
-template<DeviceType DstDeviceType>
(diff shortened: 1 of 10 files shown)


ccummingsNV (approved):
I find the 'NoDevice' api pretty confusing tbh - took me a while to realise it was an extra flag - initially I thought Vulkan | D3D12 | NoDevice meant 'run 3 versions of this test, with vulkan, then d3d12, then no device'.

Looks technically solid though.

PR #466, opened 2025-07-29
Update Dawn WebGPU

skallweitNV:
This is currently blocked by https://github.com/shader-slang/slang/issues/6571

PR #512, opened 2025-09-04
Fix Vulkan comparison sampler creation

jkwak-work:
On Vulkan, the sampler can be one of two types,
- depth compare
- color reduction (min/max/average).

You cannot set both depth compare and reduction.
For Vulkan, it is sufficient to just set `compareEnable` and `compareOp` values.

This fixes a regression on the following slang-test,
```
tests/metal/texture.slang.5 (vk)
```

Related to
- https://github.com/shader-slang/slang-rhi/pull/466


jkwak-work:
Closing, because this issue was already resolved in
- #468

PR #513, opened 2025-09-05
Add deferTargetCompilation to pipeline descs

skallweitNV:
This flag allows for deferred target code compilation at dispatch time. In the future, we plan to compile all target code needed for a command buffer in parallel. For now, compilation still happens single-threaded.


ccummingsNV (approved):
LGTM. It's not clear to me why you'd ask to defer compilation or not, but assuming it's necessary the PR makes sense.

PR #521, opened 2025-09-18
Disable the log callback before optixDeviceContextDestroy

gtong-nv:


skallweitNV:
So the issue here is that the debug callback is called during the `Device` destructor? I would assume this is something that we want to support, so is this to fix an issue in slang-rhi or for clients of slang-rhi?

gtong-nv:
Let's close this for now. This was a WAR to enable slang-rhi device cache in slang-test, but this is not a good fix.
I would need to make sure slang-test manages the callback lifetime better instead.

PR #522, opened 2025-10-01
Update HitObject.Invoke calls to remove AccelerationStructure parameter

szihs:
Remove the acceleration structure parameter from HitObject.Invoke calls in tests to match the corrected API signature.

Related to shader-slang/slang#7762

### DIFF
--- a/tests/test-ray-tracing-reorder.slang
+++ b/tests/test-ray-tracing-reorder.slang
@@ -32,7 +32,7 @@ void rayGenShaderReorderHint()
     uint numHintBits = 1;
     ReorderThread(hint, numHintBits);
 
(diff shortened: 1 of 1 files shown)


skallweitNV:
Let's not skip these tests but instead remove the AS parameter together with updating to a slang version that has the fix.

ksavoie-nv:
@szihs FYI I merged a few more tests that call HitObject.Invoke.

Also, I don't think you need to disable the tests when you remove the accel structure parameter from invoke in slang. The slang-rhi repo doesn't track slang ToT; it fetches a specific version. You should be able to remove the parameter without causing CI failures in slang-rhi, then we can bump the slang version and update the tests to remove the parameter in a separate PR.

PR #524, opened 2025-10-01
Enable running unit tests on github runners using WARP emulation

skallweitNV:
- Make D3D11 device creation more robust (allow debug layers to be absent)
- Enable debug layers based on `IRHI::enableDebugLayers` only (not on `DeviceDesc::enableValidation` which is for enabling RHI validation)
- Enable running D3D11/D3D12 unit tests on github runners through WARP
- Switch windows/aarch64 build to an actual aarch64 machine

### DIFF
--- a/tests/testing.cpp
+++ b/tests/testing.cpp
@@ -212,14 +212,7 @@ public:
         }
         else if (type == DebugMessageType::Warning)
         {
-            if (shouldIgnoreError(type, source, message))
-            {
-                output(msg);
-            }
-            else
-            {
-                FAIL(msg);
-            }
+            output(msg);
         }
         else if (type == DebugMessageType::Error)
         {


src/d3d11/d3d11-device.cpp
```diff
@@ -109,48 +109,52 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
...
         );
+        if (SUCCEEDED(result))
+            break;
     }
     if (FAILED(result))
     {
         printError("D3D11CreateDevice failed: %08x\n", result);
         return SLANG_FAIL;
     }
+    if (isDebugLayersEnabled() && (usedCreateFlags & UseDebug) == 0)
```
ccummingsNV, line 154: Not sure where the 'isDebugLayersEnabled' is used, but should it be set to false at this point, if it turned out not to be usable. i.e. so some other system doesn't check that bool and get told debug layers are on when they aren't. Or is this just from a define?

skallweitNV: Maybe the naming isn't great but `isDebugLayersEnabled` basically returns true if debug layers are requested by the user (by setting `IRHI::enableDebugLayers`). So we just try our best to enable them when requested.

PR #525, opened 2025-10-01
Fix selected defines for running tests

skallweitNV:
A bug from the adapter PR leads to no tests being run by default, this fixes it.


tdavidovicNV (approved):
LGTM

PR #529, opened 2025-10-06
Fix format conversion

skallweitNV:


src/format-conversion.h
```diff
@@ -6,134 +6,154 @@
...
 
 FormatConversionFuncs getFormatConversionFuncs(Format format);
 
 template<size_t N>
-inline void packInt8(const uint32_t in[4], void* out)
+inline void packUint8(const uint32_t in[4], void* out)
 {
     for (size_t i = 0; i < N; ++i)
-        reinterpret_cast<uint8_t*>(out)[i] = clamp(in[i], 0u, 255u);
+        reinterpret_cast<uint8_t*>(out)[i] = in[i] & 0xff;
```
tdavidovicNV, line 52: I could imagine asserting here if out of range.

src/format-conversion.h
```diff
@@ -6,134 +6,154 @@
...
         out[i] = reinterpret_cast<const uint8_t*>(in)[i];
 }
 
 template<size_t N>
-inline void packInt16(const uint32_t in[4], void* out)
+inline void clampUint8(uint32_t inout[4])
 {
     for (size_t i = 0; i < N; ++i)
-        reinterpret_cast<uint16_t*>(out)[i] = clamp(in[i], 0u, 65535u);
+        inout[i] = clamp(inout[i], 0u, 255u);
```
tdavidovicNV, line 66: Have you considered using `numeric_limits` instead of actual literals?

src/format-conversion.h
```diff
@@ -6,134 +6,154 @@
...
 }
 
 template<size_t N>
-inline void unpackInt16(const void* in, uint32_t out[4])
+inline void packSint8(const uint32_t in[4], void* out)
```
tdavidovicNV, line 70: Nothing checks that `N <= 4`?

skallweitNV: we could do a `static_assert` but usage seems very obvious, and these are not user-facing functions

src/format-conversion.h
skallweitNV: could, but this is not generic code so feels ok to me to have the hardcoded constants

src/format-conversion.h
skallweitNV: nope, the idea is to just use the relevant bits and ignore the rest

src/format-conversion.h
tdavidovicNV: Actually quite possible that the compiler would yell, trying to iterate past the array boundaries

PR #530, opened 2025-10-06
Detect adapter type

skallweitNV:
- Add `AdapterInfo::adapterType` reporting the adapter type (`Discrete`, `Integrated`, `Software`, `Unknown`)
- Select first discrete adapter by default (D3D11, D3D12, Vulkan)

### DIFF
--- a/src/d3d11/d3d11-device.cpp
+++ b/src/d3d11/d3d11-device.cpp
@@ -29,20 +29,8 @@ inline Result getAdaptersImpl(std::vector<AdapterImpl>& outAdapters)
 
     for (const auto& dxgiAdapter : dxgiAdapters)
     {
-        DXGI_ADAPTER_DESC desc;
-        dxgiAdapter->GetDesc(&desc);
-        AdapterInfo info = {};
+        AdapterInfo info = getAdapterInfo(dxgiAdapter);
         info.deviceType = DeviceType::D3D11;
-        info.adapterType = desc.DedicatedVideoMemory > 0 ? AdapterType::Discrete : AdapterType::Integrated;
-        if (desc.VendorId == 0x1414 && desc.DeviceId == 0x8c)
-        {
-            info.adapterType = AdapterType::Software;
-        }
-        auto name = string::from_wstring(desc.Description);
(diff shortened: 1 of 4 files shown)


src/d3d11/d3d11-device.cpp
```diff
@@ -33,6 +33,11 @@ inline Result getAdaptersImpl(std::vector<AdapterImpl>& outAdapters)
         dxgiAdapter->GetDesc(&desc);
         AdapterInfo info = {};
         info.deviceType = DeviceType::D3D11;
+        info.adapterType = desc.DedicatedVideoMemory > 0 ? AdapterType::Discrete : AdapterType::Integrated;
+        if (desc.VendorId == 0x1414 && desc.DeviceId == 0x8c)
```
jkwak-work: I understand that this is a direct copy-and-paste from the previous code.
But now with this change/refactor, the magic numbers became more magic.
Previously the code was implying that the number 0x1414 means WARP,
```
adapter.m_isWarp = (desc.VendorId == 0x1414 && desc.DeviceId == 0x8c);
```
And now it became less obvious.

I think it will be better if the code is more like the following,
```
constexpr const static int VENDOR_ID_MICROSOFT_WARP = 0x1414;
if (desc.VendorId == VENDOR_ID_MICROSOFT_WARP && ...
```

Just FYI, Codex suggested a little different approach, which I am not familiar with.
```
  ComPtr<IDXGIAdapter1> adapter1;
  dxgiAdapter.As(&adapter1);

  DXGI_ADAPTER_DESC1 desc1 = {};
  adapter1->GetDesc1(&desc1);

  bool isWarp = (desc1.Flags & DXGI_ADAPTER_FLAG_SOFTWARE) != 0;
  if (isWarp && ...)
```

skallweitNV: Good point, a comment would clear things up. I was considering using `IDXGIAdapter1` to check the SW flag, maybe I should just do that (but still keep a fallback and checking vendor/device ID).

skallweitNV: Done

PR #531, opened 2025-10-07
Improve CUDA dispatch

skallweitNV:


src/cuda/cuda-command.cpp
```diff
@@ -442,22 +442,28 @@ void CommandExecutor::cmdDispatchCompute(const commands::DispatchCompute& cmd)
...
+                computePipeline->m_globalParamsSize,
+                bindingData->globalParamsSize,
+                computePipeline->m_kernelName.c_str()
             );
+            computePipeline->m_warnedAboutGlobalParamsSizeMismatch = true;
         }
+        SLANG_CUDA_ASSERT_ON_FAIL(cuMemcpyAsync(
+            computePipeline->m_globalParams,
+            bindingData->globalParams,
+            min(bindingData->globalParamsSize, computePipeline->m_globalParamsSize),
```
ccummingsNV, line 464: This feels dodgy - if sometimes slang computes it incorrectly, does that mean one of the values is consistently correct and one isn't? And should it be an outright error for 1 way round?

skallweitNV: I found that slang computes the padding at the end incorrectly. Don't remember if slang reports padding that it shouldn't or vice-versa, so that's why I put the `min` there. In any case, this is a bug in Slang that needs to be fixed.

PR #532, opened 2025-10-07
Texture access tests

skallweitNV:
Adds a lot of tests for shader side load and store operations on textures.
Tests pass on D3D12/Vulkan/Metal but there issues with CUDA that need to fixed in Slang, so most of the CUDA tests are skipped for now.

### DIFF
--- a/tests/test-texture-view.cpp
+++ b/tests/test-texture-view.cpp
@@ -1,6 +1,23 @@
 #include "testing.h"
 #include "texture-test.h"
+#include "format-conversion.h"
 #include <map>
+
+// This series of tests tests reading/writing texture from shaders.
+// The following is covered:
+// - reading from textures using both Load() and the subscript operator
+//   - read-only textures:
+//     - all layers/all mips (Load() only)
+//     - all layers/single mip (Load() and subscript load operator)
+//     - single layer/single mip (Load() and subscript load operator)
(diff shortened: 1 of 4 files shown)


tests/test-texture-view.cpp
```diff
@@ -148,183 +172,1271 @@ static const std::vector<Format> kFormats = {
...
+    }
+}
 
-    struct PipelineKey
+static void compareTexelData(Format format, span<TexelData> a, span<TexelData> b)
+{
+    REQUIRE(a.size() == b.size());
+    const FormatInfo& info = getFormatInfo(format);
+
+    float atol[4] = {0.f, 0.f, 0.f, 0.f};
```
tdavidovicNV, line 224: I find this name rather confusing (see `std::atol`)

skallweitNV: yeah, but its a pretty common name for "absolute tolerance" .. what would be a better name in your opinion?

tdavidovicNV: I'd go with `tolerance`, or maybe `abs_err`.. but maybe a comment explaining the name would be enough. (fwiw, first time I am seeing that acronym like this)

skallweitNV: ok will rename this in the next PR enabling the CUDA tests

PR #539, opened 2025-10-13
Put more HitObject intrinsics under test

ksavoie-nv:
Add tests that exercise the HitObject. All but one of the new tests trace a ray to create a HitObject from a different program type, then invoke the resulting HitObject. The test `ray-tracing-hitobject-make-hit` calls `MakeHit` instead of tracing a ray. Note that the `MakeHit` test must be compiled as a separate module because the NVAPI headers fail to compile with newer shader models.

Other changes:
- Rename `test-ray-tracing-a` and `test-ray-tracing-b` to make it clearer what they're testing.
- Extend `DeviceExtraOptions` to allow specifying a maximum shader model for a test.

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -7,24 +7,19 @@
 
 #include <slang-rhi/capabilities.h>
 
-#if defined(SLANG_RHI_DYNAMIC)
+#if SLANG_RHI_SHARED
+#ifdef SLANG_RHI_SHARED_EXPORT
+#define SLANG_RHI_API SLANG_DLL_EXPORT
+#else
 #if defined(_MSC_VER)
-#ifdef SLANG_RHI_DYNAMIC_EXPORT
-#define SLANG_RHI_API SLANG_DLL_EXPORT
-#else
 #define SLANG_RHI_API __declspec(dllimport)
-#endif
 #else
-// TODO: need to consider compiler capabilities
-// #     ifdef SLANG_DYNAMIC_EXPORT
-#define SLANG_RHI_API SLANG_DLL_EXPORT
-// #     endif
-#endif
-#endif
-
-#ifndef SLANG_RHI_API
 #define SLANG_RHI_API
 #endif
+#endif
+#else // SLANG_RHI_SHARED
+#define SLANG_RHI_API
+#endif // SLANG_RHI_SHARED
 
 // Needed for building on cygwin with gcc
 #undef Always
@@ -3436,19 +3431,6 @@ public:
     virtual SLANG_NO_THROW Result SLANG_MCALL setTaskPool(ITaskPool* taskPool) = 0;
 };
 
-// Global public functions
-
-extern "C"
-{
-    /// Get the global interface to the RHI.
-    SLANG_RHI_API IRHI* SLANG_MCALL getRHI();
-}
-
-inline const FormatInfo& getFormatInfo(Format format)
-{
-    return getRHI()->getFormatInfo(format);
-}
-
 // Extended descs.
 struct D3D12ExperimentalFeaturesDesc
 {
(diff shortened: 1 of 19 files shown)


tests/testing.h
```diff
@@ -256,6 +256,11 @@ struct DeviceExtraOptions
     IPersistentCache* persistentPipelineCache = nullptr;
     bool enableCompilationReports = false;
     DeviceNativeHandles existingDeviceHandles;
+
+    // D3D12-specific (no effect for other devices): Limit the maximum shader model. When set to 0
+    // (default), uses automatic detection.
+    // This value is passed to D3D12DeviceExtendedDesc::highestShaderModel.
+    uint32_t d3d12MaxShaderModel = 0;
```
skallweitNV: Maybe `d3d12HighestShaderModel` would be more in line with the underlying parameter name.

ksavoie-nv: @skallweitNV struct field renamed

include/slang-rhi.h
```diff
@@ -2942,6 +2942,8 @@ class IDevice : public ISlangUnknown
     virtual SLANG_NO_THROW bool SLANG_MCALL hasCapability(Capability capability) = 0;
     virtual SLANG_NO_THROW bool SLANG_MCALL hasCapability(const char* capability) = 0;
 
+    virtual SLANG_NO_THROW uint32_t SLANG_MCALL getOptixVersion() = 0;
```
skallweitNV: The optix version is already available in `DeviceInfo::optixVersion` so no need for this.

ksavoie-nv: Ah, I missed that. I removed the new method and changed the guard to use the device info instead.

tests/test-ray-tracing-hitobject-intrinsics.cpp
```diff
@@ -0,0 +1,646 @@
...
+        const TestResult* result = reinterpret_cast<const TestResult*>(resultBlob->getBufferPointer());
+        CHECK_EQ(result->rayDirection[0], 0.0f);
+        CHECK_EQ(result->rayDirection[1], 0.0f);
+        CHECK_EQ(result->rayDirection[2], 1.0f);
+    }
+}
+
+GPU_TEST_CASE("ray-tracing-hitobject-make-hit", ALL | DontCreateDevice)
+{
+    // Limit the shader model to SM 6.6 for this test, since the NVAPI headers don't support MakeHit
```
skallweitNV, line 620: Curious, is this going to be fixed in a future version of NVAPI?

tests/test-ray-tracing-hitobject-intrinsics.cpp
```diff
@@ -0,0 +1,646 @@
...
+    device = createTestingDevice(ctx, ctx->deviceType, false, &extraOptions);
+    REQUIRE(device);
+
+    if (!device->hasFeature(Feature::RayTracing))
+        SKIP("ray tracing not supported");
+    if (!device->hasFeature(Feature::ShaderExecutionReordering))
+        SKIP("shader execution reordering not supported");
+
+
+    // Disabled under pre OptiX 9.0 due to https://github.com/shader-slang/slang/issues/8723
```
skallweitNV, line 633: There is at least one HitObject API (setting sbt index) that is not supported in OptiX 8.1, which otherwise supports most of the SER HitObject API. So the question will come if we just require OptiX 9.0 for SER.

skallweitNV (approved):

tests/test-ray-tracing-hitobject-intrinsics.cpp
ksavoie-nv: I couldn't say (I'm not involved with NVAPI development), but I wouldn't be surprised if they don't. MakeHit was mostly deprecated in DX, and the implementations in NVAPI look like they're for backward compatibility.

PR #552, opened 2025-10-17
System for synchronous cuda error checking

ccummingsNV:


src/cuda/cuda-utils.h
```diff
@@ -8,6 +8,11 @@
 /// This is useful for debugging to ensure that the CUDA context is set correctly when calling CUDA APIs.
 #define SLANG_RHI_ENABLE_CUDA_CONTEXT_CHECK 0
 
+/// Enable synchronous CUDA error checking by calling cuCtxSynchronize after each CUDA call and checking
```
skallweitNV, line 11: before and after

PR #559, opened 2025-10-29
Update slang-rhi for library rename and versioning

jhelferty-nv:
For shader-slang/slang#8828

Handle new filename patterns introduced by shader-slang/slang#4016

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,41 +261,37 @@ endif()
 set(SLANG_RHI_SLANG_INCLUDE_DIR ${SLANG_RHI_SLANG_INCLUDE_DIR} CACHE STRING "Slang include directory")
 set(SLANG_RHI_SLANG_BINARY_DIR ${SLANG_RHI_SLANG_BINARY_DIR} CACHE STRING "Slang binary directory")
 unset(SLANG_RHI_SLANG_INCLUDE_DIR)
-unset(SLANG_RHI_SLANG_BINARY_DIR)
 if(NOT SLANG_RHI_BUILD_FROM_SLANG_REPO)
     function(slang_rhi_add_library_slang)
         add_library(slang SHARED IMPORTED GLOBAL)
 
-        if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
-            set(slang_lib_dir "${SLANG_RHI_SLANG_BINARY_DIR}/bin")
-            set(slang_lib_prefix "")
-            set(slang_lib_suffix "dll")
+        # Slang renamed the compiler library from slang to slang-compiler in v2025.21
+        # Detect if the old or new library name is used.
+        if(CMAKE_SYSTEM_NAME STREQUAL "Windows" AND EXISTS ${SLANG_RHI_SLANG_BINARY_DIR}/bin/slang-compiler.dll)
+            set(SLANG_LIB_NAME slang-compiler)
+        elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND EXISTS ${SLANG_RHI_SLANG_BINARY_DIR}/lib/libslang-compiler.so)
+            set(SLANG_LIB_NAME slang-compiler)
+        elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND EXISTS ${SLANG_RHI_SLANG_BINARY_DIR}/lib/libslang-compiler.dylib)
+            set(SLANG_LIB_NAME slang-compiler)
         else()
-            set(slang_lib_dir "${SLANG_RHI_SLANG_BINARY_DIR}/lib")
-            set(slang_lib_prefix "lib")
-            if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -266,27 +266,43 @@ if(NOT SLANG_RHI_BUILD_FROM_SLANG_REPO)
     add_library(slang SHARED IMPORTED GLOBAL)
 
     if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+        if(EXISTS ${SLANG_RHI_SLANG_BINARY_DIR}/bin/slang-compiler.dll)
```
skallweitNV: Can we do this test once on `${SLANG_RHI_SLANG_BINARY_DIR}/bin/slang-compiler.*` or similar to make it work for all platforms?

CMakeLists.txt
```diff
@@ -266,27 +266,43 @@ if(NOT SLANG_RHI_BUILD_FROM_SLANG_REPO)
...
-            IMPORTED_LOCATION ${SLANG_RHI_SLANG_BINARY_DIR}/lib/libslang.dylib
+            IMPORTED_LOCATION ${SLANG_RHI_SLANG_BINARY_DIR}/lib/lib${SLANG_LIBBASE}.dylib
         )
     endif()
 
     target_link_libraries(slang-rhi PUBLIC slang)
 
     if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
         copy_file(${SLANG_RHI_SLANG_BINARY_DIR}/bin/slang.dll .)
+        copy_file(${SLANG_RHI_SLANG_BINARY_DIR}/bin/slang-compiler.dll .)
```
skallweitNV, line 305: Does one of the copies just silently fail here?

CMakeLists.txt
jhelferty-nv: I can, but even if we ignore the extension difference, the file's in a `bin` vs `lib` dir on windows, so we'll still need a per-platform `if` at the start, and then another at the end because `IMPORTED_IMPLIB` is windows-specific.

I can try it and see what it looks like.

CMakeLists.txt
jhelferty-nv: Yes. This is intentional, since I figured we'd want to land this change before the rename/versioning patch. Plus it gives us some flexibility, in the case of rollback, as well as a bit of a grace period for clean-up when we pull the `slang.dll` next year.

CMakeLists.txt
jhelferty-nv: Updated. let me know what you think

CMakeLists.txt
```diff
@@ -263,30 +263,50 @@ set(SLANG_RHI_SLANG_BINARY_DIR ${SLANG_RHI_SLANG_BINARY_DIR} CACHE STRING "Slang
...
+            else()
+                set(slang_lib_suffix "so")
+            endif()
+        endif()
+
+        if(EXISTS ${slang_lib_dir}/${slang_lib_prefix}slang-compiler.${slang_lib_suffix})
+            set(SLANG_LIBBASE slang-compiler)
+        else()
+            set(SLANG_LIBBASE slang)
+        endif()
```
skallweitNV: Not sure this is an improvement :) sorry, I thought cmake had an easy way to combine multiple `EXIST` expression into the same `if` but looks like it doesn't. So probably something like this would be the easiest on the eyes:
```
# Slang renamed the compiler library from slang to slang-compiler in version x.x.x
# Detect if the old or new library name is used.
if(EXISTS ${SLANG_RHI_SLANG_BINARY_DIR/bin/slang-compiler.dll)
    set(SLANG_LIB_NAME slang-compiler)
elseif(EXISTS ${SLANG_RHI_SLANG_BINARY_DIR/lib/slang-compiler.so)
    set(SLANG_LIB_NAME slang-compiler)
elseif(EXISTS ${SLANG_RHI_SLANG_BINARY_DIR/lib/slang-compiler.dylib)
    set(SLANG_LIB_NAME slang-compiler)
else()
    set(SLANG_LIB_NAME slang)
endif()
```

CMakeLists.txt
skallweitNV: I agree, we need to make sure slang-rhi works with both names flawlessly.

skallweitNV (changes requested):

CMakeLists.txt
jhelferty-nv: Done.

CMakeLists.txt
jhelferty-nv: As an aside, I had considered moving this windows-specific chunk into `slang_rhi_add_library_slang()` and using `SLANG_LIB_NAME` to only copy the one we're using over, but I've kept it copying both (whichever exist) so that both end up in the destination. I'm not sure if the destination is actually used by anything downstream, but it seems useful to keep the specific `slang.dll` proxy library that goes with the `slang-compiler.dll` alongside it, in case someone needs to use both of them in a project.

skallweitNV (approved):

PR #560, opened 2025-10-31
Add test for GetHitKind and refactor ray tracing tests

ksavoie-nv:
# New Tests

Add three new tests to exercise `GetHitKind`:
- ray-tracing-hitobject-query-hit-kind-front-face
- ray-tracing-hitobject-query-hit-kind-back-face
- ray-tracing-hitobject-query-hit-kind-custom

These tests exercise `GetHitKind` with triangle and custom geometry under D3D12 and Vulkan. They don't exercise OptiX, as `GetHitKind` hasn't been implemented for it.

# Refactor

It's become clear that most tests create simple acceleration structures and shader tables, and that much of their code can be shared. Pull this common code into `test-ray-tracing-common.h`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,12 +92,12 @@ jobs:
 
       - name: Unit Tests (OptiX 8.0)
         if: contains(matrix.flags, 'unit-test')
-        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -optix-version=80000
+        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -check-devices -optix-version=80000
         working-directory: build/${{matrix.config}}
 
       - name: Unit Tests (OptiX 8.1)
         if: contains(matrix.flags, 'unit-test')
-        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -optix-version=80100
+        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -check-devices -optix-version=80100
         working-directory: build/${{matrix.config}}
 
       - name: Coverage Report
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -1549,8 +1549,8 @@ struct Binding
     Binding(IBuffer* buffer, IBuffer* counter, const BufferRange& range = kEntireBuffer) : type(BindingType::BufferWithCounter), resource(buffer), resource2(counter), bufferRange(range) {}
     Binding(const ComPtr<IBuffer>& buffer, const ComPtr<IBuffer>& counter, const BufferRange& range = kEntireBuffer) : type(BindingType::BufferWithCounter), resource(buffer), resource2(counter), bufferRange(range) {}
 
-    Binding(ITexture* texture) : type(BindingType::Texture), resource(texture->getDefaultView()) {}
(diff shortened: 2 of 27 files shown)


skallweitNV (changes requested):
Looks good from a first glance, but I would probably not introduce the `tests/ray-tracing` folder, we don't have this for any of the other tests either and unnecessarily confuses the git diff. I don't see what changed in `test-ray-tracing-lss.cpp` for example, just old one removed, new one added.

ksavoie-nv:
> Looks good from a first glance, but I would probably not introduce the `tests/ray-tracing` folder, we don't have this for any of the other tests either and unnecessarily confuses the git diff. I don't see what changed in `test-ray-tracing-lss.cpp` for example, just old one removed, new one added.

Agreed, that diff was pretty gnarly. I moved the tests back out of the subdirectory.

tests/test-ray-tracing-common.h
```diff
@@ -0,0 +1,734 @@
...
+//
+// This header provides classes and functions that provide these common building blocks.
+
+namespace rhi::testing {
+struct Vertex
+{
+    float position[3];
+};
+
+struct TriangleBlas
```
skallweitNV: minor naming conventions stuff (no need to change):
- for class names we should probably use `BLAS` and `TLAS`, so `TriangleBLAS` here
- for members and variables it would be `blas` and `tlas`, so `blasBuffer` instead of `BLASBuffer`

tests/test-ray-tracing-common.cpp
```diff
@@ -0,0 +1,69 @@
+#include "test-ray-tracing-common.h"
+
+namespace rhi::testing {
+
+Result loadShaderPrograms(
```
skallweitNV, line 5: Is there a similar function in `testing.h` that could be used? Seems like this is not necessarily raytracing test specific.

tests/test-ray-tracing-common.cpp
```diff
@@ -0,0 +1,69 @@
...
+    SLANG_RETURN_ON_FAIL(result);
+
+    ShaderProgramDesc programDesc = {};
+    programDesc.slangGlobalScope = linkedProgram;
+    SLANG_RETURN_ON_FAIL(device->createShaderProgram(programDesc, outProgram));
+
+    return SLANG_OK;
+}
+
+void launchPipeline(
```
skallweitNV, line 48: If this is the only remaining thing in this `cpp` I would just inline it in the header file and get rid of the cpp.

skallweitNV (approved):
Some minor propositions but feel free to merge as is.

tests/test-ray-tracing-common.cpp
ksavoie-nv: There are two very similar functions (`loadGraphicsProgram` and `loadComputeProgram`), but neither of them take an arbitrary number of entry points.

I suspect all three versions could be replaced by a single function with this signature:
```
Result loadProgram(
    IDevice* device,
    ComPtr<IShaderProgram>& outShaderProgram,
    const char* shaderModuleName,
    span<const char*> entryPointNames,
    slang::ProgramLayout*& slangReflection
);
```

I could refactor them in this PR, but I think I'd prefer to do that separately if we're interested.

skallweitNV: Yes, separate PR is better.

skallweitNV (approved):

PR #564, opened 2025-11-04
Add CUDA_PATH/bin/x64 to search path for nvrtc

gtong-nv:
Since CUDA 13.0, cuda dll have been moved to %CUDA_PATH%\bin\x64
This PR should fix https://github.com/shader-slang/slangpy/issues/614

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,12 +92,12 @@ jobs:
 
       - name: Unit Tests (OptiX 8.0)
         if: contains(matrix.flags, 'unit-test')
-        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -optix-version=80000
+        run: ./slang-rhi-tests -tc=ray-tracing*.cuda -check-devices -optix-version=80000
         working-directory: build/${{matrix.config}}
 
       - name: Unit Tests (OptiX 8.1)
         if: contains(matrix.flags, 'unit-test')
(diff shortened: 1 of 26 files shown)


src/cuda/cuda-nvrtc.cpp
```diff
@@ -58,6 +58,12 @@ inline void findNVRTCPaths(std::vector<std::filesystem::path>& outPaths)
             {
                 outPaths.push_back(path);
             }
+            // Since CUDA 13.0, cuda dll have been moved to bin\x64
+            std::filesystem::path pathX64 = version / "bin" / "x64";
```
skallweitNV, line 62: What about ARM on windows and linux? Now that we ship at least aarch64 packages for linux we should ensure that this also works there.

gtong-nv: I need to check that. Will update tomorrow

gtong-nv: In Linux x86, the cuda libs are located in 

/usr/local/cuda-12.8/targets/x86_64-linux/lib
/usr/local/cuda-13.0/targets/x86_64-linux/lib

13.0 does seem to change it

gtong-nv: in Linux aarch64, 
it's `/usr/local/cuda-12.6/targets/aarch64-linux/lib/`.
I don't have access to Linux aarch64 with the latest cuda version for now. But I assume the cuda lib path is not changed

PR #568, opened 2025-11-05
CombinedTextureSampler support in CUDA

skallweitNV:


src/cuda/cuda-sampler.h
```diff
@@ -4,11 +4,30 @@
...
+    CUfilter_mode filterMode;
+    unsigned int maxAnisotropy;
+    CUfilter_mode mipmapFilterMode;
+    float mipmapLevelBias;
+    float minMipmapLevelClamp;
+    float maxMipmapLevelClamp;
+    float borderColor[4];
+    bool operator==(const SamplerSettings& other) const
+    {
+        return std::memcmp(this, &other, sizeof(SamplerSettings)) == 0;
```
tdavidovicNV: This is technically wrong, because +0 and -0 in the floats will compare as `!=` even though it should be `==`.
I personally think that the convenience of this trumps the technicallity, but I've been in endless debates about this so I think we should all sign off on this being okay.

But we switch to C++20 only, this should be replaced with https://en.cppreference.com/w/cpp/language/default_comparisons.html so I think it is good enough for now no matter what.

src/cuda/cuda-shader-object.cpp
```diff
@@ -60,6 +60,18 @@ void shaderObjectSetBinding(
...
     }
+    case slang::BindingType::CombinedTextureSampler:
+    {
+        TextureViewImpl* textureView = checked_cast<TextureViewImpl*>(slot.resource.get());
+        SamplerImpl* sampler = checked_cast<SamplerImpl*>(slot.resource2.get());
+        uint64_t handle = 0;
+        if (textureView && sampler)
+        {
+            handle = textureView->getTexObjectWithSamplerSettings(sampler->m_samplerSettings);
+        }
```
tdavidovicNV, line 71: Shouldn't we report an error if either `textureView` or `sampler` is missing, rather than just silently not bind it?

tdavidovicNV (approved):
There is also some unexplained shuffling in the resources, I just assume it is a cleanup?

src/cuda/cuda-shader-object.cpp
skallweitNV: This is in line with other backends. We generally want to catch errors like that in the validation layer and not pay the cost when not needed.

tdavidovicNV: Good point

src/cuda/cuda-sampler.h
skallweitNV: Fixed

tdavidovicNV (approved):

PR #569, opened 2025-11-06
Unify functions for loading test shaders

ksavoie-nv:
Implement refactor from this discussion: https://github.com/shader-slang/slang-rhi/pull/560#discussion_r2491335436

There are currently multiple similar functions for loading shaders into tests. Combine them into a single function, `loadProgram`, which supports multiple entry points, linking, and fetching reflection data.

Expose that function via a few convenience overloads that allow callers to easily skip supplying a Slang session or getting reflection data.

Note that we pass entry points as a `std::vector`. That isn't great, as it requires dynamically allocating for each call, but it allows us to use brace initializers when we call the function.

Also, delete `test-ray-tracing-common.cpp`. It only had a single function after we unified the functions to load shader programs, so it makes sense to inline that function and delete the cpp file.

### DIFF
--- a/src/cpu/cpu-shader-object.cpp
+++ b/src/cpu/cpu-shader-object.cpp
@@ -171,31 +171,6 @@ Result BindingDataBuilder::writeObjectData(
             }
         }
         break;
-#if 0
-        case slang::BindingType::ExistentialValue:
-            // We can only bind information for existential-typed sub-object
-            // ranges if we have a static type that we are able to specialize to.
-            //
-            if (subObjectLayout)
-            {
-                // The data for objects in this range will always be bound into
-                // the "pending" allocation for the parent block/buffer/object.
-                // As a result, the offset for the first object in the range
-                // will come from the `pending` part of the range's offset.
-                //
-                SimpleBindingOffset objOffset = rangeOffset;
-                SimpleBindingOffset objStride = rangeStride;
-
-                for (uint32_t i = 0; i < count; ++i)
-                {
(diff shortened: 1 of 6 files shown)


tests/test-bind-pointers.cpp
```diff
@@ -9,8 +9,7 @@ using namespace rhi::testing;
 GPU_TEST_CASE("bind-pointers-single-copy", Vulkan | CUDA)
 {
     ComPtr<IShaderProgram> shaderProgram;
-    slang::ProgramLayout* slangReflection = nullptr;
-    REQUIRE_CALL(loadComputeProgram(device, shaderProgram, "test-pointer-copy", "computeMain", slangReflection));
+    REQUIRE_CALL(loadProgram(device, "test-pointer-copy", {"computeMain"}, shaderProgram));
```
skallweitNV: We could introduce an overload to `loadProgram` that takes a single entry point name (and passes it on in a vector to build on the existing `loadProgram`). It's a very common use-case and would simplify all of these call sites. But having `{"computemain"}` is fine as well.

skallweitNV (approved):
LGTM. I like making the reflection output optional. The only minor suggestion would be adding overloads that take a single entry point name instead of a list. So feel free to merge as-is.

tests/test-bind-pointers.cpp
ksavoie-nv: That's a good idea. I added a couple of overloads for that the single entry point use case and updated the call sites.

skallweitNV (approved):
Thanks!

PR #574, opened 2025-11-12
CUDA 12.2 support

skallweitNV:
- Remove relying on `cuFuncGetParamInfo` which is not available before CUDA 12.4
- Compute parameter buffer size based on layout information (Slang workaround)


ccummingsNV (dismissed):
If it was me, I'd delete the commented out bits given they're gone forever. other than that looks good.

PR #575, opened 2025-11-12
Cluster Acceleration Structure builds in slang-rhi

dlacewellNV:
This PR plumbs `buildClusterAccelerationStructure()` and related functions and data types through slang-rhi to the OptiX backend, along with new tests and a new ABI for cluster build arguments passed in device memory. 
 
Clusters in a nutshell enable very fast acceleration structure builds by reusing local topology with dynamic vertex positions.  I'm working up to a Slangpy example, but this current PR is all about plumbing. 

Cluster builds are "indirect", i.e., most of the detailed, per-cluster data is passed in device memory and in practice is usually written by compute kernels.  Multiple cluster acceleration structures are built with a single call, and the output data is left in device memory specified by the user, with no automatic readback of the handles.  So these builds are quite different than the existing BLAS/TLAS builds.  

For more background on clusters, aka Megageometry, please refer to these blog posts:
https://developer.nvidia.com/blog/fast-ray-tracing-of-dynamic-scenes-using-nvidia-optix-9-and-nvidia-rtx-mega-geometry/
https://developer.nvidia.com/blog/nvidia-rtx-mega-geometry-now-available-with-new-vulkan-samples/

Details:
  * new entry points: `CommandEncoder::buildClusterAccelerationStructure()` and `Device::getClusterAccelerationStructureSizes()` that map 1-1 with corresponding OptiX functions in the backend  
  *  device ABI for the per-cluster arguments passed in device memory to the build.  The NVIDIA driver requires these arguments to match the same binary layout defined in the OptiX headers, Vulkan cluster extension, and NVAPI for DX.
  * feature capability check for OptiX 9+
  * debug layer with sanity checks on host arguments
  * tests in test-ray-tracing-clusters.cpp.  The main test builds two clusters, shoots rays to create an image buffer, and checks some pixels against expected values.  It also uses the GetClusterID() slang shader intrinsic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,10 +317,6 @@ if(EXISTS ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
     target_include_directories(slang-rhi PUBLIC ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
 endif()
 
-# Export the path to *.slangh files, so that consumers can build a runtime search path for Slang compilation
-file(TO_CMAKE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/include" SLANG_RHI_INCLUDE_DIR_PATH)
-target_compile_definitions(slang-rhi INTERFACE SLANG_RHI_INCLUDE_DIR="${SLANG_RHI_INCLUDE_DIR_PATH}")
-
 # Fetch & setup DirectXShaderCompiler
 if(SLANG_RHI_ENABLE_D3D12 AND (CMAKE_SYSTEM_NAME STREQUAL "Windows"))
     message(STATUS "Fetching DirectXShaderCompiler ...")
@@ -856,7 +852,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED)
         SLANG_RHI_DEBUG=$<BOOL:$<CONFIG:Debug>>
         SLANG_RHI_TESTS_DIR="${CMAKE_CURRENT_SOURCE_DIR}/tests"
         SLANG_RHI_NVAPI_INCLUDE_DIR="${SLANG_RHI_NVAPI_INCLUDE_DIR}"
-        SLANG_RHI_INCLUDE_DIR="${SLANG_RHI_INCLUDE_DIR_PATH}"
+        SLANG_RHI_INCLUDE_DIR="${CMAKE_CURRENT_SOURCE_DIR}/include"
         SLANG_RHI_OPTIX_DEVICE_HEADER_INCLUDE_DIR="${SLANG_RHI_OPTIX_DEVICE_HEADER_INCLUDE_DIR}"
         SLANG_RHI_BUILD_TESTS_WITH_GLFW=$<BOOL:${SLANG_RHI_BUILD_TESTS_WITH_GLFW}>
         $<$<PLATFORM_ID:Windows>:NOMINMAX> # do not define min/max macros
--- a/include/slang-rhi/cluster_accel_args_host.h
+++ b/include/slang-rhi/cluster_accel_args_host.h
@@ -1,9 +1,0 @@
-// Host shim for Cluster Acceleration args.
-
-#ifndef SLANG_RHI_CLUSTER_ACCEL_ARGS_HOST_H
-#define SLANG_RHI_CLUSTER_ACCEL_ARGS_HOST_H
-
-#include <cstdint>
-#include "cluster_accel_args.slangh"
-
-#endif // SLANG_RHI_CLUSTER_ACCEL_ARGS_HOST_H
--- a/tests/test-ray-tracing-clusters.cpp
+++ b/tests/test-ray-tracing-clusters.cpp
@@ -1,7 +1,7 @@
 #include "testing.h"
 #include "texture-utils.h"
 #include <slang-rhi/acceleration-structure-utils.h>
-#include <slang-rhi/cluster_accel_args_host.h>
+#include <slang-rhi/cluster_accel_args.h>
 
 using namespace rhi;
 using namespace rhi::testing;
--- a/include/slang-rhi/cluster_accel_args.h
+++ b/include/slang-rhi/cluster_accel_args.h
@@ -1,0 +1,6 @@
+// Host shim for Cluster Acceleration args.
+
+#pragma once
+
+#include <cstdint>
+#include "cluster_accel_args.slangh"
--- a/include/slang-rhi/cluster_accel_args.slangh
+++ b/include/slang-rhi/cluster_accel_args.slangh
@@ -2,8 +2,7 @@
 // Usable from both Slang and C++ (host).
 // C++ includers must include <cstdint> before including this file.
 
-#ifndef SLANG_RHI_CLUSTER_ACCEL_ARGS_SLANGH
-#define SLANG_RHI_CLUSTER_ACCEL_ARGS_SLANGH
+#pragma once
 
 namespace cluster_abi {
 
@@ -173,5 +172,3 @@ static inline ClustersArgs makeClustersArgs(uint32_t count, uint64_t handlesBuff
 }
 
 } // namespace cluster_abi
-
-#endif // SLANG_RHI_CLUSTER_ACCEL_ARGS_SLANGH


shaders/cluster_accel/cluster_accel_abi.common.inc
dlacewellNV: The *.inc pattern here is an attempt at safely sharing struct definitions between host and device.  U32/64 map to uint32_t and uint64_t in C++ but Slang seemed to prefer "uint" as its 32-bit type.

Also, should the slang version of the header go into the slang repo rather than slang-rhi?

include/slang-rhi/cluster_accel_abi_host.h
```diff
@@ -0,0 +1,17 @@
+// Host mirror for Cluster Acceleration args ABI.
+// Maps ABI_U32/ABI_U64 to C++ POD types and includes the shared common.
+
+#ifndef SLANG_RHI_CLUSTER_ACCEL_ABI_HOST_H
+#define SLANG_RHI_CLUSTER_ACCEL_ABI_HOST_H
+
+namespace cluster_abi {
+    #define ABI_U32 uint32_t
+    #define ABI_U64 uint64_t
+    #include "../../shaders/cluster_accel/cluster_accel_abi.common.inc"
```
skallweitNV: Can we move `cluster_accel_abi.common.inc` into `include/slang-rhi` so we distribute it along the other include files?

include/slang-rhi.h
```diff
@@ -122,6 +122,8 @@ enum class DeviceType
     x(RayQuery,                                 "ray-query"                                     ) \
     x(ShaderExecutionReordering,                "shader-execution-reordering"                   ) \
     x(RayTracingValidation,                     "ray-tracing-validation"                        ) \
+    /* Cluster acceleration structure */                                                          \
```
skallweitNV: Comment is not needed

shaders/cluster_accel/cluster_accel_abi.slang
```diff
@@ -0,0 +1,17 @@
+// Device-only ABI for Cluster Acceleration args.
```
skallweitNV: Maybe we should move this to `slang-rhi/include` as well and use `.slangh` extension. We have a soft-convention to use `.slangh` as the extension for slang files that are supposed to be `#include`ed instead of `#import`ed.

src/cpu/cpu-command.cpp
```diff
@@ -64,6 +64,7 @@ class CommandExecutor
     void cmdInsertDebugMarker(const commands::InsertDebugMarker& cmd);
     void cmdWriteTimestamp(const commands::WriteTimestamp& cmd);
     void cmdExecuteCallback(const commands::ExecuteCallback& cmd);
+    void cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd);
```
skallweitNV: Can we move this up to before `cmdConvertCooperativeVectorMatrix` so that all raytracing commands are in the same place?

src/cpu/cpu-command.cpp
```diff
@@ -287,6 +288,12 @@ void CommandExecutor::cmdDeserializeAccelerationStructure(const commands::Deseri
     NOT_SUPPORTED(S_CommandEncoder_deserializeAccelerationStructure);
 }
 
+void CommandExecutor::cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd)
```
skallweitNV, line 291: The implementation already is in the right spot :)

src/cuda/cuda-device.cpp
```diff
@@ -349,6 +349,8 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         {
             addFeature(Feature::AccelerationStructureSpheres);
             addFeature(Feature::AccelerationStructureLinearSweptSpheres);
+            // OptiX 9+ exposes the Cluster Acceleration Structure API; surface as a device feature.
+            addFeature(Feature::ClusterAccelerationStructure);
```
skallweitNV: Is this guaranteed to be available on older GPUs? Or do we need some HW support check as well?

src/cuda/optix-api-impl.cpp
```diff
@@ -16,9 +16,15 @@
 #include <optix.h>
 #include <optix_stubs.h>
 
+#if OPTIX_VERSION >= 90000
+#include <optix_types.h>
+#endif
+
+#ifdef EXPECTED_OPTIX_VERSION
```
skallweitNV: Can we remove this `#ifdef` again or what is it's purpose?

src/cuda/optix-api-impl.cpp
```diff
@@ -942,6 +1070,162 @@ class ContextImpl : public Context
...
+            buildInput.triangles.maxVertexCountPerArg = desc.limits.limitsTriangles.maxVertexCountPerArg;
+            buildInput.triangles.maxSbtIndexValue = desc.limits.limitsTriangles.maxUniqueSbtIndexCountPerArg ? (desc.limits.limitsTriangles.maxUniqueSbtIndexCountPerArg - 1) : 0;
+            break;
+        }
+        default:
+            return;
+        }
+
+        // Pass device-side args buffer directly to OptiX with stride 0 (tightly packed) and count 0 (use maxArgCount)
+        optixClusterAccelBuild(
```
skallweitNV: Add `SLANG_OPTIX_ASSERT_ON_FAIL` to catch errors in debug builds.

src/d3d12/d3d12-command.cpp
```diff
@@ -110,6 +110,7 @@ class CommandRecorder
     void cmdInsertDebugMarker(const commands::InsertDebugMarker& cmd);
     void cmdWriteTimestamp(const commands::WriteTimestamp& cmd);
     void cmdExecuteCallback(const commands::ExecuteCallback& cmd);
+    void cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd);
```
skallweitNV: move up

src/d3d12/d3d12-command.cpp
```diff
@@ -1328,6 +1329,12 @@ void CommandRecorder::cmdExecuteCallback(const commands::ExecuteCallback& cmd)
     cmd.callback(cmd.userData);
 }
 
+void CommandRecorder::cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd)
```
skallweitNV: move up

src/debug-layer/debug-command-encoder.cpp
```diff
@@ -753,6 +753,167 @@ void DebugCommandEncoder::deserializeAccelerationStructure(IAccelerationStructur
...
 
+void DebugCommandEncoder::buildClusterAccelerationStructure(
+    const ClusterAccelBuildDesc& desc
+)
+{
+    SLANG_RHI_API_FUNC;
+    requireOpen();
+    requireNoPass();
+    // Validate required fields based on build mode and operation.
+    auto fail = [&](const char* message) {
```
skallweitNV: Can you use `RHI_VALIDATION_ERROR` instead?

src/metal/metal-command.cpp
```diff
@@ -100,6 +100,7 @@ class CommandRecorder
     void cmdInsertDebugMarker(const commands::InsertDebugMarker& cmd);
     void cmdWriteTimestamp(const commands::WriteTimestamp& cmd);
     void cmdExecuteCallback(const commands::ExecuteCallback& cmd);
+    void cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd);
```
skallweitNV: move up

src/metal/metal-command.cpp
```diff
@@ -882,6 +883,12 @@ void CommandRecorder::cmdExecuteCallback(const commands::ExecuteCallback& cmd)
     cmd.callback(cmd.userData);
 }
 
+void CommandRecorder::cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd)
```
skallweitNV: move up

src/vulkan/vk-command.cpp
```diff
@@ -112,6 +112,7 @@ class CommandRecorder
     void cmdInsertDebugMarker(const commands::InsertDebugMarker& cmd);
     void cmdWriteTimestamp(const commands::WriteTimestamp& cmd);
     void cmdExecuteCallback(const commands::ExecuteCallback& cmd);
+    void cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd);
```
skallweitNV: move up

src/wgpu/wgpu-command.cpp
```diff
@@ -92,6 +92,7 @@ class CommandRecorder
     void cmdInsertDebugMarker(const commands::InsertDebugMarker& cmd);
     void cmdWriteTimestamp(const commands::WriteTimestamp& cmd);
     void cmdExecuteCallback(const commands::ExecuteCallback& cmd);
+    void cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd);
```
skallweitNV: move up

src/wgpu/wgpu-command.cpp
```diff
@@ -846,6 +847,12 @@ void CommandRecorder::cmdExecuteCallback(const commands::ExecuteCallback& cmd)
     cmd.callback(cmd.userData);
 }
 
+void CommandRecorder::cmdBuildClusterAccelerationStructure(const commands::BuildClusterAccelerationStructure& cmd)
```
skallweitNV: move up

CMakeLists.txt
```diff
@@ -317,6 +317,10 @@ if(EXISTS ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
     target_include_directories(slang-rhi PUBLIC ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
 endif()
 
+# Export shared shader directory to consumers (normalized to forward slashes)
+file(TO_CMAKE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/shaders" SLANG_RHI_SHADERS_DIR_PATH)
```
skallweitNV: We should remove this and just provide a define to slang-rhi-tests similar to `SLANG_RHI_NVAPI_INCLUDE_DIR`

include/slang-rhi.h
```diff
@@ -1438,6 +1440,116 @@ struct AccelerationStructureSizes
...
+    /// - TemplatesFromTriangles: use limitsTriangles
+    /// - CLASFromTemplates: use limitsTriangles (reuses triangle limits, matching all backends)
+    union
+    {
+        ClusterAccelLimitsTriangles limitsTriangles;
+        ClusterAccelLimitsClusters limitsClusters;
+    } limits = {};
+
+    // Build mode and associated parameters. Defaults to Implicit when unspecified.
+    enum class BuildMode : uint32_t { Implicit = 0, Explicit = 1, GetSizes = 2 };
```
ksavoie-nv: Is there a reason why we explicitly specify this enum as a `uint32_t` and assign it values?

src/command-list.cpp
```diff
@@ -299,6 +299,12 @@ void CommandList::write(commands::DeserializeAccelerationStructure&& cmd)
     writeCommand(std::move(cmd));
 }
 
+void CommandList::write(commands::BuildClusterAccelerationStructure&& cmd)
+{
+    // Copy desc by value is fine; buffers are passed via device addresses in desc.
```
ksavoie-nv: It looks like this method is avoiding copying, so I assume the copy happens in `writeCommand`? It might be worth updating the comment to mention if that's the case.

src/cuda/optix-api-impl.cpp
```diff
@@ -16,9 +16,15 @@
 #include <optix.h>
 #include <optix_stubs.h>
 
+#if OPTIX_VERSION >= 90000
+#include <optix_types.h>
```
ksavoie-nv: `optix_types.h` should come in with `optix.h`.

src/cuda/optix-api-impl.cpp
```diff
@@ -824,6 +835,123 @@ class ContextImpl : public Context
...
+        buildMode.getSize.outputSizesStrideInBytes = 0;
+
+        switch (desc.op)
+        {
+        case ClusterAccelBuildOp::CLASFromTriangles:
+        {
+            buildInput.type = OPTIX_CLUSTER_ACCEL_BUILD_TYPE_CLUSTERS_FROM_TRIANGLES;
+            // Sizing query uses IMPLICIT mode to get total required buffer sizes
+            buildMode.mode = OPTIX_CLUSTER_ACCEL_BUILD_MODE_IMPLICIT_DESTINATIONS;
+            if (desc.limits.limitsTriangles.maxArgCount == 0 ||
```
ksavoie-nv: Is this check necessary? It looks like we check it in the debug layer, and I imagine OptiX's validation mode should check it as well.

tests/test-ray-tracing-clusters.cpp
```diff
@@ -0,0 +1,727 @@
...
+    enc->buildClusterAccelerationStructure(blasDesc);
+    queue->submit(enc->finish());
+    queue->waitOnHost();
+
+    REQUIRE_CALL(device->readBuffer(result.outputBuffer, 0, sizeof(uint64_t), &result.blasHandle));
+
+    return result;
+}
+
+GPU_TEST_CASE("cluster-accel-sizes-optix", CUDA)
```
ksavoie-nv: I would delete the `-optix` suffix here. The test suite will automatically add a `.cuda`, `.vulkan`, etc. suffix depending on which API the test uses.

tests/test-ray-tracing-clusters.cpp
```diff
@@ -0,0 +1,727 @@
...
+#include <slang-rhi/cluster_accel_abi_host.h>
+
+using namespace rhi;
+using namespace rhi::testing;
+
+struct Float3 { float x, y, z; };
+
+static void requireClusterAccelOrSkip(IDevice* device)
+{
+    if (device->getDeviceType() != DeviceType::CUDA)
```
ksavoie-nv: I don't think you need to check that you're using a CUDA and OptiX 9.0 device here. Device instantiation should check that and set the `ClusterAccelerationFeature` accordingly.

And if you only check for the feature, whoever adds vulkan support won't have to track this check down and update it.

tests/test-ray-tracing-clusters.cpp
```diff
@@ -0,0 +1,727 @@
...
+        
+        // PrimitiveID resets per cluster. For cell c=1 and tri0, expected primitive index is 2*c = 2
+        CHECK_EQ(id0.y, 2u);
+        CHECK_EQ(id1.y, 2u);
+    }
+
+    // Read back and validate two horizontal hit strips (rows with any non-zero pixel)
+    ComPtr<ISlangBlob> image;
+    SubresourceLayout layout;
+    REQUIRE_CALL(device->readTexture(resultTexture, 0, 0, image.writeRef(), &layout));
```
ksavoie-nv, line 728: You could add this snippet here:
```
#if 0 // for debugging only
        writeImage("test.hdr", resultBlob, width, height, layout.rowPitch, layout.colPitch);
#endif
```
Other ray tracing tests have that snippet to make it convenient to grab the result image.

shaders/cluster_accel/cluster_accel_abi.slang
dlacewellNV: Yes, I like this idea.  I think we can do a single header with the .slangh extension and probably find a way to get rid of the *.inc file. 

For one of your general questions: yes the cluster argument types passed in device memory are the same for all backends -- they go through directly to the driver.  We should do a VK/DX backend in the near future using the same plumbing.

I'll go through your other comments over the next few days and fix the Windows builds.  Appreciate the detailed comments.

src/command-list.cpp
skallweitNV: I don't think the comment is really necessary in the first place.

src/cuda/optix-api-impl.cpp
skallweitNV: Good point, I was wondering too if this include is necessary.

tests/test-ray-tracing-clusters.cpp
skallweitNV: Agreed, tests should check features and they will automatically be run on devices that have support.

shaders/cluster_accel/cluster_accel_abi.slang
dlacewellNV: I consolidated headers into one main file (cluster_accel_args.slangh) intended to be included into .slang files, and another small shim file to make it includable into host code as well for debugging.  (Sometimes it's useful to fill a simple arg buffer from the host for simple tests, or read back to the host and examine the args there).

src/cuda/cuda-device.cpp
dlacewellNV: excellent point.  I updated this and made use of the device caps query for cluster accels, exposed in OptiX.  Should have done that to begin with.  It's available on Ada+, but the query covers it.

CMakeLists.txt
dlacewellNV: I renamed this path variable but left the definition here.  I think it's needed, but correct me if I'm wrong - I want the new slangh header (cluster_accel_args.slangh) to be usable in sample code in slangpy.  For example:

```
    const std::filesystem::path RHI_INCLUDE_DIR(SLANG_RHI_INCLUDE_DIR);
    ref<Device> device = Device::create({
        .type = DeviceType::cuda,
        .enable_debug_layers = true,
        .compiler_options = {.include_paths = {EXAMPLE_DIR, RHI_INCLUDE_DIR}},
    });
```

include/slang-rhi.h
dlacewellNV: the reason is GPT-5.  fixed.

tests/test-ray-tracing-clusters.cpp
dlacewellNV: this worked great.  image attached in case you're curious what the test looks like; it's just two small clusters, each consisting of a small strip of triangles.

image

src/cuda/optix-api-impl.cpp
dlacewellNV: removed these extra checks

dlacewellNV (commented):
Updated files and addressed most comments.  There is still an open discussion about whether we should export `SLANG_RHI_INCLUDE_DIR` from CMakeLists.txt, to make it easier for parent projects to find `cluster_accel_args.slangh`.

CMakeLists.txt
dlacewellNV: fwiw, a non-CMake approach might be to define a utility function in slang-rhi for the include directory:

```
#include <filesystem>

namespace rhi {
    inline std::filesystem::path getIncludeDirectory() {
        // Returns the directory containing this header file
        return std::filesystem::path(__FILE__).parent_path();
    }
}
```

Or, third option, we could let a slangpy sample go find the .slangh file on its own by looking at the slang-rhi location and building a path.  Not my favorite.

include/slang-rhi/cluster_accel_args_host.h
```diff
@@ -0,0 +1,9 @@
+// Host shim for Cluster Acceleration args.
```
skallweitNV: Minor:

This could just be `cluster_accel_args.h` instead.
We typically use `#pragma once` instead of the include guards.

CMakeLists.txt
```diff
@@ -317,6 +317,10 @@ if(EXISTS ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
     target_include_directories(slang-rhi PUBLIC ${SLANG_RHI_SLANG_INCLUDE_DIR}/../prelude)
 endif()
 
+# Export the path to *.slangh files, so that consumers can build a runtime search path for Slang compilation
+file(TO_CMAKE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/include" SLANG_RHI_INCLUDE_DIR_PATH)
+target_compile_definitions(slang-rhi INTERFACE SLANG_RHI_INCLUDE_DIR="${SLANG_RHI_INCLUDE_DIR_PATH}")
```
skallweitNV: `SLANG_RHI_INCLUDE_DIR` is not gonna be massively useful for clients as this is an absolute path and will not necessarily be available at runtime, when the library was copied somewhere else. So I would remove this for now and just provide the a `SLANG_RHI_INCLUDE_DIR` define to `slang-rhi-tests`.

CMakeLists.txt
```diff
@@ -851,6 +856,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED)
         SLANG_RHI_DEBUG=$<BOOL:$<CONFIG:Debug>>
         SLANG_RHI_TESTS_DIR="${CMAKE_CURRENT_SOURCE_DIR}/tests"
         SLANG_RHI_NVAPI_INCLUDE_DIR="${SLANG_RHI_NVAPI_INCLUDE_DIR}"
+        SLANG_RHI_INCLUDE_DIR="${SLANG_RHI_INCLUDE_DIR_PATH}"
```
skallweitNV:
```suggestion
        SLANG_RHI_INCLUDE_DIR="${CMAKE_CURRENT_SOURCE_DIR}/include"
```

skallweitNV (approved):

PR #576, opened 2025-11-13
Initial cooperative vector support for OptiX

skallweitNV:
Refactor cooperative vectors API to allow supporting OptiX backend.

- Add `CooperativeVectorMatrixDesc` for describing single matrix for conversion
- Add `ICommandEncoder::convertCooperativeVectorMatrix()` overload taking new descriptors
- Add `IDevice::convertCooperativeVectorMatrix()` overload taking new descriptors
- Add `IDevice::computeCooperativeVectorMatrixSize()` for computing size of cooperative vectors matrix
  - Automatically pad computed size to 64 bytes
  - Automatically assume tight row/col packing if `rowColumnStride` is set to zero
- Add tests for new APIs (including conversion of multiple matrices)
- Mark old APIs as deprecated (can be removed later when slangpy doesn't depend on them anymore)

### DIFF
--- a/tests/test-cooperative-vector.cpp
+++ b/tests/test-cooperative-vector.cpp
@@ -225,7 +225,7 @@ public:
     {
     }
 
-    static constexpr size_t size_bytes() { return Rows * Cols * sizeof(T); }
+    static constexpr size_t sizeBytes() { return Rows * Cols * sizeof(T); }
 
     T& operator()(size_t r, size_t c) { return m_data[getIndex(r, c)]; }
     const T& operator()(size_t r, size_t c) const { return m_data[getIndex(r, c)]; }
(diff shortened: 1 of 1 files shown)


include/slang-rhi.h
```diff
@@ -2161,6 +2161,7 @@ enum class CooperativeVectorMatrixLayout
     TrainingOptimal = 3,
 };
 
+// deprecated
 struct ConvertCooperativeVectorMatrixDesc
```
ccummingsNV, line 2165: Any reason we're keeping this old API around? Seems like a painful path to start down

skallweitNV: we keep it until slangpy doesn't depend on it anymore

PR #577, opened 2025-11-15
Fix MSVC warning C4709: comma operator within a subscript expression

corporateshark:
Fixed an MSVC warning `C4709`: comma operator within a subscript expression.

### DIFF
--- a/src/core/reverse-map.h
+++ b/src/core/reverse-map.h
@@ -15,7 +15,9 @@ auto reverseMap(Func func, From min, From max, From defaultValue = From(0))
         std::unordered_map<To, From> map;
         for (int i = int(min); i <= int(max); i++)
         {
-            map[func(From(i))] = From(i);
+            const From fromI = From(i);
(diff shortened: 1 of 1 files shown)


src/core/reverse-map.h
```diff
@@ -15,7 +15,8 @@ auto reverseMap(Func func, From min, From max, From defaultValue = From(0))
         std::unordered_map<To, From> map;
         for (int i = int(min); i <= int(max); i++)
         {
-            map[func(From(i))] = From(i);
+            const To key = func(From(i)); // Fixed MSVC warning C4709: comma operator within a subscript expression
+            map[key] = From(i);
```
jkwak-work: The warning makes no sense to me too but this fix works.
I guess it is a bug in vs2026.

Although I think this is fine as it is, I prefer to avoid calling the function twice,
```
-            map[func(From(i))] = From(i);
+            auto fromI = From(i);
+            auto funcFromI = func(fromI);
+            map[funcFromI] = fromI;
```

jkwak-work (approved):
Looks good to me.

PR #584, opened 2025-11-20
Add tests for motion blur

ksavoie-nv:
Expose motion blur in the VK backend and add tests for `TraceMotionRay` and `HitObject.TraceMotionRay`. 

Changes:
- Add `Feature::RayTracingMotionBlur`. During VK device creation, query for motion blur extension and enable feature if it's present.
- Add flags for enabling motion blur during accel builds and pipeline creation
- Update VK accel structure conversion to handle motion blur
- Update VK accel structure creation to chain motion info when it is supplied
- Update non-VK backends to reject motion blur build inputs
- Add tests for motion blur intersection (static, matrix, and SRT) and hit object creation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,7 +138,7 @@ endif()
 
 # NVAPI
 if(NOT DEFINED SLANG_RHI_NVAPI_URL)
-    set(SLANG_RHI_NVAPI_URL "https://github.com/NVIDIA/nvapi/archive/7cb76fce2f52de818b3da497af646af1ec16ce27.zip") # R575
+    set(SLANG_RHI_NVAPI_URL "https://github.com/NVIDIA/nvapi/archive/3d34a4faf095996663321646ebe003539a908f89.zip") # R580
 endif()
 
 # OptiX headers
--- a/include/slang-rhi/capabilities.h
+++ b/include/slang-rhi/capabilities.h
@@ -70,6 +70,7 @@
     x(_cuda_sm_7_0) \
     x(_cuda_sm_8_0) \
     x(_cuda_sm_9_0) \
+    x(optix_coopvec) \
     x(vertex) \
     x(fragment) \
     x(compute) \
@@ -98,6 +99,7 @@
     x(SPV_KHR_quad_control) \
     x(SPV_KHR_fragment_shader_barycentric) \
     x(SPV_KHR_non_semantic_info) \
+    x(SPV_KHR_device_group) \
     x(SPV_KHR_ray_tracing) \
     x(SPV_KHR_ray_query) \
     x(SPV_KHR_ray_tracing_position_fetch) \
@@ -106,12 +108,18 @@
     x(SPV_KHR_subgroup_rotate) \
     x(SPV_NV_ray_tracing_motion_blur) \
     x(SPV_NV_shader_invocation_reorder) \
+    x(SPV_NV_cluster_acceleration_structure) \
+    x(SPV_NV_linear_swept_spheres) \
     x(SPV_NV_shader_image_footprint) \
     x(SPV_KHR_compute_shader_derivatives) \
     x(SPV_GOOGLE_user_type) \
     x(SPV_EXT_replicated_composites) \
+    x(SPV_KHR_vulkan_memory_model) \
     x(SPV_NV_cooperative_vector) \
     x(SPV_KHR_cooperative_matrix) \
+    x(SPV_NV_tensor_addressing) \
+    x(SPV_NV_cooperative_matrix2) \
+    x(spvDeviceGroup) \
     x(spvAtomicFloat32AddEXT) \
     x(spvAtomicFloat16AddEXT) \
     x(spvAtomicFloat64AddEXT) \
@@ -142,6 +150,8 @@
     x(spvRayQueryKHR) \
     x(spvRayQueryPositionFetchKHR) \
     x(spvShaderInvocationReorderNV) \
+    x(spvRayTracingClusterAccelerationStructureNV) \
+    x(spvRayTracingLinearSweptSpheresGeometryNV) \
     x(spvShaderClockKHR) \
     x(spvShaderNonUniformEXT) \
     x(spvShaderNonUniform) \
@@ -151,8 +161,16 @@
     x(spvCooperativeVectorNV) \
     x(spvCooperativeVectorTrainingNV) \
     x(spvCooperativeMatrixKHR) \
+    x(spvCooperativeMatrixReductionsNV) \
+    x(spvCooperativeMatrixConversionsNV) \
+    x(spvCooperativeMatrixPerElementOperationsNV) \
+    x(spvCooperativeMatrixTensorAddressingNV) \
+    x(spvCooperativeMatrixBlockLoadsNV) \
+    x(spvTensorAddressingNV) \
     x(spvMaximalReconvergenceKHR) \
     x(spvQuadControlKHR) \
+    x(spvVulkanMemoryModelKHR) \
+    x(spvVulkanMemoryModelDeviceScopeKHR) \
     x(_GL_EXT_buffer_reference) \
     x(_GL_EXT_buffer_reference_uvec2) \
     x(_GL_EXT_debug_printf) \
@@ -168,6 +186,7 @@
     x(_GL_EXT_shader_atomic_float_min_max) \
     x(_GL_EXT_shader_atomic_float2) \
     x(_GL_EXT_shader_atomic_int64) \
+    x(_GL_EXT_shader_explicit_arithmetic_types) \
     x(_GL_EXT_shader_explicit_arithmetic_types_int64) \
     x(_GL_EXT_shader_image_load_store) \
     x(_GL_EXT_shader_realtime_clock) \
@@ -175,6 +194,7 @@
     x(_GL_EXT_texture_shadow_lod) \
     x(_GL_EXT_maximal_reconvergence) \
     x(_GL_EXT_shader_quad_control) \
+    x(_GL_EXT_device_group) \
     x(_GL_ARB_derivative_control) \
     x(_GL_ARB_fragment_shader_interlock) \
     x(_GL_ARB_gpu_shader5) \
@@ -208,7 +228,9 @@
     x(_GL_NV_shader_atomic_fp16_vector) \
     x(_GL_NV_shader_invocation_reorder) \
     x(_GL_NV_shader_subgroup_partitioned) \
-    x(_GL_NV_shader_texture_footprint)
+    x(_GL_NV_shader_texture_footprint) \
+    x(_GL_NV_cluster_acceleration_structure) \
+    x(_GL_NV_cooperative_vector)
 // clang-format on
 
 namespace rhi {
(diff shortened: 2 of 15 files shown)


include/slang-rhi.h
```diff
@@ -1295,7 +1296,7 @@ struct AccelerationStructureBuildInputInstances
     uint32_t instanceCount;
 };
 
-static const uint32_t kMaxAccelerationStructureMotionKeyCount = 2;
+inline constexpr uint32_t kMaxAccelerationStructureMotionKeyCount = 2;
```
skallweitNV, line 1255: We should probably change over the other `static const` constants as well, but this can be a separate PR.

src/vulkan/vk-acceleration-structure.cpp
```diff
@@ -148,6 +159,19 @@ Result AccelerationStructureBuildDescConverter::convert(
...
                 triangles.preTransformBuffer ? triangles.preTransformBuffer.getDeviceAddress() : 0;
 
+            if (useMotion)
+            {
+                VkAccelerationStructureGeometryMotionTrianglesDataNV& motionData = motionTrianglesDatas[i];
+                motionData.sType = VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_GEOMETRY_MOTION_TRIANGLES_DATA_NV;
+                motionData.pNext = nullptr;
+                // The vertexData points to a buffer containing vertex data for all motion keys
+                // laid out contiguously. We use the first vertex buffer address.
+                motionData.vertexData.deviceAddress = triangles.vertexBuffers[0].getDeviceAddress();
```
skallweitNV: Shouldn't this set the address of `vertexBuffers[1]`?
According to Vulkan:
> If VkAccelerationStructureGeometryMotionTrianglesDataNV is included in the pNext chain of a [VkAccelerationStructureGeometryTrianglesDataKHR](https://docs.vulkan.org/refpages/latest/refpages/source/VkAccelerationStructureGeometryTrianglesDataKHR.html) structure, the basic vertex positions are used for the position of the triangles in the geometry at time 0.0 and the vertexData in VkAccelerationStructureGeometryMotionTrianglesDataNV is used for the vertex positions at time 1.0, with positions linearly interpolated at intermediate times.

src/vulkan/vk-acceleration-structure.cpp
```diff
@@ -156,11 +180,17 @@ Result AccelerationStructureBuildDescConverter::convert(
...
+        {
+            return SLANG_E_NOT_AVAILABLE;
+        }
+
         for (uint32_t i = 0; i < buildDesc.inputCount; ++i)
         {
             const AccelerationStructureBuildInputProceduralPrimitives& proceduralPrimitives =
                 buildDesc.inputs[i].proceduralPrimitives;
-            if (proceduralPrimitives.aabbBufferCount != 1)
+            if (proceduralPrimitives.aabbBufferCount > kMaxAccelerationStructureMotionKeyCount)
```
skallweitNV: The previous check was probably better because we don't support motion blur.

src/vulkan/vk-acceleration-structure.cpp
```diff
@@ -181,11 +211,17 @@ Result AccelerationStructureBuildDescConverter::convert(
...
+        {
+            return SLANG_E_NOT_AVAILABLE;
+        }
+
         spheresDatas.resize(buildDesc.inputCount);
         for (uint32_t i = 0; i < buildDesc.inputCount; ++i)
         {
             const AccelerationStructureBuildInputSpheres& spheres = buildDesc.inputs[i].spheres;
-            if (spheres.vertexBufferCount != 1)
+            if (spheres.vertexBufferCount > kMaxAccelerationStructureMotionKeyCount)
```
skallweitNV: Same here, previous check was better IMO.

src/vulkan/vk-acceleration-structure.cpp
```diff
@@ -227,11 +263,17 @@ Result AccelerationStructureBuildDescConverter::convert(
...
+        {
+            return SLANG_E_NOT_AVAILABLE;
+        }
+
         linearSweptSpheresDatas.resize(buildDesc.inputCount);
         for (uint32_t i = 0; i < buildDesc.inputCount; ++i)
         {
             const AccelerationStructureBuildInputLinearSweptSpheres& lss = buildDesc.inputs[i].linearSweptSpheres;
-            if (lss.vertexBufferCount != 1)
+            if (lss.vertexBufferCount > kMaxAccelerationStructureMotionKeyCount)
```
skallweitNV: Same

tests/test-ray-tracing-hitobject-intrinsics.cpp
```diff
@@ -433,3 +472,106 @@ GPU_TEST_CASE("ray-tracing-hitobject-make-hit", ALL | DontCreateDevice)
...
+
+    if (!device->hasFeature(Feature::RayTracing))
+        SKIP("ray tracing not supported");
+    if (!device->hasFeature(Feature::ShaderExecutionReordering))
+        SKIP("shader execution reordering not supported");
+    if (!device->hasFeature(Feature::RayTracingMotionBlur))
+        SKIP("ray tracing motion blur not supported");
+
+    // Disabled under pre OptiX 9.0 due to https://github.com/shader-slang/slang/issues/8723
+    if (device->getDeviceType() == DeviceType::CUDA && device->getInfo().optixVersion < 90000)
```
skallweitNV, line 533: Is https://github.com/shader-slang/slang/issues/8723 fixable? I.e. can we support SER with OptiX 8.1 at all? If not, we should just require OptiX 9.0 for SER and remove this check.

skallweitNV (commented):
Looks good overall.

I'm a bit surprised this works for Vulkan given that we provide the vertex data for motion key 0 at both ends.

Other than that, the main broader point this PR brings up in my mind is how we deal with incorrect data passed to the slang-rhi API. What should be checked in the validation layer vs. what should be checked in the backends. We need to establish better guidelines.

tests/test-ray-tracing-hitobject-intrinsics.cpp
ksavoie-nv: I'm not sure, but I can definitely take a look.

src/vulkan/vk-acceleration-structure.cpp
ksavoie-nv: Good catch, thanks. I agree; I accidentally used the wrong index and didn't realize I was getting the wrong triangle in the output image.

include/slang-rhi.h
```diff
@@ -1401,6 +1409,8 @@ struct AccelerationStructureDesc
     const void* next = nullptr;
 
     uint64_t size;
+    AccelerationStructureBuildFlags flags = AccelerationStructureBuildFlags::None;
+    AccelerationStructureMotionInfo* motionInfo = nullptr;
```
skallweitNV: These extra options feel wrong to me.
An acceleration structure should just be a chunk of memory. I haven't looked at why this is needed specifically, but we should try to get rid of this.

include/slang-rhi.h
```diff
@@ -1372,10 +1373,17 @@ enum class AccelerationStructureBuildFlags
...
-    MinimizeMemory = (1 << 4)
+    MinimizeMemory = (1 << 4),
+    CreateMotion = (1 << 5)
 };
 SLANG_RHI_ENUM_CLASS_OPERATORS(AccelerationStructureBuildFlags);
 
+struct AccelerationStructureMotionInfo
+{
+    uint32_t maxInstances = 0;
+    uint32_t flags = 0;
```
skallweitNV: I'd remove this for now, as Vulkan has these flags `0` and reserved for future use.

include/slang-rhi.h
skallweitNV: Looks like the `maxInstances` is needed. But, I don't think `flags` is needed, so let's remove that.

skallweitNV: To keep things more consistent with the build desc, I'd be in favor of having something this

```c++
struct AccelerationStructureMotionOptions
{
    bool enabled = false;
    uint32_t maxInstances = 0;
}

struct AccelerationStructureDesc
{
    StructType structType = StructType::AccelerationStructureDesc;
    const void* next = nullptr;

    uint64_t size;

    AccelerationStructureMotionOptions motionOptions;

    const char* label = nullptr;
};
```

src/vulkan/vk-acceleration-structure.cpp
```diff
@@ -117,10 +122,20 @@ Result AccelerationStructureBuildDescConverter::convert(
...
+        if (useMotion)
+        {
+            motionTrianglesDatas.resize(buildDesc.inputCount);
+        }
+
         for (uint32_t i = 0; i < buildDesc.inputCount; ++i)
         {
             const AccelerationStructureBuildInputTriangles& triangles = buildDesc.inputs[i].triangles;
-            if (triangles.vertexBufferCount != 1)
+            if (triangles.vertexBufferCount > kMaxAccelerationStructureMotionKeyCount)
```
skallweitNV, line 134: For Vulkan I think the only valid case when `useMotion == true` is `vertexBufferCount == 2`. This sort of happens now because `kMaxAccelerationStructureMotionKeyCount == 2` but that could be lifted for OptiX if we really wanted to.

tests/test-ray-tracing-common.h
```diff
@@ -363,19 +363,134 @@ struct SingleCustomGeometryBLAS
...
+        ComPtr<IQueryPool> compactedSizeQuery;
+        QueryPoolDesc queryPoolDesc;
+        queryPoolDesc.count = 1;
+        queryPoolDesc.type = QueryType::AccelerationStructureCompactedSize;
+        REQUIRE_CALL(device->createQueryPool(queryPoolDesc, compactedSizeQuery.writeRef()));
+
+        ComPtr<IAccelerationStructure> draftAS;
+        AccelerationStructureDesc draftCreateDesc;
+        draftCreateDesc.size = sizes.accelerationStructureSize;
+        REQUIRE_CALL(device->createAccelerationStructure(draftCreateDesc, draftAS.writeRef()));
```
skallweitNV, line 471: For both of these calls to `createAccelerationStructure` you don't set `maxInstances`?! Wouldn't that be required to get a proper BLAS?

skallweitNV: Ah I guess the `maxInstances` applies to the TLAS then? Maybe that should be clarified in the slang-rhi.h header then.

include/slang-rhi.h
ksavoie-nv: > These extra options feel wrong to me.
An acceleration structure should just be a chunk of memory. I haven't looked at why this is needed specifically, but we should try to get rid of this.

Heh, agreed. I spent some time scratching my head because I assumed the options should come in during the build, but it turns out they're required during creation.

I'll update the options to follow your suggestion.

tests/test-ray-tracing-common.h
ksavoie-nv: Yes, I believe `maxInstances` is only required for the TLAS.

ksavoie-nv:
@skallweitNV I updated the new motion options struct to follow your suggestion (pulling it directly into `AccelerationStructureDesc` and noting that it's only required for TLAS builds).

I also made a few other changes to make its usage and the header a bit clearer.

skallweitNV (approved):

PR #585, opened 2025-11-20
Introduce slang-rhi-device.h

skallweitNV:
This PR introduces a new `slang-rhi-device.h` header file that contains definitions that can be used from both the host and the deivce/slang side. This new header is automatically included in `slang-rhi.h`. It needs to be manually included in slang shaders.

- Introduce `slang-rhi-device.h` shared header
- Move `DeviceAddress`, indirect arg structs and acceleration structure instance descs to shared header
- Merge `cluster_accel_args.h/slangh` into shared header
- Cleanup cluster structs and document how they map to OptiX, D3D12 and Vulkan
- Introduce constants for enums used in cluster structs
- Remove packing functions and switch to `uint16_t` and bit-fields in the cluster structs
- Adjust the cluster test
- Remove `SLANG_RHI_INCLUDE_DIR` compile definition in `slang-rhi-tests` and use relative include of `slang-rhi-device.h` instead to enable Slang language server support.

### DIFF
--- a/include/slang-rhi-device.h
+++ b/include/slang-rhi-device.h
@@ -164,6 +164,8 @@ struct TriangleClusterArgs
     /// (optional) Address of an array of indices into the OMM array. Note that an additional OMM special index is
     /// reserved and can be used to turn off OMM for specific triangles.
     DeviceAddress opacityMicromapIndexBuffer;
+    /// TODO: Needed for OptiX. D3D12/Vulkan have a different struct adding this field but sharing all other fields.
+    DeviceAddress instantiationBoundingBoxLimit;
 };
 
 /// Arguments for instantiating cluster from template.


tests/test-ray-tracing-clusters.slang
```diff
@@ -1,5 +1,6 @@
 // Test shaders for clusters build-and-shoot
-#include <slang-rhi/cluster_accel_args.slangh>
+// Using relative path to allow Slang language server to find the file.
+#include "../include/slang-rhi-device.h"
```
tdavidovicNV, line 3: Can't we instead add the include path to the slang language server settings instead?

skallweitNV: I tried and for some reason it didn't work. I found this to be an OK alternative, zero-setup and should "just work".

PR #593, opened 2025-12-04
Add support for SLANG_RHI_TESTS_DIR env var

### LINKED ISSUE #592: `slang-rhi-tests` fails when run on different machine than build due to hardcoded `SLANG_RHI_TESTS_DIR`
### Problem Description
The `slang-rhi-tests` binary cannot find test `.slang` files when executed on a different machine (or different path) than where it was built. This is because `SLANG_RHI_TESTS_DIR` is set at compile-time to an absolute path.

### Root Cause
In `CMakeLists.txt` line 852:
```cmake
SLANG_RHI_TESTS_DIR="${CMAKE_CURRENT_SOURCE_DIR}/tests"
```

This hardcodes the **build machine's absolute path** into the binary. When the binary is run on a different machine with a different checkout path, it cannot find the test shaders.

### Reproduction
1. Build `slang-rhi-tests` on machine A at path `/path/A/slang-rhi/`
2. Copy the binary to machine B where the repo is at `/path/B/slang-rhi/`
3. Run `slang-rhi-tests` on machine B
4. All tests fail with: `error 1: cannot open file 'test-benchmark-command.slang'`

### Real-World Impact
This breaks CI workflows where:
- Build happens on one machine/container (e.g., ubuntu-22.04 GitHub runner with CUDA container)
- Tests run on a different machine (e.g., self-hosted GPU runner)

Specifically affects the Slang project's Linux GPU CI where CUDA builds happen in containers but GPU tests run on self-hosted runners with different paths.

### Proposed Solutions

**Option 1: Runtime environment variable (recommended)**
Check for `SLANG_RHI_TESTS_DIR` environment variable at runtime, falling back to compile-time default:
```cpp
const char* getTestsDir()
{
    const char* envDir = std::getenv("SLANG_RHI_TESTS_DIR");
    if (envDir && envDir[0] != '\0')
        return envDir;
    return SLANG_RHI_TESTS_DIR; // compile-time fallback
}
```

**Option 2: Relative path from executable**
Detect the executable's location at runtime and compute relative path to tests directory.

**Option 3: Command-line argument**
Add a `-tests-dir <path>` argument to `slang-rhi-tests`.

### Environment
- slang-rhi version: current main
- Discovered while implementing Linux GPU CI for shader-slang/slang

skallweitNV:
Fixes #592

### DIFF
--- a/tests/testing.cpp
+++ b/tests/testing.cpp
@@ -73,6 +73,21 @@ void cleanupTestTempDirectories()
     remove_all(gTestTempDirectory);
 }
 
+const char* getEnvVariable(const char* name, const char* defaultValue = nullptr)
+{
+#if SLANG_WINDOWS_FAMILY
+    static char value[4096];
+    size_t len = 0;
+    if (::getenv_s(&len, value, sizeof(value), "SLANG_RHI_TESTS_DIR") == 0 && len > 0)
+        return static_cast<const char*>(value);
+    else
+        return defaultValue;
+#else
+    const char* value = ::getenv(name);
+    return value ? value : defaultValue;
+#endif
+}
+
 std::string readFile(std::string_view path)
 {
     std::ifstream file(std::string(path).c_str());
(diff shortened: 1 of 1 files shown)


tests/testing.cpp
```diff
@@ -733,17 +733,8 @@ void releaseCachedDevices()
...
-        static char value[4096];
-        size_t len = 0;
-        if (::getenv_s(&len, value, sizeof(value), "SLANG_RHI_TESTS_DIR") == 0 && len > 0 && value[0] != '\0')
-        {
-            return static_cast<const char*>(value);
-        }
-        return SLANG_RHI_TESTS_DIR;
-    }();
-    return testsDir;
+    const char* value = ::getenv("SLANG_RHI_TESTS_DIR");
```
jkiviluoto-nv: Other option is to use `_dupenv_s()` -based solution on Windows, which might be more targeted and leave warnings enabled.

skallweitNV: Using `getenv_s` on windows now.

PR #596, opened 2025-12-04
Initialize slang capabilities

skallweitNV:


include/slang-rhi/capabilities.h
```diff
@@ -60,6 +60,7 @@
     x(_sm_6_9) \
     x(hlsl_nvapi) \
     x(hlsl_2018) \
+    x(hlsl_coopvec_poc) \
```
jkwak-work: I am currently trying to remove this capability:
- https://github.com/shader-slang/slang/pull/9213

skallweitNV:
> I didn't know that slang-rhi has a list of capability. Does it mean we should modify slang-rhi side `include/slang-rhi/capabilities.h` whenever we changes the capability on Slang repo?

I opened https://github.com/shader-slang/slang/issues/7005 for that. It would be neat if slang shipped with a header file defining all available capabilities. Currently I have to update that list manually ..

include/slang-rhi/capabilities.h
skallweitNV: Ok, I can remove that one.

src/cuda/cuda-device.cpp
```diff
@@ -356,21 +400,19 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
             if (m_ctx.optixContext->getCooperativeVectorSupport())
             {
                 addFeature(Feature::CooperativeVector);
+                // addCapability(Capability::optix_coopvec);
```
ccummingsNV, line 403: Not clear to me why this is here but commented out - maybe delete or say why?

skallweitNV: Good point. This is commented out because when setting that capability in a session that is used to compile a compute kernel it fails in slang. Guess the reason is that this capability is invalid in a compute/CUDA-only context, but why slang doesn't just ignore it is to be answered. Because there are other capabilities that have no meaning in certain entry point types for example, which all seem to work fine.

PR #601, opened 2025-12-11
Add Aftermath support

skallweitNV:
This adds support for NVIDIA Aftermath:
- `SLANG_RHI_ENABLE_AFTERMATH` cmake option enables/disables fetching and building with Aftermath SDK (enabled by default)
- Aftermath is supported on D3D11, D3D12 and Vulkan backends
- Aftermath is enabled using the `DeviceDesc::enableAftermath` flag.
- Aftermath can be further configured using the `DeviceDesc::aftermathFlags` flags.
- Full support for dumping shaders (use shader debug info to get source level info in NSight)
- Full support for debug group tracking on D3D12 and Vulkan
- Crash dumps are written to the current working directory under a new `crash-YYYY-MM-DD-hh-mm-ss` directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,13 +100,18 @@ option(SLANG_RHI_ENABLE_CPU "Enable CPU backend" ON)
 cmake_dependent_option(SLANG_RHI_ENABLE_D3D11 "Enable D3D11 backend" ON "SLANG_RHI_HAS_D3D11" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_D3D12 "Enable D3D12 backend" ON "SLANG_RHI_HAS_D3D12" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_AGILITY_SDK "Enable Agility SDK" ON "SLANG_RHI_HAS_AGILITY_SDK" OFF)
-cmake_dependent_option(SLANG_RHI_ENABLE_AFTERMATH "Enable Aftermath support" ON "SLANG_RHI_HAS_AFTERMATH" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_NVAPI "Enable NVAPI support" ON "SLANG_RHI_HAS_NVAPI" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_VULKAN "Enable Vulkan backend" ON "SLANG_RHI_HAS_VULKAN" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_METAL "Enable Metal backend" ON "SLANG_RHI_HAS_METAL" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_CUDA "Enable CUDA backend" ON "SLANG_RHI_HAS_CUDA" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_OPTIX "Enable OptiX support" ON "SLANG_RHI_HAS_CUDA" OFF)
 cmake_dependent_option(SLANG_RHI_ENABLE_WGPU "Enable WebGPU backend" ON "SLANG_RHI_HAS_WGPU" OFF)
+option(SLANG_RHI_ENABLE_AFTERMATH "Enable Aftermath support" OFF)
+
+# Check Aftermath availability
+if(SLANG_RHI_ENABLE_AFTERMATH AND NOT SLANG_RHI_HAS_AFTERMATH)
+    message(FATAL_ERROR "Aftermath SDK is not available on this platform")
+endif()
 
 # Enabled OptiX versions
 if(NOT DEFINED SLANG_RHI_OPTIX_VERSIONS)


tests/testing.h
```diff
@@ -326,6 +327,8 @@ struct DeviceExtraOptions
     // (default), uses automatic detection.
     // This value is passed to D3D12DeviceExtendedDesc::highestShaderModel.
     uint32_t d3d12HighestShaderModel = 0;
+
+    bool enableAftermath = false;
```
jhelferty-nv, line 331:
```
aftermath-tdr.d3d11                                             SKIPPED (manual test only)
aftermath-tdr.d3d12                                             SKIPPED (manual test only)
aftermath-tdr.vulkan                                            SKIPPED (manual test only)
```
Are we planning to enable test coverage for Aftermath in CI? (low incidence coverage of some sort?)

src/vulkan/vk-command.cpp
```diff
@@ -2057,11 +2084,24 @@ CommandBufferImpl::CommandBufferImpl(Device* device, CommandQueueImpl* queue)
...
+#endif
 }
 
 CommandBufferImpl::~CommandBufferImpl()
 {
     DeviceImpl* device = getDevice<DeviceImpl>();
+#if SLANG_RHI_ENABLE_AFTERMATH
+    if (device->m_aftermathCrashDumper)
+    {
+        device->m_aftermathCrashDumper->registerMarkerTracker(&m_aftermathMarkerTracker);
```
jhelferty-nv: Should be unregistering?

src/aftermath.cpp
```diff
@@ -0,0 +1,299 @@
...
+            {
+                GFSDK_Aftermath_SpirvCode spirv = {};
+                spirv.pData = shader.code;
+                spirv.size = shader.codeSize;
+                GFSDK_Aftermath_ShaderBinaryHash shaderHash = {};
+                GFSDK_Aftermath_GetShaderHashSpirv(GFSDK_Aftermath_Version_API, &spirv, &shaderHash);
+                shader.hash = shaderHash.hash;
+            }
+        }
+#endif
```
jhelferty-nv, line 234: This `#endif` should be up a line?

src/aftermath.cpp
```diff
@@ -0,0 +1,299 @@
...
+    char name[128];
+    snprintf(name, sizeof(name), "%08x-%08x.nvdbg", (uint32_t)identifier.id[0], (uint32_t)identifier.id[1]);
+    std::filesystem::path path = dumpDir / name;
+    writeFile(path, pShaderDebugInfo, shaderDebugInfoSize);
+}
+
+static void descriptionCallback(PFN_GFSDK_Aftermath_AddGpuCrashDumpDescription addDescription, void* pUserData)
+{
+    // AftermathCrashDumper* dumper = reinterpret_cast<AftermathCrashDumper*>(pUserData);
+    // addDescription(GFSDK_Aftermath_GpuCrashDumpDescriptionKey_ApplicationName, "slang-rhi");
```
jhelferty-nv, line 131: Is this instended to be commented out?

src/d3d11/d3d11-device.cpp
```diff
@@ -146,22 +150,45 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
...
+        else
         {
-            printWarning("Failed to initialize aftermath: %d\n", int(initResult));
+            printWarning("Failed to initialize Aftermath: %d\n", int(initResult));
         }
     }
+#else
+    if (desc.enableAftermath)
+    {
+        printWarning("Aftermath requested but not enabled in build.");
```
jhelferty-nv: missing newline

src/aftermath.h
```diff
@@ -0,0 +1,115 @@
...
+    };
+
+    struct MarkerEntry
+    {
+        std::string name;
+        uint64_t hash;
+    };
+
+    MarkerName m_markerName;
+    std::array<MarkerEntry, 16> m_entries;
```
jhelferty-nv: Asking because I don't know.. is 16 always a good enough number, or is it worth making this configurable?

tests/testing.h
skallweitNV: This is something we can consider in the future I think. It feels a bit risky to run the GPU into a TDR during CI. It's even debatable if we want to enable Aftermath in the first place, or if we should have some specialized CI pipeline that builds with Aftermath and have the main one disable it. For now I enabled fetching and building with Aftermath by default, but we might decide that it's better to keep the default building without it.

src/vulkan/vk-command.cpp
skallweitNV: Yes it probably should, will fix this.

src/aftermath.cpp
skallweitNV: Yes! Good catch

src/d3d11/d3d11-device.cpp
skallweitNV: Again, good catch! Fixed in all backends.

src/aftermath.h
skallweitNV: Well, depends on how many markers are added to a command list. Increased it to 64 which should cover a lot.

src/aftermath.cpp
skallweitNV: Yeah, doesn't make a lot of sense to use `slang-rhi` as the application name. But we don't have anything from the client we could use here.

tests/testing.h
jhelferty-nv: I see. If we have lower-rate testing (per-day or per-week) I'd advocate for including at least one TDR test in that test plan to make sure it has some actual coverage. This feels like one of the more brittle pieces of functionality, and while I agree with your point that testing it is more risky, that seems like a good argument for making sure *something* is testing it regularly, vs. not at all. 😊

skallweitNV: I think ideally we would do the following:
- Disable Aftermath support by default (it incurs a minimal runtime overhead).
- Add additional pipelines (nightly or not) on the supported platforms that both enable Aftermath and run the test case.

skallweitNV: Added https://github.com/shader-slang/slang-rhi/issues/606 for tracking this

PR #602, opened 2025-12-12
Add spirv_nv capability for Vulkan SER support

szihs:
Add spirv_nv capability when SER is detected on Vulkan to fix HitObject compilation errors.

### DIFF
--- a/src/vulkan/vk-device.cpp
+++ b/src/vulkan/vk-device.cpp
@@ -843,6 +843,7 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
                 availableFeatures.push_back(Feature::ShaderExecutionReordering);
                 availableCapabilities.push_back(Capability::SPV_NV_shader_invocation_reorder);
                 availableCapabilities.push_back(Capability::spvShaderInvocationReorderNV);
+                availableCapabilities.push_back(Capability::spirv_nv);
             }
         );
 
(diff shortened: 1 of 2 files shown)


src/vulkan/vk-device.cpp
```diff
@@ -843,6 +843,7 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
                 availableFeatures.push_back(Feature::ShaderExecutionReordering);
                 availableCapabilities.push_back(Capability::SPV_NV_shader_invocation_reorder);
                 availableCapabilities.push_back(Capability::spvShaderInvocationReorderNV);
+                availableCapabilities.push_back(Capability::spirv_nv);
```
skallweitNV, line 846: Not sure what `spirv_nv` exactly stands for but is this really just available when SER is available? Isn't this always available on NVIDIA GPUs?

szihs: We had to support both EXT and NV variant for SER Extension.

  /// Represents SPIRV targeting NVIDIA-specific SER (for backward compatibility).
  /// [Target]
  def spirv_nv : spirv + spvShaderInvocationReorderNV;

  The spirv_nv capability is specifically defined as SPIRV + SER (NV extension). 
  The __target_switch cases in HitObject.MakeHit use case spirv_nv: because those code paths require SER functionality.

  ---
https://github.com/shader-slang/slang/blob/d67d3819be1c4f86c3099d932a2eb1274b1735dd/source/slang/hlsl.meta.slang#L21458-L21464

skallweitNV: Ok then it makes sense where it is set. All good, thanks for the explanation.

PR #607, opened 2025-12-24
Add Android platform support for Vulkan backend

guoxx:
- Load libvulkan.so instead of libvulkan.so.1 on Android
- Define VK_USE_PLATFORM_ANDROID_KHR for Android builds
- Use vkCreateAndroidSurfaceKHR for surface creation on Android
- Add VK_KHR_ANDROID_SURFACE_EXTENSION_NAME to instance extensions

### DIFF
--- a/src/vulkan/vk-api.h
+++ b/src/vulkan/vk-api.h
@@ -8,7 +8,7 @@
 #elif SLANG_APPLE_FAMILY
 #define VK_USE_PLATFORM_METAL_EXT 1
 #elif SLANG_LINUX_FAMILY
-#ifdef SLANG_ANDROID
+#if SLANG_ANDROID
 #define VK_USE_PLATFORM_ANDROID_KHR 1
 #else
 #define VK_USE_PLATFORM_XLIB_KHR 1
@@ -196,7 +196,7 @@ protected:
     x(vkCreateMetalSurfaceEXT) \
     /* */
 #elif SLANG_LINUX_FAMILY
-    #ifdef SLANG_ANDROID
+    #if SLANG_ANDROID
         #   define VK_API_INSTANCE_PLATFORM_KHR_PROCS(x)          \
             x(vkCreateAndroidSurfaceKHR) \
             /* */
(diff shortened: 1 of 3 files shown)


src/vulkan/vk-api.h
```diff
@@ -8,8 +8,12 @@
 #elif SLANG_APPLE_FAMILY
 #define VK_USE_PLATFORM_METAL_EXT 1
 #elif SLANG_LINUX_FAMILY
+#ifdef SLANG_ANDROID
```
skallweitNV: Most `SLANG_XXX` macros are defined as `0` or `1`, so you have to use `#if` instead of `#ifdef` here.

src/vulkan/vk-api.h
```diff
@@ -192,9 +196,15 @@ struct VulkanModule
     x(vkCreateMetalSurfaceEXT) \
     /* */
 #elif SLANG_LINUX_FAMILY
-#   define VK_API_INSTANCE_PLATFORM_KHR_PROCS(x)          \
-    x(vkCreateXlibSurfaceKHR) \
-    /* */
+    #ifdef SLANG_ANDROID
```
skallweitNV: Use `#if SLANG_ANDROID`

src/vulkan/vk-device.cpp
```diff
@@ -309,7 +309,11 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
 #elif SLANG_APPLE_FAMILY
         instanceExtensions.push_back(VK_EXT_METAL_SURFACE_EXTENSION_NAME);
 #elif SLANG_LINUX_FAMILY
+#ifdef SLANG_ANDROID
```
skallweitNV: Use `#if SLANG_ANDROID`

src/vulkan/vk-surface.cpp
```diff
@@ -61,6 +61,17 @@ Result SurfaceImpl::init(DeviceImpl* device, WindowHandle windowHandle)
         break;
     }
 #elif SLANG_LINUX_FAMILY
+
+#ifdef SLANG_ANDROID
```
skallweitNV: Use `#if SLANG_ANDROID`

skallweitNV (changes requested):
Thanks for the submission. Looks good overall but the detection of the Android platform is not correct yet, hence a lot of the tests are failing in CI.

Sorry for the late review, we were all out on vacation

src/vulkan/vk-api.h
guoxx: I replaced all #ifdef SLANG_ANDROID with #if SLANG_ANDROID. Thanks for the clarification.

PR #608, opened 2025-12-24
Fix MSVC build errors in slang-rhi for C++20 compatibility

lujinwangnv:
- arena-allocator.h: Replace deprecated std::is_pod_v with std::is_trivially_copyable_v (fixes C4996)
- d3d11-shader-object.h: Fix illegal qualified name in member declaration (fixes C4596)
- CMakeLists.txt: Add /EHsc to enable C++ exception handling (fixes C4530 warning when using std::vector and other STL containers)

Required for building with MSVC 17.7+ with warnings-as-errors enabled and c++20.

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -2068,6 +2068,7 @@ enum class WindowHandleType
     HWND,
     NSWindow,
     XlibWindow,
+    AndroidWindow,
 };
 
 struct WindowHandle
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -518,6 +518,7 @@ target_compile_options(slang-rhi-warnings INTERFACE
         /wd4702 # unreachable code
         /wd4127 # conditional expression is constant
         /wd4389 # signed/unsigned mismatch
+        /EHsc   # Standard C++ exception handling
```
skallweitNV, line 521: Technically we should probably keep this out of `slang-rhi-warnings` target but it's fine for now.

PR #609, opened 2025-12-27
Enable WebGPU backend for Emscripten builds

j8asic:
Added support for compiling `slang-rhi` using Emscripten (WASM), while enabling the WebGPU backend and handling dependencies for the web platform:

- Added logic to detect emscripten
- enable `SLANG_RHI_HAS_WGPU` and disable incompatible native APIs when building for Emscripten
- Skipped fetching the native Dawn binary package when building for Emscripten
- target_compile_options use modern `--use-port=emdawnwebgpu` for Emscripten builds
- core rhi code switches between dawn/emscripten functionality

## Summary by CodeRabbit

* **New Features**
  * Emscripten/WebAssembly build presets and CI steps
  * WebGPU (WASM) backend with canvas-based surface and window-handle canvas selector
  * Buffer APIs extended to accept native and shared handles
  * New wasm test program and minimal HTML host page

* **Compatibility**
  * WASM-specific texture/format handling and compiler/linker adjustments
  * WASM-target warning and linker flag tweaks

* **Chores**
  * Tooling improved for WebGPU proc extraction
  * .gitignore updated for emscripten build dir

* **Bug Fixes**
  * Safer init/cleanup and synchronous wait handling for WASM


include/slang-rhi.h
```diff
@@ -2097,6 +2098,13 @@ struct WindowHandle
         handle.handleValues[1] = xwindow;
         return handle;
     }
+    static WindowHandle fromWGPUCanvas(const char* canvasSelector)
```
skallweitNV, line 2172: Referencing a string like that dangerous. The handle might outlive the string allocation.

src/core/arena-allocator.h
```diff
@@ -78,7 +78,9 @@ class ArenaAllocator
         uintptr_t begin;
         uintptr_t end;
     };
+#if INTPTR_MAX == INT64_MAX
     static_assert(sizeof(Page) == 32);
```
skallweitNV: I wonder if we need this static_assert to begin with. Maybe better to just remove it. There is nothing that requires the page struct to be 32 bytes.

src/core/platform.cpp
```diff
@@ -4,7 +4,7 @@
 
 #if SLANG_WINDOWS_FAMILY
 #include <windows.h>
-#elif SLANG_LINUX_FAMILY || SLANG_APPLE_FAMILY
+#elif SLANG_LINUX_FAMILY || SLANG_APPLE_FAMILY || defined(__EMSCRIPTEN__)
```
skallweitNV: It would be better to use `... || SLANG_WASM` here, but this will need an updated version of Slang with the following fix in: https://github.com/shader-slang/slang/pull/9502

CMakeLists.txt
```diff
@@ -198,6 +205,8 @@ if(NOT DEFINED SLANG_RHI_DAWN_URL)
         set(DAWN_ARCH "x86_64")
     elseif(SLANG_RHI_ARCHITECTURE MATCHES "aarch64|arm64")
         set(DAWN_ARCH "aarch64")
+    elseif(SLANG_RHI_ARCHITECTURE MATCHES "wasm32")
+        set(DAWN_ARCH "wasm32")
```
skallweitNV: You are not fetching dawn if compiling for WASM so this looks like some leftover that can be removed?

skallweitNV (changes requested):
Thanks for the submission. Overall this looks like it's going into the right direction.

FYI, the WebGPU backend in slang-rhi is very much a work-in-progress and very low priority for us at this point in time. We don't expect anything apart from the most trivial examples to run.

If we want to add support for WASM to slang-rhi, we need to add proper testing through CI. My main fear is that we get incompatible versions of Dawn and `emdawnwebgpu`. Do you know how `emdawnwebgpu` is versioned? Is it just implicitly versioned through the emscripten version?

include/slang-rhi.h
j8asic: agreed, added copying

src/core/platform.cpp
j8asic: replaced with __EMSCRIPTEN__  with SLANG_WASM all around

CMakePresets.json
```diff
@@ -72,6 +72,22 @@
...
+            "name": "emscripten",
+            "description": "Emscripten-based WebAssembly build",
+            "generator": "Ninja Multi-Config",
+            "binaryDir": "${sourceDir}/build.em",
+            "cacheVariables": {
+                "SLANG_RHI_BUILD_TESTS": "OFF",
+                "SLANG_RHI_BUILD_EXAMPLES": "OFF",
+                "SLANG_RHI_FETCH_SLANG": "OFF",
+                "SLANG_RHI_BUILD_FROM_SLANG_REPO": "ON",
+                "SLANG_RHI_SLANG_INCLUDE_DIR": "${sourceDir}/../slang/include",
```
skallweitNV: Should we instead add support for fetching the slang WASM release binaries?
Taking a look at the WASM release, there are no header files, so I guess this won't work. I haven't looked into the details, what would be the "right way" to deal with the slang dependency?

include/slang-rhi.h
```diff
@@ -2073,13 +2073,18 @@ enum class WindowHandleType
...
 };
 
 struct WindowHandle
 {
     WindowHandleType type = WindowHandleType::Undefined;
-    uint64_t handleValues[2];
+    union
+    {
+        uint64_t handleValues[2];
+        char canvasSelector[128];
```
skallweitNV: This is unfortunate as it increases the handle size. This might not be a big issue but it really is only used in the WASM case.
I wonder if it would be better to `#if SLANG_WASM` the `WGPUCanvas` enum value, `canvasSelector` field (remove the union) and `fromWGPUCanvas`.

src/wgpu/wgpu-api.h
```diff
@@ -4,30 +4,91 @@
...
 #include <dawn/webgpu.h>
+#else
+#include <webgpu/webgpu.h>
+#endif
 
 // clang-format off
 
-#define SLANG_RHI_WGPU_PROCS(x) \
-    x(AdapterInfoFreeMembers) \
+// Define native-only procedures that are not available or different in Emscripten
```
skallweitNV: I don't really like this separation because I currently use the `tools/extract-webgpu-procs.py` script to generate the `SLANG_RHI_WGPU_PROCS(x)` definition whenever we update Dawn. This will make this tedious.

Maybe just define all of `SLANG_RHI_WGPU_PROCS` for both the WASM and Dawn case.

CMakeLists.txt
```diff
@@ -184,8 +191,8 @@ if(NOT DEFINED SLANG_RHI_METAL_CPP_URL)
     set(SLANG_RHI_METAL_CPP_URL "https://developer.apple.com/metal/cpp/files/metal-cpp_macOS14.2_iOS17.2.zip")
 endif()
 
-# Dawn WebGPU
-if(NOT DEFINED SLANG_RHI_DAWN_URL)
+# Dawn WebGPU (not used for Emscripten - uses emdawnwebgpu port instead)
+if(NOT DEFINED SLANG_RHI_DAWN_URL AND NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
```
skallweitNV: I would remove this change, this doesn't affect anything in the WASM build where Dawn is not fetched.

CMakeLists.txt
```diff
@@ -563,10 +574,14 @@ target_compile_options(slang-rhi-warnings INTERFACE
...
-    $<$<CXX_COMPILER_ID:Clang>:-Werror>
-    $<$<CXX_COMPILER_ID:AppleClang>:-Werror>
 )
+if(NOT EMSCRIPTEN)
+    target_compile_options(slang-rhi-warnings-as-errors INTERFACE
+        $<$<CXX_COMPILER_ID:GNU>:-Werror>
+        $<$<CXX_COMPILER_ID:Clang>:-Werror>
+        $<$<CXX_COMPILER_ID:AppleClang>:-Werror>
+    )
+endif()
```
skallweitNV, line 639: I don't like this too much, we should also have warning as errors in the emscripten build, no?

CMakePresets.json
j8asic: Yes exactly, this is just temporary; as soon as my PR gets merged we can directly fetch and use emscripten libs and headers: https://github.com/shader-slang/slang/pull/9556

CMakeLists.txt
j8asic: without this, it enters the body and gives unsupported arch error. if i put some checks inside this body, it would still declare some dawn stuff, so i think it is cleaner as currently proposed

include/slang-rhi.h
j8asic: sure, will do. this was lazy from my side

CMakeLists.txt
j8asic: agreed, i suggest to just add `$<$<BOOL:${EMSCRIPTEN}>:-Wno-c2y-extensions>` to slang-rhi-warnings

src/wgpu/wgpu-api.h
j8asic: This is the main issue for maintenance. I iterated two-three different solutions and found a compromise as you say to "define all of SLANG_RHI_WGPU_PROCS": since you already have a workflow, I extended your script to fetch (according commit of the) EM `webgpu.h`, and use it to typedef missing stubs.

CMakePresets.json
skallweitNV: ok, that's great then!

src/wgpu/wgpu-device.cpp
```diff
@@ -97,6 +97,7 @@ static inline Result createWGPUAdapter(API& api, WGPUInstance instance, WGPUAdap
         WGPUFuture future = api.wgpuInstanceRequestAdapter(instance, &options, callbackInfo);
 #if SLANG_WASM
         // Poll with wgpuInstanceProcessEvents and emscripten_sleep
+        (void)future;
```
skallweitNV: Would be better to use `SLANG_UNUSED(future)` here

src/wgpu/wgpu-command.cpp
```diff
@@ -640,6 +654,18 @@ void CommandRecorder::cmdDrawIndexedIndirect(const commands::DrawIndexedIndirect
         cmd.countBuffer ? checked_cast<BufferImpl*>(cmd.countBuffer.buffer)->m_buffer : nullptr,
         cmd.countBuffer.offset
     );
+#else
```
skallweitNV, line 652: Seems like a better idea to not draw at all if the API is not supported.

src/wgpu/wgpu-api.h
skallweitNV: Why can't we extract two sets of stubs, one from the currently used Dawn library, the other from the webgpu headers? Maybe that's the wrong approach, but I'm not sure what mixing the two buys us?!

j8asic: Ok, to be straightforward the py script now explicitly defines two sets of procs. Em list is loaded from statically linkage, Dawn dynamically. Initial tests for creating the adapter, device, queue work nicely

j8asic:
Added explicit distinction of dawn/em API to be loaded, and tests are running nicely. N.b. I have not created any html tests in the repo, to be run. If you made a decision on how to approach automatic testing let me know.

Please re-review, IMHO this can now be merged, as we are running late with WASM-libs PR (https://github.com/shader-slang/slang/pull/9818). When done, I will add auto fetching from Slang releases.

Btw I have SGL almost running via WASM

skallweitNV:
@j8asic sorry for not responding earlier. I will take another look tomorrow hopefully. Thanks for all the effort on this and nice to hear SGL almost works in WASM :)

j8asic:
SGL now works in browsers...
image

src/wgpu/wgpu-buffer.cpp
```diff
@@ -4,6 +4,10 @@
 
 #include "core/deferred.h"
 
+#ifdef SLANG_WASM
```
skallweitNV: Instead of splatting the `<emscripten.h>` include everywhere, we could just put it into `wgpu-api.h` once if we need emscripten API in the WASM case in different places.

src/wgpu/wgpu-buffer.cpp
```diff
@@ -138,9 +175,23 @@ Result DeviceImpl::mapBuffer(IBuffer* buffer, CpuAccessMode mode, void** outData
...
+    uint64_t timeoutNS = 0;
+    WGPUWaitStatus waitStatus = WGPUWaitStatus_Success;
+    while (true)
+    {
+        waitStatus = m_ctx.api.wgpuInstanceWaitAny(m_ctx.instance, SLANG_COUNT_OF(futures), futures, timeoutNS);
+        if (futures[0].completed)
+            break;
+#ifdef __EMSCRIPTEN__
+        emscripten_sleep(1);
+#endif
```
skallweitNV: This is used in a few places, should we wrap this into a helper and put it into `wgpu-wasm-utils.h` or something? Then actually the emscripten include could go there as well.

tools/extract-webgpu-procs.py
```diff
@@ -1,29 +1,106 @@
 from pathlib import Path
```
skallweitNV, line 1: Changes here look good. Thanks for moving towards generating the two sets of procs.

CMakeLists.txt
```diff
@@ -197,8 +204,8 @@ if(NOT DEFINED SLANG_RHI_METAL_CPP_URL)
     set(SLANG_RHI_METAL_CPP_URL "https://developer.apple.com/metal/cpp/files/metal-cpp_macOS14.2_iOS17.2.zip")
 endif()
 
-# Dawn WebGPU
-if(NOT DEFINED SLANG_RHI_DAWN_URL)
+# Dawn WebGPU (not used for Emscripten - uses emdawnwebgpu port instead)
+if(NOT DEFINED SLANG_RHI_DAWN_URL AND NOT CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
```
skallweitNV: Should this also use `AND NOT EMSCRIPTEN`?

CMakeLists.txt
```diff
@@ -474,6 +481,10 @@ if(SLANG_RHI_ENABLE_WGPU)
         copy_file(${dawn_SOURCE_DIR}/lib/libdawn.dylib .)
     endif()
     target_link_libraries(slang-rhi PRIVATE slang-rhi-dawn)
+elseif(SLANG_RHI_ENABLE_WGPU AND EMSCRIPTEN)
+    # For Emscripten, WebGPU headers are provided by emdawnwebgpu port
+    # Add compile option to use the port which provides include paths
+    target_compile_options(slang-rhi PRIVATE "--use-port=emdawnwebgpu")
 endif()
```
skallweitNV: I would just move this to a second if block commented with `# Link with Emscripten WebGPU`. The `elseif` doesn't really make sense with the `# Fetch & setup Dawn WebGPU`

skallweitNV (changes requested):
LGTM. Some suggestions. I think especially the code duplication should be reduced unless there are slight differences that I didn't notice.

src/wgpu/wgpu-buffer.cpp
j8asic: good stuff. i managed to unify the logic into a utils function that should be also optimized for browsers as well. pushing soon

skallweitNV:
@j8asic Looks like there are bunch of compilation failures. Can you also look at the review comments from coderabbit? I'll take another look when you feel it's ready.

CMakeLists.txt
```diff
@@ -264,6 +271,16 @@ endmacro()
...
     set(SLANG_VERSION ${SLANG_RHI_FETCH_SLANG_VERSION})
+
+    if(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")
+        # Emscripten WASM releases are available starting from v2026.3.1
+        set(SLANG_WASM_MIN_VERSION "2026.3.1")
+        if(SLANG_VERSION VERSION_LESS SLANG_WASM_MIN_VERSION)
+            message(STATUS "Slang ${SLANG_VERSION} did not release WASM libraries, overriding to ${SLANG_WASM_MIN_VERSION}")
+            set(SLANG_VERSION ${SLANG_WASM_MIN_VERSION})
+        endif()
+    endif()
```
skallweitNV: Can we remove this? slang-rhi is now upated to use 2026.3.1 so this should be obsolete.

CMakeLists.txt
```diff
@@ -328,6 +353,17 @@ if(NOT SLANG_RHI_BUILD_FROM_SLANG_REPO)
                 INTERFACE_INCLUDE_DIRECTORIES ${SLANG_RHI_SLANG_INCLUDE_DIR}
                 IMPORTED_LOCATION ${SLANG_RHI_SLANG_BINARY_DIR}/lib/lib${SLANG_LIB_NAME}.dylib
             )
+        elseif(EMSCRIPTEN)
```
skallweitNV: can we use `elseif(CMAKE_SYSTEM_NAME STREQUAL "Emscripten")` here for consistency?

CMakeLists.txt
```diff
@@ -946,20 +988,36 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED)
...
+        target_compile_features(wasm-test PRIVATE cxx_std_17)
+        target_link_libraries(wasm-test PRIVATE slang-rhi slang)
+        set_target_properties(wasm-test PROPERTIES 
+            OUTPUT_NAME "wasm-test"
+            SUFFIX ".html"
+        )
+        target_link_options(wasm-test PRIVATE 
+            "--use-port=emdawnwebgpu"
+            "--shell-file" "${CMAKE_CURRENT_SOURCE_DIR}/examples/wasm/index.html"
+        )
```
skallweitNV, line 1035: Not super happy about this .. I think the goal should be that examples generally also work on WASM. But we can leave this for now.

CMakePresets.json
```diff
@@ -78,6 +78,21 @@
...
+            "description": "Emscripten-based WebAssembly build",
+            "generator": "Ninja Multi-Config",
+            "binaryDir": "${sourceDir}/build.em",
+            "toolchainFile": "$env{EMSCRIPTEN_ROOT}/cmake/Modules/Platform/Emscripten.cmake",
+            "cacheVariables": {
+                "SLANG_RHI_BUILD_TESTS": "OFF",
+                "SLANG_RHI_BUILD_EXAMPLES": "OFF",
+                "SLANG_RHI_ENABLE_CPU": "OFF",
+                "CMAKE_C_FLAGS_INIT": "-fwasm-exceptions -Os",
+                "CMAKE_CXX_FLAGS_INIT": "-fwasm-exceptions -Os",
```
skallweitNV: Do we want `-Os` size optimizations?

CMakePresets.json
```diff
@@ -78,6 +78,21 @@
...
+                "CMAKE_EXE_LINKER_FLAGS": "-sASSERTIONS -sALLOW_MEMORY_GROWTH -fwasm-exceptions --export=__cpp_exception -sASYNCIFY -sASYNCIFY_IGNORE_INDIRECT=1 -sASYNCIFY_ONLY=['*main*','*wait*','*initialize*','*map*','*readBuffer*','*createBuffer*']"
```
skallweitNV: What does `ASYNCIFY` do? Feels bad to have to add this to the linker flags ...

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 .vscode/
 build/
+build.em/
```
skallweitNV: Why does emscripten build not use `build` folder?

src/wgpu/wgpu-utils.h
```diff
@@ -31,4 +31,10 @@ WGPUBlendOperation translateBlendOperation(BlendOp op);
 WGPULoadOp translateLoadOp(LoadOp op);
 WGPUStoreOp translateStoreOp(StoreOp op);
 
+struct Context;
+typedef Context WGPUContext;
```
skallweitNV: Can't we just use `Context` without introducing a type alias?

src/wgpu/wgpu-buffer.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "core/deferred.h"
 
+
```
skallweitNV: extra newline

src/wgpu/wgpu-command.cpp
```diff
@@ -6,6 +6,7 @@
 #include "wgpu-shader-object.h"
 #include "wgpu-utils.h"
 
+
```
skallweitNV: extra newline

src/wgpu/wgpu-device.cpp
```diff
@@ -56,42 +54,59 @@ static inline Result createWGPUAdapter(API& api, WGPUInstance instance, WGPUAdap
...
     options.nextInChain = &togglesDesc.chain;
+#endif
 
-    WGPUAdapter adapter = {};
+    struct AdapterRequestState
     {
         WGPURequestAdapterStatus status = WGPURequestAdapterStatus(0);
+        WGPUAdapter adapter = nullptr;
+    };
+    AdapterRequestState* state = new AdapterRequestState();
```
skallweitNV: Does this need to be heap allocated?

skallweitNV (changes requested):
Mostly looks good now.

CMakeLists.txt
j8asic: ok that is good news, this was temp till rhi gets updated

CMakePresets.json
j8asic: no

CMakePresets.json
j8asic: this is the price for using sync functions without freezing the browser, and this is the best outcome i.e. to tag _only specific functions_ needing to use asyncify. Otherwise the file size would be large, and compilation is slow

.gitignore
j8asic: TBH don't know, you tell me (taken this practice from the Slang repo)

src/wgpu/wgpu-device.cpp
j8asic: I've read about the browser caveats, and went with safer option not to trust asyncify stack saving. But maybe newer emscripten and browsers have it under control. Let me test the stack version and then update

CMakeLists.txt
j8asic: i would rather try to make another PR for testing (iirc you wrote you need to think how to handle automatic testing?)

skallweitNV: yes, this is fine for now

.gitignore
skallweitNV: is our build ending up in `build.em` at all?

j8asic: yes `"binaryDir": "${sourceDir}/build.em",`

src/wgpu/wgpu-buffer.cpp
```diff
@@ -65,6 +65,7 @@ Result DeviceImpl::createBuffer(const BufferDesc& desc_, const void* initData, I
     buffer->m_buffer = m_ctx.api.wgpuDeviceCreateBuffer(m_ctx.device, &bufferDesc);
     if (!buffer->m_buffer)
     {
+        *outBuffer = nullptr;
```
skallweitNV, line 68: I think client code should not rely on this. Is this needed?

src/wgpu/wgpu-command.cpp
```diff
@@ -625,13 +625,18 @@ void CommandRecorder::cmdDrawIndirect(const commands::DrawIndirect& cmd)
         cmd.countBuffer ? checked_cast<BufferImpl*>(cmd.countBuffer.buffer)->m_buffer : nullptr,
         cmd.countBuffer.offset
     );
+#else
+    SLANG_UNUSED(cmd);
+    NOT_SUPPORTED(S_RenderPassEncoder_drawIndirect);
```
skallweitNV: this will need to be changed to `NOT_SUPPORTED(IRenderPassEncoder, drawIndirect);` when updating to latest `main`.

src/wgpu/wgpu-command.cpp
```diff
@@ -640,6 +645,10 @@ void CommandRecorder::cmdDrawIndexedIndirect(const commands::DrawIndexedIndirect
         cmd.countBuffer ? checked_cast<BufferImpl*>(cmd.countBuffer.buffer)->m_buffer : nullptr,
         cmd.countBuffer.offset
     );
+#else
+    SLANG_UNUSED(cmd);
+    NOT_SUPPORTED(S_RenderPassEncoder_drawIndexedIndirect);
```
skallweitNV: `NOT_SUPPORTED(IRenderPassEncoder, drawIndexedIndirect);`

src/wgpu/wgpu-utils.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "core/assert.h"
 
+#include "wgpu-device.h"
```
skallweitNV: wgpu includes should come first, move this before or after `wgpu-utils.h`

skallweitNV:
@coderabbitai full review

skallweitNV:
@j8asic I'm doing a final pass over this myself and will merge it after. Thanks for all the work and patience with this!

j8asic:
@skallweitNV no worries, I have been OoO last month so thank you for finalising!

Issue #611, opened 2026-01-06
[Vulkan] GPU-AV vulkan validation error is hit when using a device-address aquired from an `AccelerationStructure`


16-Bit-Dog:
Problem:
`vkGetBufferDeviceAddress` is being used to get the device address of `AccelerationStructure`'s. This is not the correct way to get the device-address of a `AccelerationStructure`. This is seen through GPU-AV hitting a validation error.

Replication of bug:
This can be replicated by running `slang-test` from [slang](https://github.com/shader-slang/slang) with any(?) test (the function `_initializeAccelerationStructure()` is the culprit here).

When running `slang-test` ensure GPU-Asissted-Validation is enabled by launching Vulkan Configurator and ticking the needed check-box
<img width="1981" height="1102" alt="Image" src="https://github.com/user-attachments/assets/fe9f117b-279d-41c3-a0fa-38cc4dc64ef7" />

PR #612, opened 2026-01-06
Fix getting device-address of acceleration structure

16-Bit-Dog:
Fixes: #611

Problem that is being addressed:
`vkGetBufferDeviceAddress` is being used to get the device address of `AccelerationStructure`'s. This is not the correct way to get the device-address of an `AccelerationStructure`. This is seen through GPU-AV hitting a validation error.

Notable changes/additions:
1. For `AccelerationStructure` use the correct function `vkGetAccelerationStructureDeviceAddressKHR` instead of `vkGetBufferDeviceAddress`. Added a test to ensure the fix works.

Note: test was removed from the PR since a separate PR will include the functionality to enable validation, which is required to make a test that validates slang-rhi for GPU-AV.


include/slang-rhi.h
```diff
@@ -3019,6 +3029,11 @@ struct DeviceDesc
     bool enableValidation = false;
     /// Enable backend API raytracing validation layer (D3D12, Vulkan and CUDA).
     bool enableRayTracingValidation = false;
+    /// Enable backend gpu-assisted-validation layer
+    bool enableGPUAssistedValidation = false;
+    /// Severity of validation to emit
+    ValidationMessageSeverity validationMessageSeverityToEmit =
+        ValidationMessageSeverity::Error | ValidationMessageSeverity::Warning;
```
skallweitNV: If we want to expose more controls validation layers, we should clarify on what backends these settings apply and implement them where supported.

There is also some naming inconsistencies around "debug layers" and "validation layers". From slang-rhi perspective "debug layers" mean the backend debug/validation layers and "validation layer" means the internal validation of the RHI interface. I guess `enableRayTracingValidation` is already somewhat a bad name here.

All of this needs a bit more thought ..

skallweitNV (commented):
This PR contains two parts:
- The fix for getting the AS device address
- Extending vulkan validation layer setup and a new test case

I think we can merge the fix as-is. The work around validation layers probably needs some more thought on my behalf. @16-Bit-Dog Do you mind splitting this PR into two?

PR #615, opened 2026-01-08
Support for reserving data in shader object without writing it

ccummingsNV:


skallweitNV (dismissed):
LGTM. As discussed, `reserveData` needs a big "this is dangerous API" comment :)

skallweitNV (approved):
LGTM

PR #619, opened 2026-01-12
Block allocator

ccummingsNV:


src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+            m_freeList = block->next;
+            return reinterpret_cast<T*>(block);
+        }
+        return allocateFromNewPageLocked();
+#endif
+    }
+
+    /// Deallocate a block (thread safe).
+    /// @param ptr Pointer to block to deallocate
+    void deallocate(T* ptr)
```
skallweitNV: I think in most other places we use `allocate` and `free`, so might be worth using the same names here.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+
+    /// A page contains multiple blocks and a link to the next page
+    struct Page
+    {
+#if SLANG_RHI_LOCKLESS_BLOCK_ALLOCATOR
+        std::atomic<Page*> next;
+#else
+        Page* next;
+#endif
+        size_t blockCount;
```
skallweitNV: This feels like redundant information. Pages have a fixed number of blocks, no?

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+        Block blocks[1];
+    };
+
+    static_assert(sizeof(Block) >= sizeof(FreeBlock*), "Block must be large enough to hold a pointer");
```
skallweitNV: I would move these up to the `Block` definition.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+    /// A page contains multiple blocks and a link to the next page
+    struct Page
+    {
+#if SLANG_RHI_LOCKLESS_BLOCK_ALLOCATOR
+        std::atomic<Page*> next;
+#else
+        Page* next;
+#endif
+        size_t blockCount;
+        Block blocks[1];
```
skallweitNV, line 139: Maybe add a comment that the size here doesn't reflect the true size of a page.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+#if SLANG_RHI_LOCKLESS_BLOCK_ALLOCATOR
+    std::atomic<FreeBlock*> m_freeList{nullptr};
+    mutable std::mutex m_pageMutex; // Only for page allocation
+    std::atomic<Page*> m_pageListHead{nullptr};
+#else
+    FreeBlock* m_freeList{nullptr};
+    mutable std::mutex m_mutex; // Protects all operations
+    Page* m_pageListHead{nullptr};
+#endif
+    uint32_t m_numPages{0};
```
skallweitNV, line 175: Minor: I'd move this below `m_blocksPerPage` just to group things into common and config-dependent members.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+#include <mutex>
+#include <cstring>
+
+#ifndef SLANG_RHI_LOCKLESS_BLOCK_ALLOCATOR
+#define SLANG_RHI_LOCKLESS_BLOCK_ALLOCATOR 0
+#endif
+
+namespace rhi {
+
+/// Block allocator for fixed-size objects.
```
skallweitNV, line 8: Maybe mention that it allocates `T`-sized blocks.

src/shader-object.cpp
```diff
@@ -4,6 +4,12 @@
 
 namespace rhi {
 
+// ----------------------------------------------------------------------------
+// Allocators
+// ----------------------------------------------------------------------------
+
+SLANG_RHI_IMPLEMENT_BLOCK_ALLOCATED(ShaderObject, 512)
```
skallweitNV: The page size seems rather conservative. How big are these objects? I think it makes sense to allocate in MBs rather than KBs

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+#else
+    FreeBlock* m_freeList{nullptr};
+    mutable std::mutex m_mutex; // Protects all operations
+    Page* m_pageListHead{nullptr};
+#endif
+    uint32_t m_numPages{0};
+};
+
+// Macro to declare block allocator support for a class
+#define SLANG_RHI_DECLARE_BLOCK_ALLOCATED(ClassName, BlocksPerPage)                                                    \
```
skallweitNV: `BlocksPerPage` is not used here. I think I'd prefer specifying it only in the "implement" macro because otherwise it's really easy to get out of sync.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+    {                                                                                                                  \
+        return s_allocator.allocate();                                                                                 \
+    }                                                                                                                  \
+                                                                                                                       \
+    void ClassName::operator delete(void* ptr)                                                                         \
+    {                                                                                                                  \
+        if (!ptr)                                                                                                      \
+            return;                                                                                                    \
+        if (s_allocator.owns(ptr))                                                                                     \
```
skallweitNV: This check seems incorrect. AFAIK calling `delete` on any object not created through `new` is incorrect and UB. If you use placement new, you must call the destructor, not `delete` for example.

src/core/block-allocator.h
```diff
@@ -0,0 +1,370 @@
...
+    void operator delete(void* ptr);                                                                                   \
+                                                                                                                       \
+private:                                                                                                               \
+    static ::rhi::BlockAllocator<ClassName> s_allocator;
+
+// Macro to implement block allocator operators in .cpp file
+#define SLANG_RHI_IMPLEMENT_BLOCK_ALLOCATED(ClassName, BlocksPerPage)                                                  \
+    ::rhi::BlockAllocator<ClassName> ClassName::s_allocator(BlocksPerPage);                                            \
+                                                                                                                       \
+    void* ClassName::operator new(size_t size)                                                                         \
```
skallweitNV: We don't have use-cases for `new[]` I think, right? Or `new(count, alignment)`?

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+    SUBCASE("reuse-after-free")
+    {
+        // Allocate and free
+        TestObject* obj1 = allocator.allocate();
+        REQUIRE(obj1 != nullptr);
+        new (obj1) TestObject(1, 1.0);
+        obj1->~TestObject();
+        allocator.deallocate(obj1);
+
+        // Allocate again - should reuse the same block
```
skallweitNV, line 94: This is testing implementation details. Maybe a better test is to allocate multiple objects (number of blocks per page) and then make sure every one of them corresponds to a previous allocation but order doesn't matter.

But also, fine to leave as is.

skallweitNV: But a comment would be good IMO.

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+    // Free half of them
+    for (int i = 0; i < 4; ++i)
+    {
+        allocator.deallocate(objects[i]);
+    }
+
+    // Reset the allocator
+    allocator.reset();
+
+    // After reset, should be able to allocate all blocks again
```
skallweitNV, line 210: This doesn't really test anything other than the allocator still "works" after reset. No effects of reset are tested in this test.

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+
+        // Allocate
+        for (int i = 0; i < allocationsPerThread; ++i)
+        {
+            TestObject* obj = allocator.allocate();
+            if (obj)
+            {
+                new (obj) TestObject(i, i * 1.5);
+                localObjects.push_back(obj);
+                totalAllocations.fetch_add(1, std::memory_order_relaxed);
```
skallweitNV, line 260: It's good practice IMO to either specify the memory order on all accesses to an atomic or none of them. Otherwise it feels like there might be an oversight.

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+    // Mega test takes about 30 mins
+    // constexpr int numThreads = 16;
+    // constexpr int iterations = 100000;
+    // constexpr int objectsPerIteration = 10000;
+
+    auto threadFunc = [&]()
+    {
+        for (int iter = 0; iter < iterations; ++iter)
+        {
+            CHECK(allocator.getNumPages() < (numThreads * objectsPerIteration * 2) / 1000);
```
skallweitNV: `1000` should probably be `blocksPerPage` and also used when setting up the allocator

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+            }
+
+            // Deallocate half
+            for (size_t i = 0; i < objects.size() / 2; ++i)
+            {
+                if (objects[i])
+                {
+                    CHECK(objects[i]->value == static_cast<int>(i));
+                    bool thread_cmp = (objects[i]->thread == id);
+                    CHECK(thread_cmp); // doc test can't compile CHECK(objects[i]->thread == id);
```
skallweitNV: Yeah, but it could probably do `CHECK_EQ`

skallweitNV: `CHECK` uses fancy expression decomposition which can go wrong in lot's of different ways ..

tests/test-block-allocator.cpp
```diff
@@ -0,0 +1,507 @@
...
+        // Allocate using standard operator new (bypassing custom operator)
+        void* memory = ::operator new(sizeof(TestMacroClass));
+        TestMacroClass* obj = ::new (memory) TestMacroClass(200);
+
+        // Should NOT be owned by the allocator
+        CHECK_FALSE(TestMacroClass::getAllocator().owns(obj));
+
+        // Manually destruct and free
+        obj->~TestMacroClass();
+        ::operator delete(memory);
```
skallweitNV: This is UB

skallweitNV (changes requested):
Some comments, overall looks pretty good to me. I didn't however review the lockless implementation. I think we should either fix that implementation ASAP or remove to avoid confusion down the road. For now at least we should tell readers that the lockless implementation is not reviewed and has issues and should not be used.

src/core/block-allocator.h
ccummingsNV: Not that I know of, but these wouldn't be able to use the block allocator anyway. 

I guess alignment could hypothetically, if the block allocator aligned at a high level and it could validate, but that'd require wasting memory.

skallweitNV: Yeah, I think basic `new`/`delete` is probably all we need.

src/core/block-allocator.h
ccummingsNV: strictly speaking, it is possible for something (eg nanobind) to do ::new to allocate the class, so it is possible to get here without using UB. But we can catch it as an error if preferred

src/shader-object.cpp
ccummingsNV: This is object count, not bytes

src/core/block-allocator.h
ccummingsNV: Ah yeah - I was toying with growing the pages at one point

tests/test-block-allocator.cpp
ccummingsNV: It tests if reset crashes, which is a reasonable thing to test. I can go into more depth with the test if preferred - probably verify after reset, the same set of pointers are returned. But in its absence, any code coverage is better than no code coverage.

ccummingsNV: (and obvs whether it leaves allocator in a valid state)

src/shader-object.cpp
skallweitNV: Yeah but if for example the object is 128 bytes this I still just a 64KB page size.

tests/test-block-allocator.cpp
skallweitNV: Yeah, my point is just that reset could be a no-op and the test would still pass :)

tests/test-block-allocator.cpp
ccummingsNV: couldn't get CHECK_EQ to do it either so I stuck with a bool+comment

ccummingsNV: (I can't repro this compile error locally either!)

src/core/block-allocator.h
skallweitNV: not addressed, but no biggie

src/core/block-allocator.h
```diff
@@ -0,0 +1,204 @@
...
+private:                                                                                                               \
+    static ::rhi::BlockAllocator<ClassName> s_allocator;
+
+/// Macro to implement block allocator operators in .cpp file.
+/// @param ClassName The class name to implement block allocation for.
+/// @param BlocksPerPage Number of blocks to allocate per page.
+#define SLANG_RHI_IMPLEMENT_BLOCK_ALLOCATED(ClassName, BlocksPerPage)                                                  \
+    ::rhi::BlockAllocator<ClassName> ClassName::s_allocator(BlocksPerPage);                                            \
+                                                                                                                       \
+    void* ClassName::operator new(size_t size)                                                                         \
```
skallweitNV: Minor: `size` should be `count` according to C++ docs

skallweitNV (dismissed):
LGTM, just minor comments.

PR #620, opened 2026-01-12
Cooperative Vector improvements

skallweitNV:
Some improvements to the cooperative vector implementation:
- Do not align matrix sizes to 64 bytes on D3D12/Vulkan (this was done to make them behave similarly to OptiX, but this causing issues for clients)
- Manually compute matrix size for non-optimal (implementation defined) matrix layouts in the CUDA backend to match the D3D12/Vulkan implementation
- Check offset alignment only when using optimal layouts
- Load `VK_EXT_shader_float8` which is required for the 8-bit floating point formats
- More exhaustive testing of `getCooperativeVectorMatrixSize`
- More exhaustive testing of the matrix conversion API
- Add `enumToString` for `CooperativeVectorMatrixLayout`

### DIFF
--- a/src/cooperative-vector-utils.h
+++ b/src/cooperative-vector-utils.h
@@ -52,4 +52,34 @@ inline size_t getTightRowColumnStride(
     return 0;
 }
 
+/// Compute the matrix size in bytes for non-optimal (RowMajor/ColumnMajor) layouts.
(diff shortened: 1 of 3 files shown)


ccummingsNV (approved):
LGTM

PR #622, opened 2026-01-13
Expand on testing & debugging code + change naming/structure of debug options

16-Bit-Dog:
Note: this can be split into multiple PRs, I just want to clear the general design before-hand since some architectural changes were made (and some of the smaller changes made justify "the new design").

Problems that this PR addresses
* Currently the naming for downstream debug-layers vs. Slang-RHI debug layers can be confusing.
    * Changed the organization/logic of debug options.
        * `DebugLayerOptions` manages per slang-rhi instance (downstream debug layer) options such as `CoreValidation` and `GPUAssistedValidation`. These are set via `setDebugLayerOptions(DebugLayerOptions debugLayerOptions)`. `enableDebugLayers()` was removed since it is redundant if a user needs to choose their `DebugLayerOptions` anyways.
* Added missing `GPUAssistedValidation` option & disabling `CoreValidation` option
* Debug layers were never disabled with D3D12 (this could cause consistency issues): (1) enable debug for a test (2) create d3d12 device [debug layers are enabled] (3) destroy d3d12 device (4) create d3d12 device for a new test that should not have validation enabled [debug layers are still enabled]
    * This was fixed by adding logic to disable debug layers (and regenerate `DXGIFactory`'s as needed since only one may be created per program). To safely have this behaviour IRHI now tracks the number of live-devices so that when changing `DebugLayerOptions` we can produce a `SLANG_FAIL` for the user in case they have any live-devices when changing device options.
        * This also solves the potentially introduced issue where we may have a Debug D3D11 device, disable debug layers, and try to create a non-debug D3D12 device (using a debug `DXGIFactory` instead of a non-debug `DXGIFactory`).
* added `GPU_TEST_CASE_EX` to allow choosing which `DebugLayerOptions` a test should use.
    * `GPU_TEST_CASE_EX` is used to run the new `tests/test-acceleration-structure-creation-with-validation.cpp` test

### DIFF
--- a/tests/testing.cpp
+++ b/tests/testing.cpp
@@ -1074,7 +1074,7 @@ int registerGpuTest(
         info->deviceType = deviceType;
         info->flags = flags;
         info->hasDebugLayerOptions = debugLayerOptions.has_value();
-        info->debugLayerOptions = debugLayerOptions.value_or({});
+        info->debugLayerOptions = debugLayerOptions.value_or(DebugLayerOptions{});
 
         char* testName = reinterpret_cast<char*>(info + 1);
         snprintf(testName, testNameLen, "%s.%s", name, deviceTypeToString(deviceType));


include/slang-rhi.h
```diff
@@ -2967,6 +2979,23 @@ struct DeviceNativeHandles
     NativeHandle handles[3] = {};
 };
 
+/// Per device options that can be toggled to enhance debugging experience
+enum class DebugDeviceOptions
```
skallweitNV: While I can see the use of flags here, I'd stick with the `enableXXX` bools we currently have in `DeviceDesc`.
We might consider renaming `enableValidation` to `enableRHIValidation`. But I feel like we can keep the names and make sure "debug layers" are called "debug layers" and not "validation layers".
In any case, I would prefer to drop `DebugDeviceOptions`.

include/slang-rhi.h
```diff
@@ -3526,8 +3552,33 @@ class IPersistentCache : public ISlangUnknown
...
+protected:
+    friend class LiveDeviceTracker;
+
+    DebugLayerOptions debugLayerOptions = DebugLayerOptions::None;
+    int totalLiveDevices = 0;
+    /// Increment the total number of live devices created by this IRHI instance.
+    void incrementLiveDeviceCount() { totalLiveDevices++; }
+    /// Decrement the total number of live devices created by this IRHI instance
+    void decrementLiveDeviceCount() { totalLiveDevices--; }
+
```
skallweitNV: We want to keep `IRHI` as a COM-compatible interface. It should not have any data members, so the live device tracking should be moved to the implementation. In general I think it's good to have the "live-tracking" as this allows us to detect when it's OK to change debug layer configuration.

include/slang-rhi.h
```diff
@@ -3581,6 +3628,29 @@ class IRHI
...
+    }
+
+    /// Get current downstream debug layer options.
+    DebugLayerOptions getDebugLayerOptions() { return debugLayerOptions; }
+
+    /// Check if downstream debug layer option is set.
+    bool isDebugLayerOptionSet(DebugLayerOptions debugLayerOption)
+    {
+        return is_set(this->debugLayerOptions, debugLayerOption);
+    }
```
skallweitNV: Again, these should all be virtual functions except `isDebugLayerOptions` which could just be an "inline helper".

include/slang-rhi.h
```diff
@@ -3526,8 +3552,33 @@ class IPersistentCache : public ISlangUnknown
     virtual SLANG_NO_THROW Result SLANG_MCALL queryCache(ISlangBlob* key, ISlangBlob** outData) = 0;
 };
 
+/// Options for downstream API debug layers.
+enum class DebugLayerOptions
```
skallweitNV: We might consider making `DebugLayerOptions` a struct. Might be more future-proof. Something like
```c++
struct DebugLayerOptions
{
    /// Enable core debug layers (D3D12 | Vulkan).
    bool enableCoreValidation = true;
    /// Enable GPU assisted debug layer (D3D12 | Vulkan).
    bool enableGPUAssistedValidation = true;
};
```

tests/testing.h
```diff
@@ -428,29 +432,47 @@ enum GpuTestFlags
...
+    GPUAssistedValidation = DebugLayerOptions::GPUAssistedValidation,
+    AllValidation = DebugLayerOptions::All,
+};
+
 struct GpuTestInfo
 {
     GpuTestFunc func;
     DeviceType deviceType;
     GpuTestFlags flags;
+    GpuTestDebugLayerFlags debugLayerFlags;
```
skallweitNV: This could just be a `DebugLayersOptions` struct instead.

include/slang-rhi.h
```diff
@@ -2905,16 +2905,28 @@ struct DeviceInfo
...
+    All = Info | Warning | Error,
 };
+SLANG_RHI_ENUM_CLASS_OPERATORS(DebugMessageType)
+
 enum class DebugMessageSource
 {
-    Layer,
-    Driver,
-    Slang
+    None = 0,
```
skallweitNV: I think having a non-flags enum here is the better option. A message cannot originate from more than one source at the same time.

include/slang-rhi.h
```diff
@@ -2905,16 +2905,28 @@ struct DeviceInfo
 
 enum class DebugMessageType
 {
-    Info,
-    Warning,
-    Error
+    None = 0,
```
skallweitNV: Do we need flags for debug levels?

16-Bit-Dog: We do not need flags, I will remove them.

I felt it could be useful (slang-rhi-test uses it nicely since it simplifies specifying multiple message types), but at the end of the day it probably introduces more chances to make mistakes with minimal benefit of convience.

A simple `IntSet<DebugMessageType>`/`std::set<DebugMessageType>` does everything the flag does, and without chance to make an error.

include/slang-rhi.h
```diff
@@ -3633,4 +3672,58 @@ inline const FormatInfo& getFormatInfo(Format format)
...
+inline Result setDebugLayerOptions(DebugLayerOptions newDebugLayerOptions)
+{
+    return getRHI()->setDebugLayerOptions(newDebugLayerOptions);
+}
+
+/// Get current downstream debug layer options.
+inline DebugLayerOptions getDebugLayerOptions()
+{
+    return getRHI()->getDebugLayerOptions();
+}
```
skallweitNV: I'd prefer if we can remove these and instead call `getRHI()->xxx()`.

include/slang-rhi.h
```diff
@@ -3633,4 +3672,58 @@ inline const FormatInfo& getFormatInfo(Format format)
...
+    case DebugMessageSource::Driver:
+        return "Driver";
+    case DebugMessageSource::Slang:
+        return "Slang";
+    default:
+        // Missing case
+        assert(false);
+        return "Invalid";
+    }
+}
```
skallweitNV: We should probably merge these with the `enum-strings.h` header and make it a public header. This can wait, fine to leave these here for now.

include/slang-rhi.h
```diff
@@ -3633,4 +3672,58 @@ inline const FormatInfo& getFormatInfo(Format format)
...
+    {
+    case DebugMessageType::Info:
+        return "Info";
+    case DebugMessageType::Warning:
+        return "Warning";
+    case DebugMessageType::Error:
+        return "Error";
+    default:
+        // Missing case
+        assert(false);
```
skallweitNV: Lets not rely on standard asserts. Just remove those.

include/slang-rhi.h
```diff
@@ -3526,9 +3530,48 @@ class IPersistentCache : public ISlangUnknown
...
+            || GPUAssistedValidation == true;
+    }
+};
+
 class IRHI
 {
+protected:
+    friend class LiveDeviceTracker;
+    /// Increment the total number of live devices created by this IRHI instance.
+    virtual SLANG_NO_THROW void SLANG_MCALL incrementLiveDeviceCount() = 0;
```
skallweitNV: Let's not expose these on `IRHI` .. this is an implementation detail. The interface should only contain public methods.

include/slang-rhi.h
```diff
@@ -3526,9 +3530,48 @@ class IPersistentCache : public ISlangUnknown
...
+    virtual SLANG_NO_THROW void SLANG_MCALL decrementLiveDeviceCount() = 0;
+
 public:
+    /// Check is downstream debug layer is enabled
+    inline bool isDebugLayersEnabled() { return getDebugLayerOptions().isDebugLayersEnabled(); }
+
+    /// Change downstream debug layer options.
+    /// All devices must be released.
+    /// Fails if not all devices are released.
+    virtual SLANG_NO_THROW Result SLANG_MCALL setDebugLayerOptions(DebugLayerOptions newDebugLayerOptions) = 0;
```
```suggestion
    virtual SLANG_NO_THROW Result SLANG_MCALL setDebugLayerOptions(DebugLayerOptions options) = 0;
```

src/vulkan/vk-device.cpp
```diff
@@ -867,9 +887,14 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
...
         }
 
-        // Only enable raytracing validation if both requested and supported
-        if (desc.enableRayTracingValidation && extendedFeatures.rayTracingValidationFeatures.rayTracingValidation)
+        if (desc.enableRayTracingValidation)
         {
+            // If unsupported, fail
+            if (!extendedFeatures.rayTracingValidationFeatures.rayTracingValidation)
+            {
+                printError("Tried to enable raytracing validation but unsupported\n");
```
skallweitNV: We should probably make this a warning instead of an error.

src/d3d12/d3d12-device.cpp
```diff
@@ -219,27 +225,68 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         }
     }
 
-    if (isDebugLayersEnabled())
+    // Only interact with `ID3D12Debug` if we enabled or previously enabled debug layers.
```
skallweitNV, line 196: Maybe it would be cleaner to move the debug layer configuration into a separate helper function, something like `setupDebugLayer(options)`.

src/rhi.cpp
```diff
@@ -135,8 +134,16 @@ inline const FormatInfo& _getFormatInfo(Format format)
 
 class RHI : public IRHI
 {
+private:
+    DebugLayerOptions debugLayerOptions = {};
+    int totalLiveDevices = 0;
```
skallweitNV:
```suggestion
    DebugLayerOptions m_debugLayerOptions = {};
    int m_liveDeviceCount = 0;
```

src/rhi.cpp
```diff
@@ -135,8 +134,16 @@ inline const FormatInfo& _getFormatInfo(Format format)
...
+
+    virtual void incrementLiveDeviceCount() override { totalLiveDevices++; }
```
skallweitNV: You can keep these and make them public functions (on the `RHI` implementation, remove them completely from the `IRHI` interface).

src/device.h
```diff
@@ -145,6 +145,13 @@ class NullDebugCallback : public IDebugCallback
     }
 };
 
+class LiveDeviceTracker : public RefObject
+{
+public:
+    LiveDeviceTracker() { getRHI()->incrementLiveDeviceCount(); }
+    ~LiveDeviceTracker() { getRHI()->decrementLiveDeviceCount(); }
```
skallweitNV: You'd have to cast `IRHI` to `RHI` here. For extra safety, use `checked_cast<RHI>(getRHI())->xxx()` which will do additional checks in debug builds.

src/rhi.cpp
skallweitNV: Also, make the live count an atomic.

tests/main.cpp
```diff
@@ -32,10 +32,6 @@ int main(int argc, const char** argv)
 
     rhi::testing::cleanupTestTempDirectories();
 
-#if SLANG_RHI_DEBUG
```
skallweitNV, line 35: Do we still enable debug layers in debug builds of the tests (and not in release builds)?

tests/test-shader-cache.cpp
```diff
@@ -203,6 +203,7 @@ struct ShaderCacheTest
         // (And in general reduce the differences (and duplication) between
         // here and render-test-main.cpp)
 #if SLANG_RHI_DEBUG
+        setDebugLayerOptions({true});
```
skallweitNV: ?

tests/testing.h
```diff
@@ -428,29 +438,63 @@ enum GpuTestFlags
     DontCacheDevice = (1 << 11),  // Do not use cached devices (create a new device for this test case)
 };
 
+/// Struct that converts into DebugLayerOptions.
```
skallweitNV, line 436: I would probably replace this with a builder pattern. Something along the lines of:
```c++
class DebugLayerOptionsBuilder
{
public:
    DebugLayerOptionsBuilder& enableCoreValidation() { m_options.coreValidation = true; return *this; }
    DebugLayerOptionsBuilder& enableGPUAssistedValidation() { ... }
    ...
    operator DebugLayerOptions() const { return m_options; }
private:
    DebugLayerOptions m_options = {};
};
```

Then you can setup options in tests like this:
```c++
GPU_TEST_CASE_EX("acceleration-structure-creation-with-validation", Vulkan, DebugLayerOptionsBuilder().enableCoreValidation().enableGPUAssistedValidation())
```

Or you set it up at the top of the test and then use the variable. Somethling like
```c++
static DebugLayerOptions kDebugLayerOptions = DebugLayerOptionsBuilder().enableCoreValidation().enableGPUAssistedValidation();

GPU_TEST_CASE_EX("acceleration-structure-creation-with-validation", Vulkan, kDebugLayerOptions)
```

skallweitNV: This is mostly in anticipation of more debug layer settings that can be tweaked in the future, which might go beyond just flags.

tests/testing.h
```diff
@@ -464,7 +508,13 @@ const char* getSkipMessage(const doctest::TestCaseData* tc);
 // In addition to the device flags, the following flags can be used:
 // - GpuTestFlag::DontCreateDevice: Do not create a device (device argument is nullptr)
 // - GpuTestFlag::DontCacheDevice: Do not use cached devices (create a new device for this test case)
-#define GPU_TEST_CASE(name, flags) GPU_TEST_CASE_IMPL(name, DOCTEST_ANONYMOUS(GPU_TEST_ANONYMOUS_), flags)
+#define GPU_TEST_CASE(name, flags)                                                                                     \
+    GPU_TEST_CASE_IMPL(name, DOCTEST_ANONYMOUS(GPU_TEST_ANONYMOUS_), flags, CoreValidation)
```
skallweitNV: I'd like to continue running tests with debug layers in Debug builds and without debug layers in Release builds. So this won't work.

src/vulkan/vk-device.cpp
16-Bit-Dog: Since we are changing this to a warning, do we still return a `SLANG_FAIL` or do we continue on skipping the unsupported feature? I would assume we don't return a `SLANG_FAIL` since a warning implies that everything can still work.

Additionally, for the purpose of consistency, if debug-layer options are unsupported for a `d3d12-device`, I would believe it is best if the approach is consistent with our `vk-device`, so what-ever is done here I will replicate for d3d12-device logic.

tests/test-shader-cache.cpp
16-Bit-Dog: I will remove this change, I suppose there is no need for debug layers to be enabled here since it is supposed to be handled else where.

tests/main.cpp
16-Bit-Dog: The changes made currently do not have this behaviour.

I will change the logic so that it does the following:
1. if release build --> never enable debug-layers.
2. if debug build --> by default tests runs with `coreValidation` enabled. `DebugLayerOptions` set by the user affect the debug-layers enabled.

src/vulkan/vk-device.cpp
16-Bit-Dog: For now I will assume this is true until further review:
> Since we are changing this to a warning, do we still return a SLANG_FAIL or do we continue on skipping the unsupported feature? I would assume we don't return a SLANG_FAIL since a warning implies that everything can still work.

src/d3d12/d3d12-device.cpp
```diff
@@ -185,6 +185,82 @@ static void __stdcall raytracingValidationMessageCallback(
...
+        {
+            ComPtr<ID3D12Debug1> debug1;
+            if (SLANG_SUCCEEDED(m_dxDebug->QueryInterface(debug1.writeRef())))
+            {
+                debug1->SetEnableGPUBasedValidation(willGPUAssistedValidationBeEnabled);
+                activeDebugLayers.GPUAssistedValidation = willGPUAssistedValidationBeEnabled;
+            }
+            else
+            {
+                printWarning("GPU Based Validation is unavailable.\n");
```
skallweitNV:
```suggestion
                printWarning("GPU based validation requested but not available.\n");
```

src/d3d12/d3d12-device.cpp
```diff
@@ -877,6 +931,10 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
                 ));
             }
         }
+        else
+        {
+            printWarning("Tried to enable raytracing validation but unsupported\n");
```
skallweitNV:
```suggestion
            printWarning("Raytracing validation requested but not available.\n");
```

src/vulkan/vk-device.cpp
```diff
@@ -867,15 +887,22 @@ Result DeviceImpl::initVulkanInstanceAndDevice(
...
-            SIMPLE_EXTENSION_FEATURE(
-                extendedFeatures.rayTracingValidationFeatures,
-                rayTracingValidation,
-                VK_NV_RAY_TRACING_VALIDATION_EXTENSION_NAME,
-                { availableFeatures.push_back(Feature::RayTracingValidation); }
-            );
+            // If unsupported, fail
+            if (!extendedFeatures.rayTracingValidationFeatures.rayTracingValidation)
+            {
+                printWarning("Tried to enable raytracing validation but unsupported\n");
```
skallweitNV:
```suggestion
                printWarning("Raytracing validation requested but not available.\n");
```

src/rhi.cpp
```diff
@@ -133,33 +133,26 @@ inline const FormatInfo& _getFormatInfo(Format format)
...
-    virtual bool isDeviceTypeSupported(DeviceType type) override;
-    virtual const char* getFeatureName(Feature feature) override;
-    virtual const char* getCapabilityName(Capability capability) override;
+void RHI::decrementLiveDeviceCount() { m_liveDeviceCount--; }
+    
+Result RHI::setDebugLayerOptions(DebugLayerOptions options)
+{
+    if (m_liveDeviceCount != 0)
+        return SLANG_FAIL;
+    this->m_debugLayerOptions = options;
```
skallweitNV:
```suggestion
    m_debugLayerOptions = options;
```

include/slang-rhi.h
```diff
@@ -3031,12 +3031,16 @@ struct DeviceDesc
     /// The format matches the OPTIX_VERSION macro, e.g. 90000 for version 9.0.0.
     uint32_t requiredOptixVersion = 0;
 
-    /// Enable RHI validation layer.
+    /// Enable validation for Slang-RHI API (All).
+    /// This does not enable down-stream API Debug Layers
+    /// (ex: Vulkan Validation Layers).
```
```suggestion
    /// Enable validation for the Slang-RHI API. 
    /// This does not enable downstream API debug layers (D3D12/Vulkan validation layers).
```

include/slang-rhi.h
```diff
@@ -3031,12 +3031,16 @@ struct DeviceDesc
...
     bool enableValidation = false;
-    /// Enable backend API raytracing validation layer (D3D12, Vulkan and CUDA).
+    /// Enable downstream API raytracing validation (CUDA | D3D12 | Vulkan).
```
skallweitNV, line 3106: Maybe in the future we should move this to the `DebugLayerOptions`, but it's fine to keep it here for now.

include/slang-rhi.h
```diff
@@ -3537,9 +3541,41 @@ class IPersistentCache : public ISlangUnknown
...
+
 class IRHI
 {
 public:
+    /// Check is downstream debug layer is enabled
+    inline bool isDebugLayersEnabled() { return getDebugLayerOptions().isDebugLayersEnabled(); }
+
+    /// Change downstream debug layer options.
+    /// All devices must be released.
+    /// Fails if not all devices are released.
```
```suggestion
    /// Fails if not all devices are released.
```

tests/main.cpp
skallweitNV: I would prefer keeping the configuration of the debug layers here instead of defining it through `kDefaultDebugLayerOptions` which hides this fairly deep into the implementation.

tests/testing.cpp
```diff
@@ -556,6 +508,30 @@ const char* deviceTypeToString(DeviceType deviceType)
     }
 }
 
+void releaseCachedDevices()
+{
+    gCachedDevices.clear();
+    getRHI()->reportLiveObjects();
+}
+
+void TryToChangeCurrentDebugLayerStateAndOptions(DebugLayerOptions targetDebugLayerOptions)
```
skallweitNV: Coding conventions ..

tests/testing.h
```diff
@@ -428,43 +438,85 @@ enum GpuTestFlags
...
+private:
+    DebugLayerOptions m_options = {};
+};
+
 struct GpuTestInfo
 {
     GpuTestFunc func;
     DeviceType deviceType;
     GpuTestFlags flags;
+    DebugLayerOptions debugLayerOptions;
```
skallweitNV, line 469: Maybe this should be `std::optional<DebugLayerOptions>` instead.

This would allow us to only set it if we want to override the settings.
When we need to switch debug layer settings for a test, we can store the current settings, switch to the requested ones for the test, run the test and reset back to the previous state after finishing the test.

What do you think?

tests/testing.h
```diff
@@ -328,7 +329,16 @@ struct DeviceExtraOptions
...
+    bool enableValidation = false;    
+    bool enableRayTracingValidation = false;
     bool enableAftermath = false;
+    
+    AftermathFlags aftermathFlags = AftermathFlags::Default;
+
+    std::set<DebugMessageType> debugMessageTypesToEmit =
+        {DebugMessageType::Info, DebugMessageType::Warning, DebugMessageType::Error};
+    std::set<DebugMessageSource> debugMessageSourcesToEmitFrom =
+        {DebugMessageSource::Slang, DebugMessageSource::Layer, DebugMessageSource::Driver };
```
skallweitNV: Do we really need this? It looks like it's not used anywhere. I would remove it.

include/slang-rhi.h
16-Bit-Dog: In my opinion, unless `enableValidation` is changed, there really is no reason to merge it with `DebugLayerOptions`.

My justification is as follows:
1. Since `enableValidation` is only validation for the slang-rhi api, it makes sense to allow per-device validation to be enabled for this option (`DebugLayerOptions` is either all devices or none of them as a limitation of downstream-api's)
2. `enableRayTracingValidation` is per device since a user may want a D3D12 device without raytracing-validation but Vulkan with raytracing-validation. `enableValidation` is much like `enableRayTracingValidation` in this respect since both are per-device 'not a debug layer but close enough'. Due to this, I would suggest that these per-device-debug options should be grouped up with other per-device debugging options instead of moving everything to `DebugLayerOptions` (`enableAftermath`, `enableValidation`, `enableRayTracingValidation`).
3. `enableValidation` would need a new more explicit name to put it in `DebugLayerOptions`. Something like `RHIValidation`. This is less of an issue and more of a remark on if we do end up merging everything.

tests/main.cpp
16-Bit-Dog: I agree that `kDefaultDebugLayerOptions` hides logic too much since its embedded in that messy header.

Despite this I think setting the 'defaults' in `main.cpp` is a problem since `main` runs after registering all GPU tests (`registerGpuTest`), so if we set defaults this won't reflect in the parameters of the GPU test trampoline function (this was partly a huge reason logic was moved out from that `main` function).

Instead what can be done, is this logic can be put in `registerGpuTest` so that all logic setting the `DebugLayerOptions` for a GPU test are kept together.

tests/testing.h
16-Bit-Dog: I think this is a good approach. It works pretty well with the suggested change [here](https://github.com/shader-slang/slang-rhi/pull/622#discussion_r2721372375).

The only issue is the method that we will use it:
> When we need to switch debug layer settings for a test, we can store the current settings, switch to the requested ones for the test, run the test and reset back to the previous state after finishing the test.

What we should probably do is provide default `DebugLayerOptions` directly in the function `registerGpuTest`, then we either pick the 'default' or what the programmer provided to the GPU test case.

This is probably best since the test infrastructure in order does the following: (1) registers all tests; (2) runs `main`; (3) runs all tests. Therefore, by keeping everything in `registerGpuTest` things are kept simple and all in one place.

16-Bit-Dog: The suggested approach has the issue that it requires us to move all state management to `gpuTestTrampoline`, after a test is already registered (deciding parameters downstream rather than before a test).

include/slang-rhi.h
skallweitNV: No, `enableValidation` and `enableAftermath` should definitely remain in `DeviceDesc`. I think `enableRayTracingValidation` could be considered to be moved to the `DebugLayerOptions` though. But you have a valid point to also keep that one in `DeviceDesc`. All good.

tests/testing.cpp
```diff
@@ -179,35 +156,11 @@ class DebugCallback : public IDebugCallback
...
-            break;
-        }
+        if (shouldIgnoreMessage(type, source, message))
+            return;
+
+        doctest::String msg = "[" + doctest::String(enumToString(type)) + "] ";
+        msg += "[" + doctest::String(enumToString(source)) + "] ";
         msg += message;
 
         auto output = [](const doctest::String& str)
```
skallweitNV: Minor: We could just fold this into the `if (type == DebugMessageType::Info || type == DebugMessageType::Warning)` block below now.

tests/testing.h
```diff
@@ -12,6 +12,8 @@
 #include <string_view>
 #include <vector>
 #include <cstring>
+#include <set>
```
skallweitNV: Is this include obsolete now?

tests/testing.h
skallweitNV: I still think the right approach is to switch the debug layer options in `gpuTestTrampoline`. This allows setting the "default debug layer options" used by all tests before running tests at runtime, and not rely on knowing them at compile time. The `registerGpuTest` should just forward the `std::optional<DebugLayerOptions>` straight to `GpuTestInfo` as an optional. Then in the trampoline we can quickly check if the test requests to override the settings (rare) and in that case try to switch debug layer options.

I expect tests to request specific debug layer options to be a fairly rare thing. If that's not the case and we really want the debug layer options to be part of a test spec, this approach would be worse. But my assumption is that's not the case.

tests/testing.cpp
```diff
@@ -1047,10 +1011,39 @@ class GpuTestAllocator
...
+#else
+    static rhi::DebugLayerOptions kDefaultDebugLayerOptions = rhi::testing::DebugLayerOptionsBuilder();
+#endif
+
+    DebugLayerOptions temporaryNewDebugLayerOptions;
+    if (maybeDebugLayerOptions.has_value())
+        temporaryNewDebugLayerOptions = *maybeDebugLayerOptions;
+    else
+        temporaryNewDebugLayerOptions = kDefaultDebugLayerOptions;
+
```
skallweitNV: All of this can be removed.

tests/testing.cpp
```diff
@@ -1047,10 +1011,39 @@ class GpuTestAllocator
 // Because doctest doesn't support any user data in the test case definition and we don't want to alter the
 // doctest implementation, we store the GpuTestInfo structure in front of the unique test name used for each
 // test instance.
-int registerGpuTest(const char* name, GpuTestFunc func, GpuTestFlags flags, const char* file, int line)
+int registerGpuTest(
+    const char* name,
+    GpuTestFunc func,
+    GpuTestFlags flags,
+    std::optional<DebugLayerOptions> maybeDebugLayerOptions,
```
skallweitNV: rename to `debugLayerOptions` or `overrideDebugLayerOptions`

tests/testing.cpp
```diff
@@ -994,6 +956,8 @@ static void gpuTestTrampoline()
         SKIP("device not selected");
     }
 
+    tryToChangeCurrentDebugLayerStateAndOptions(info->debugLayerOptions);
```
skallweitNV: Here we just check if we have custom debug layer options in which case we do:
- remove cached devices
- cache current settings
- set temporary settings (report error if that fails)
- run test
- remove cached devices
- set previous settings

tests/main.cpp
```diff
@@ -33,7 +33,9 @@ int main(int argc, const char** argv)
     rhi::testing::cleanupTestTempDirectories();
 
 #if SLANG_RHI_DEBUG
-    rhi::getRHI()->enableDebugLayers();
+    rhi::DebugLayerOptions defaultDebugLayerOptions{};
+    defaultDebugLayerOptions.coreValidation = true;
+    rhi::getRHI()->setDebugLayerOptions(defaultDebugLayerOptions);
```
skallweitNV:
```suggestion
    rhi::DebugLayerOptions debugLayerOptions{};
    debugLayerOptions.coreValidation = true;
    rhi::getRHI()->setDebugLayerOptions(debugLayerOptions);
```

tests/testing.cpp
```diff
@@ -598,8 +556,14 @@ ComPtr<IDevice> createTestingDevice(
         deviceDesc.enableCompilationReports = extraOptions->enableCompilationReports;
         deviceDesc.existingDeviceHandles = extraOptions->existingDeviceHandles;
         deviceDesc.enableAftermath = extraOptions->enableAftermath;
+        deviceDesc.enableRayTracingValidation = extraOptions->enableRayTracingValidation;
+        deviceDesc.enableValidation = extraOptions->enableValidation;
+        deviceDesc.aftermathFlags = extraOptions->aftermathFlags;
     }
 
+    if (rhi::getRHI()->isDebugLayersEnabled())
+        deviceDesc.debugCallback = &sDebugCallback;
```
skallweitNV: Previously we always set the callback in DEBUG mode .. I think we should keep that, we want validation messages even when debug layers are disabled.

tests/testing.cpp
```diff
@@ -556,6 +493,30 @@ const char* deviceTypeToString(DeviceType deviceType)
...
+    getRHI()->reportLiveObjects();
+}
+
+void tryToChangeCurrentDebugLayerStateAndOptions(DebugLayerOptions targetDebugLayerOptions)
+{
+    // By default debug layer is 'off'
+    static DebugLayerOptions currentDebugLayerOptions = {};
+
+    // Nothing to change
+    if (currentDebugLayerOptions == targetDebugLayerOptions)
```
skallweitNV: Looks wrong to compare to the statically initialized `currentDebugLayerOptions` which not necessarily match with the current set options (from main.cpp).

I think we should move the comparison to current settings to `gpuTestTrampoline` and only call this function when we actually want to change the options. Then the local static and this test are obsolete.

tests/testing.cpp
```diff
@@ -851,6 +808,7 @@ DeviceAvailabilityResult checkDeviceTypeAvailable(DeviceType deviceType)
     desc.deviceType = deviceType;
     desc.adapter = getSelectedDeviceAdapter(deviceType);
 #if SLANG_RHI_DEBUG
+    desc.enableValidation = true;
```
skallweitNV: Any reason why to enable validation here now?

include/slang-rhi.h
```diff
@@ -3561,10 +3590,6 @@ class IRHI
         return AdapterList(blob);
     }
 
-    /// Enable debug layers, if available.
-    /// If this is called, it must be called before creating any devices.
-    virtual SLANG_NO_THROW void SLANG_MCALL enableDebugLayers() = 0;
```
skallweitNV, line 3642: Actually one more thing. We should probably keep this API and mark it as deprecated (in a comment will do). We have a bunch of downstream projects relying on this API, so not breaking them seems like a good thing to do.

tests/testing.cpp
16-Bit-Dog: My thinking went like this: if we are in a debug build, `checkDeviceTypeAvailable` should only pass if enabling validation doesn't break the Slang-RHI DeviceType.

I understand why this isn't very correct though, I will remove this change.

include/slang-rhi.h
16-Bit-Dog: I suppose since the goal is to keep downstream projects working then the definition (in `RHI`) will not do any `m_liveDeviceCount` validation either.

skallweitNV (dismissed):
LGTM. Thanks a lot for all the work on this!

skallweitNV:
Well, there seem to be some remaining issues.

16-Bit-Dog:
I tried running the program on my older rtx3050 machine and the test indeed started to fail, I will see what went wrong (SDK or other).

16-Bit-Dog:
What was happening:
1. We set the new debug settings (GPU-AV debug layer) ==> fails due to old vulkan sdk(?)
2. `REQUIRE_CALL` triggers an assert ==> we never switch the new debug settings back to the default
3. Every test now fails since they all use GPU-AV

I pushed some changes code to fix this bug (we should only see `acceleration-structure-creation-with-validation.vulkan` failing now).

This behaviour was reproduced on Vulkan SDK 1.4.313.2. Vulkan SDK 1.4.321.0 does not have this problem.

skallweitNV:
Oh, that's a good catch! It looks like we have a bug in all `getQueue` implementations. They call `m_queue->establishStrongReferenceToDevice();` which is wrong, because the strong/weak reference handling is actually supposed to be handled by `RefObject` using the "internal/external" mechanism. So this looks to be leftover code from earlier days. I'll put up a separate PR to fix this. Then we can see if this resolves the issues we see here.

skallweitNV:
On second thought, this call to `establishStrongReferenceToDevice` shouldn't really matter. It's not needed but shouldn't lead to a ref-cycle because that cycle is broken up in `CommandQueue::makeInternal` when the ref-count of the queue goes to `1` (which happens when the client releases the com ptr).

skallweitNV:
I've just run this branch on my local machine with `SLANG_RHI_ENABLE_REF_OBJECT_TRACKING` and there are zero issues with lifetimes, everything is cleaned up correctly AFAICT. Not sure what you've seen w.r.t. queues keeping the device alive.

What I see from the latest run on CI is that the remaining issue is that the test requiring Vulkan GPU side validation just fails on the CI machines. So I guess the right move now would be to skip the test if validation cannot be enabled. Is that something we can easily accomplish?

tests/testing.cpp
```diff
@@ -556,6 +493,21 @@ const char* deviceTypeToString(DeviceType deviceType)
     }
 }
 
+void releaseCachedDevices()
+{
+    gCachedDevices.clear();
+    getRHI()->reportLiveObjects();
+}
+
+void tryToChangeCurrentDebugLayerStateAndOptions(DebugLayerOptions targetDebugLayerOptions)
```
skallweitNV: My suggestion is to make this function return `Result` and only warn if the debug layer options cannot be set.
Then in the test trampoline we still continue running tests even if the debug layer options couldn't be set. This would allow tests to check if the required debug layer features are actually enabled and skip them if not available. This will make testing more robust and automatically skip tests if they can't run because they depend on a certain debug layer feature to be available.

16-Bit-Dog: I think skipping a test with a warning when `tryToChangeCurrentDebugLayerStateAndOptions` fails makes sense. The downside of-course is that the test framework no longer validates correctness regarding `tryToChangeCurrentDebugLayerStateAndOptions`, although a separate test can be made for this.

Similarly, skipping a test conditionally on if a device fails to create with a particular set of custom `DebugLayerOptions` is also fine but with a caveat. To ensure `createDevice` works with combinations of `DebugLayerOptions` (since regular tests no longer verify functionality), we will need to add a new test that ensures all combinations of `DebugLayerOptions` still works with `createDevice`. 

If there are issues with this plan, please share.

skallweitNV: Wait, I think I was wrong here. `tryToChangeCurrentDebugLayerStateAndOptions` should never fail, because we should always be able to destroy all cached devices and succeed in setting the debug layer options. If this doesn't work, something is wrong with the testing framework. So I guess that function should be renamed and use `REQUIRE` to make sure setting the debug layer options succeeds.

The question is how we handle debug layer features not being available on device creation. This can either be a hard error, which it seems to be in the case of the acceleration structure test leading to `createDevice` to fail, or it could just be best-effort with warnings if certain features cannot be enabled, which I think we do for other debug layer features. I would generally prefer the latter. However, if a test or user application requires certain features to be available, we need a way to figure out if they are. If we just warn and continue, the client doesn't know. One way to do this would be if we had a function on `Device` to query the enabled debug layer features. This would generally be the same set of features that was requested, but may be fewer. With this, tests could check and skip if they absolutely need a feature to work.

This all feels a bit convoluted though, so not 100% on any of this ...

skallweitNV: The acceleration structure test works without the GPU validation feature, right? So I think a warning instead of a failure in the device creation code is all we really need.

16-Bit-Dog:
> I've just run this branch on my local machine with SLANG_RHI_ENABLE_REF_OBJECT_TRACKING and there are zero issues with lifetimes, everything is cleaned up correctly AFAICT. Not sure what you've seen w.r.t. queues keeping the device alive.

When I run `-tc=*heap-pointer-stress-test*` I see the following output with `#define SLANG_RHI_ENABLE_REF_OBJECT_TRACKING 1`:
```
Live object: 0x1FC8C0EA6C0 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::optix::v9_0::ContextImpl"
Live object: 0x1FC8C16ED10 referenceCount=1 internalReferenceCount=0 type="class rhi::LiveDeviceTracker"
Live object: 0x1FC8D1A2800 referenceCount=1 internalReferenceCount=0 type="class rhi::LiveDeviceTracker"
Live object: 0x1FC8D55C398 referenceCount=6 internalReferenceCount=0 type="class rhi::vk::ComputePipelineImpl"
Live object: 0x1FC8D596A18 referenceCount=6 internalReferenceCount=0 type="class rhi::vk::ComputePipelineImpl"
Live object: 0x1FC8D5B98A8 referenceCount=14 internalReferenceCount=0 type="class rhi::cuda::DeviceImpl"
Live object: 0x1FC8D5B9BC8 referenceCount=0 internalReferenceCount=0 type="class rhi::ShaderCache"
Live object: 0x1FC8D5B9C68 referenceCount=0 internalReferenceCount=0 type="class rhi::StagingHeap"
Live object: 0x1FC8D5B9D48 referenceCount=0 internalReferenceCount=0 type="class rhi::StagingHeap"
Live object: 0x1FC8DD7EC98 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::ShaderProgramImpl"
Live object: 0x1FC8DD7EDF8 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::ShaderProgramImpl"
Live object: 0x1FC93FFF4E0 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::ShaderObjectLayoutImpl"
Live object: 0x1FC94004C30 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::ShaderObjectLayoutImpl"
Live object: 0x1FC9400B7C8 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400B808 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400B938 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400B978 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400BAA8 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400BAE8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400BC18 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400BC58 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400BD88 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400BDC8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400BEF8 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400BF38 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400C068 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400C0A8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400C348 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400C388 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400C628 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400C668 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC9400CD58 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::CommandBufferImpl"
Live object: 0x1FC9400CD98 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FC943E85C8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::ShaderProgramImpl"
Live object: 0x1FC943E99E8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::ShaderProgramImpl"
Live object: 0x1FC95DEF048 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::HeapImpl"
Live object: 0x1FC95DEF2E8 referenceCount=1 internalReferenceCount=0 type="class rhi::cuda::HeapImpl"
Live object: 0x1FC960F0E80 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BindlessDescriptorSet"
Live object: 0x1FC960F3948 referenceCount=1 internalReferenceCount=1 type="class rhi::vk::CommandQueueImpl"
Live object: 0x1FC96115A70 referenceCount=2 internalReferenceCount=0 type="class rhi::cuda::RootShaderObjectLayoutImpl"
Live object: 0x1FC96118170 referenceCount=2 internalReferenceCount=0 type="class rhi::cuda::RootShaderObjectLayoutImpl"
Live object: 0x1FC97E7CCF8 referenceCount=1 internalReferenceCount=1 type="class rhi::cuda::CommandQueueImpl"
Live object: 0x1FC97E7D248 referenceCount=10 internalReferenceCount=0 type="class rhi::cuda::ComputePipelineImpl"
Live object: 0x1FC97E7D578 referenceCount=10 internalReferenceCount=0 type="class rhi::cuda::ComputePipelineImpl"
Live object: 0x1FEB6ADEC88 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE0888 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE0DC8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE0F88 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE1148 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE2108 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB6AE2488 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::BufferImpl"
Live object: 0x1FEB7377048 referenceCount=10 internalReferenceCount=0 type="class rhi::vk::DeviceImpl"
Live object: 0x1FEB7377368 referenceCount=0 internalReferenceCount=0 type="class rhi::ShaderCache"
Live object: 0x1FEB7377408 referenceCount=0 internalReferenceCount=0 type="class rhi::StagingHeap"
Live object: 0x1FEB73774E8 referenceCount=0 internalReferenceCount=0 type="class rhi::StagingHeap"
Live object: 0x1FEB7CF2178 referenceCount=1 internalReferenceCount=1 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF21B8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF22D8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF2318 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF26F8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF2738 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF2858 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF2898 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF2C78 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF2CB8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF3778 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF37B8 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7CF38D8 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::CommandBufferImpl"
Live object: 0x1FEB7CF3918 referenceCount=0 internalReferenceCount=0 type="class rhi::CommandList"
Live object: 0x1FEB7FBE0A0 referenceCount=2 internalReferenceCount=0 type="class rhi::vk::RootShaderObjectLayoutImpl"
Live object: 0x1FEB7FBF7C0 referenceCount=2 internalReferenceCount=0 type="class rhi::vk::RootShaderObjectLayoutImpl"
Live object: 0x1FEBB6DEE80 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::EntryPointLayout"
Live object: 0x1FEBB6E0880 referenceCount=1 internalReferenceCount=0 type="class rhi::vk::EntryPointLayout"
73 leaked objects detected!
```

tests/testing.cpp
16-Bit-Dog: > If this doesn't work, something is wrong with the testing framework. So I guess that function should be renamed and use REQUIRE to make sure setting the debug layer options succeeds.

I agree with this observation

> The question is how we handle debug layer features not being available on device creation. This can either be a hard error, which it seems to be in the case of the acceleration structure test leading to createDevice to fail, or it could just be best-effort with warnings if certain features cannot be enabled, which I think we do for other debug layer features. I would generally prefer the latter. 

I think either approach works, as long as we are consistent in approach and users have a programmatic way to determine if a debug layer is enabled, either: (1) always error if failure; (2) always warn and allow querying for if there is support for a specific `DebugLayerOptions` configuration.

> One way to do this would be if we had a function on Device to query the enabled debug layer features. This would generally be the same set of features that was requested, but may be fewer. With this, tests could check and skip if they absolutely need a feature to work.

`DebugLayerOptions` compatibility can be queried in all cases that we use them currently (probably best to do this by requiring 0 live devices before querying):
1. D3D12 (the entire `setupDebugLayer` function can be made to return FAIL/OK, then options are disabled)
2. Vulkan (requires creating a dummy vkInstance)
3. DXGIFactory (check if Factory is created with the Debug flag)
4. D3D11 (check if device is created)

> This all feels a bit convoluted though, so not 100% on any of this ...

I agree that allowing a user to query is a lot of effort and can likely get quite messy. The simplest solution is to just error if something fails right now since that cannot really 'go wrong'.

If we choose to go the 'error' approach, what we can do in the testing framework is:
```
        if (testRequestsDifferentDebugLayerOptions)
        {
            REQUIRE_CALL(tryToChangeCurrentDebugLayerStateAndOptions(info->debugLayerOptions));
            cachedPreviousDebugLayer = true;

            // Ensure our device supports the suggested custom `DebugLayerOptions`, else
            // Check if we lack support for the custom configuration
            if (checkDeviceTypeAvailable(deviceType).available == false)
                SKIP("Requested `DebugLayersOptions` are unsupported, skipping test"); // we can also revert custom `DebugLayerOptions` set and run the test without the options.
        }
```

16-Bit-Dog: > The acceleration structure test works without the GPU validation feature, right? So I think a warning instead of a failure in the device creation code is all we really need.

The test will pass if we warn instead of error, but if we are going to warn we could instead just `SKIP` the test if a custom `DebugLayerOptions` fails to create a device.

I would imagine having the test say 'pass' without the test actually passing can get very confusing if regressions show up (since the regressions would not be present in the test results).

skallweitNV: | I agree that allowing a user to query is a lot of effort and can likely get quite messy. The simplest solution is to just error if something fails right now since that cannot really 'go wrong'.

While error on device creation is the simplest solution, it's also not great for applications. I strongly prefer the warning in this setting because everything still works, we just miss debug layers. One solution to this problem is if we add a `required` field in the `DebugLayerOptions`. The `required` flag would be `false` by default, meaning that device creation will just warn if a debug feature cannot be enabled. If `true, device creation fails if the debug layer options cannot be applied. This way, tests can require the settings if they are specified, but user applications can still work with best-effort enabling of debug layers. What do you think about this?

16-Bit-Dog: I like this approach a lot.

When I next have the time I'll make the changes need.

skallweitNV: Great! I think this will conclude this journey :D thanks a lot for pulling through!!

PR #623, opened 2026-01-15
Vulkan: release upload buffer after init copy

csyonghe:
Vulkan init‑upload memory retention — per‑buffer staging kept alive
In Vulkan createBuffer(initData) the backend created a per‑buffer upload/staging buffer (m_uploadBuffer) and kept it as a member, so host memory scaled with every buffer ever initialized. The fix uses a local upload buffer and releases it after the copy completes (with flushAndWait() to match D3D’s semantics), eliminating the persistent host‑memory footprint.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal buffer initialization workflow for device-local memory allocation, reducing code complexity.


ccummingsNV (approved):

ccummingsNV (changes requested):
Originally approved, but looking at it, this means we end up blocking the device for an upload doesn't it?

If so, this would be better implemented using the proper upload mechanism that slang-rhi already has (using the command queue - see uploadBufferData)

csyonghe:
@ccummingsNV Addressed this in 5c138a1 after merging the latest official/main into the PR branch. The Vulkan init path now uses the existing ICommandEncoder::uploadBufferData() / command queue upload mechanism instead of a per-buffer upload buffer plus flushAndWait(), so it no longer blocks the device for each init upload and no longer keeps a dedicated staging buffer attached to the buffer object.

I also rebuilt slang-rhi-tests locally and reran the Vulkan-focused coverage for cmd-upload-buffer* and buffer-from-handle*; both passed. If you have a moment, I’d appreciate another look.

skallweitNV:
Closing this as this will be solved by https://github.com/shader-slang/slang-rhi/pull/829

PR #626, opened 2026-01-21
PyTorch-style caching allocator for the CUDA backend with proper multi-stream support.

szihs:
Summary

  This PR implements a PyTorch-style caching allocator for the CUDA backend with proper multi-stream support. The key improvements are:

  - Same-stream immediate memory reuse - Memory freed on the same stream can be immediately reused without waiting for GPU completion (CUDA FIFO ordering guarantees safety)
  - Lazy events optimization - Single-stream workloads skip event creation entirely, using cuStreamQuery() instead
  - Multi-stream tracking - Cross-stream usage is tracked via events, with proper synchronization before reuse
  - Explicit stream passing - Removed fragile device-level m_currentStream global state, replaced with explicit stream parameters through the allocation call chain

  - Added kNoStream sentinel value to distinguish "no stream context" from nullptr (the default CUDA stream)
  - Added stream field to HeapAllocDesc for explicit stream context passing

  Heap Infrastructure (src/heap.h, src/heap.cpp)

  - Added stream field to PageDesc
  - Added Page::notifyUse(void* stream) virtual method for multi-stream tracking
  - Stream passed from HeapAllocDesc → PageDesc → Page::notifyUse()

  CUDA Heap Implementation (src/cuda/cuda-heap.h, src/cuda/cuda-heap.cpp)

  - PageImpl stream tracking: Each page tracks its allocation stream (m_stream) and cross-stream events (m_pendingEvents)
  - Same-stream immediate reuse: HeapImpl::free() immediately retires allocations when same-stream with no cross-stream events
  - Cross-stream deferred free: Allocations used across streams are deferred until events complete
  - PageCache: Two-pass algorithm preferring same-stream pages for reuse

  CUDA Command Queue (src/cuda/cuda-command.cpp)

  - Explicit stream passing: Removed device->m_currentStream global state
  - Lazy events optimization: Single-stream workloads use cuStreamQuery() instead of events
  - Event cleanup in destructor: Fixed event handle leak with cuEventSynchronize() before cuEventDestroy()
  - Context sync on destruction: Added cuCtxSynchronize() to prevent use-after-free on device destruction

  Constant Buffer Pool (src/cuda/cuda-constant-buffer-pool.h, src/cuda/cuda-constant-buffer-pool.cpp)

  - Added stream parameter to allocate() and createPage() methods
  - Stream context flows from command encoder to heap allocation

  Shader Object Binding (src/cuda/cuda-shader-object.h)

  - Added m_stream field to BindingDataBuilder
  - Set from m_queue->m_stream during binding

### DIFF
--- a/src/cuda/cuda-command.cpp
+++ b/src/cuda/cuda-command.cpp
@@ -494,8 +494,30 @@ void CommandExecutor::cmdDispatchCompute(const commands::DispatchCompute& cmd)
 
 void CommandExecutor::cmdDispatchComputeIndirect(const commands::DispatchComputeIndirect& cmd)
 {
-    SLANG_UNUSED(cmd);
-    NOT_SUPPORTED(S_ComputePassEncoder_dispatchComputeIndirect);
+    if (!m_computeStateValid)
+        return;
+
+    // Emulate indirect dispatch by reading the dispatch arguments from the buffer on the host.
+    // This requires a GPU->CPU synchronization and stalls the GPU pipeline, but CUDA does not provide
+    // a way to perform indirect dispatches natively.
+    // In the future, we might consider using CUDA dynamic parallelism to launch a helper kernel
+    // that performs the actual dispatch, but that would require more setup and complexity.
+    BufferImpl* argBuffer = checked_cast<BufferImpl*>(cmd.argBuffer.buffer);
+    CUdeviceptr srcPtr = (CUdeviceptr)argBuffer->m_cudaMemory + cmd.argBuffer.offset;
+
+    IndirectDispatchArguments args;
+    SLANG_CUDA_ASSERT_ON_FAIL(cuMemcpyDtoH(&args, srcPtr, sizeof(IndirectDispatchArguments)));
+
+    // Skip dispatch if any dimension is zero (CUDA doesn't allow zero-sized dispatches).
+    if (args.threadGroupCountX == 0 || args.threadGroupCountY == 0 || args.threadGroupCountZ == 0)
+        return;
+
+    // Dispatch with the retrieved arguments.
+    commands::DispatchCompute dispatchCmd;
+    dispatchCmd.x = args.threadGroupCountX;
+    dispatchCmd.y = args.threadGroupCountY;
+    dispatchCmd.z = args.threadGroupCountZ;
+    cmdDispatchCompute(dispatchCmd);
 }
 
 void CommandExecutor::cmdBeginRayTracingPass(const commands::BeginRayTracingPass& cmd)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,13 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     set(SLANG_RHI_HAS_METAL ON)
     set(SLANG_RHI_HAS_CUDA OFF)
     set(SLANG_RHI_HAS_WGPU ON)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Android")
+    set(SLANG_RHI_HAS_D3D11 OFF)
+    set(SLANG_RHI_HAS_D3D12 OFF)
+    set(SLANG_RHI_HAS_VULKAN ON)
+    set(SLANG_RHI_HAS_METAL OFF)
+    set(SLANG_RHI_HAS_CUDA OFF)
+    set(SLANG_RHI_HAS_WGPU OFF)
 endif()
 
 set(SLANG_RHI_HAS_AGILITY_SDK OFF)
@@ -845,6 +852,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED)
         tests/test-cmd-upload-buffer.cpp
         tests/test-cmd-upload-texture.cpp
         tests/test-combined-texture-sampler.cpp
+        tests/test-compute-indirect.cpp
         tests/test-compute-smoke.cpp
         tests/test-compute-trivial.cpp
         tests/test-cooperative-vector.cpp
--- a/src/shader-object.cpp
+++ b/src/shader-object.cpp
@@ -61,10 +61,34 @@ Result ShaderObject::getEntryPoint(uint32_t index, IShaderObject** outEntryPoint
 
 Result ShaderObject::setData(const ShaderOffset& offset, const void* data, Size size)
 {
+    // TODO: Ideally we would use the following code here, but slang-test relies
+    // on out-of-bounds `size` to be clamped to the actual buffer size.
+#if 0
     void* dest;
     SLANG_RETURN_ON_FAIL(reserveData(offset, size, &dest));
     ::memcpy(dest, data, size);
     return SLANG_OK;
+#else
+    SLANG_RETURN_ON_FAIL(checkFinalized());
+
+    size_t dataOffset = offset.uniformOffset;
+    size_t dataSize = size;
+
+    uint8_t* dest = m_data.data();
+    size_t availableSize = m_data.size();
+
+    // This should be an error, but slang-test relies on this behavior.
+    if ((dataOffset + dataSize) > availableSize)
+    {
+        dataSize = availableSize - dataOffset;
+    }
+
+    ::memcpy(dest + dataOffset, data, dataSize);
+
+    incrementVersion();
+
+    return SLANG_OK;
+#endif
 }
 
 Result ShaderObject::reserveData(const ShaderOffset& offset, Size size, void** outData)
@@ -77,13 +101,9 @@ Result ShaderObject::reserveData(const ShaderOffset& offset, Size size, void** o
     uint8_t* dest = m_data.data();
     size_t availableSize = m_data.size();
 
-    // TODO(shaderobject): We really should bounds-check access rather than silently ignoring sets
-    // that are too large, but we have several test cases that set more data than
-    // an object actually stores on several targets...
-    //
     if ((dataOffset + dataSize) > availableSize)
     {
-        dataSize = availableSize - dataOffset;
+        return SLANG_E_INVALID_ARG;
     }
 
     *outData = dest + dataOffset;
--- a/tests/test-compute-indirect.cpp
+++ b/tests/test-compute-indirect.cpp
@@ -1,0 +1,288 @@
+#include "testing.h"
+
+using namespace rhi;
+using namespace rhi::testing;
+
+// Test dispatchComputeIndirect with a simple compute shader.
+// The test sets up an indirect argument buffer with dispatch dimensions written by the GPU,
+// then verifies the compute shader ran with the correct number of threads.
+GPU_TEST_CASE("compute-indirect", D3D12 | Vulkan | CUDA)
+{
+    ComPtr<IShaderProgram> program;
+    REQUIRE_CALL(loadProgram(device, "test-compute-indirect", "computeMain", program.writeRef()));
+
+    ComPtr<IShaderProgram> writeArgsProgram;
+    REQUIRE_CALL(loadProgram(device, "test-compute-indirect", "writeDispatchArgs", writeArgsProgram.writeRef()));
+
+    ComputePipelineDesc pipelineDesc = {};
+    pipelineDesc.program = program.get();
+    ComPtr<IComputePipeline> pipeline;
+    REQUIRE_CALL(device->createComputePipeline(pipelineDesc, pipeline.writeRef()));
+
+    ComputePipelineDesc writeArgsPipelineDesc = {};
+    writeArgsPipelineDesc.program = writeArgsProgram.get();
+    ComPtr<IComputePipeline> writeArgsPipeline;
+    REQUIRE_CALL(device->createComputePipeline(writeArgsPipelineDesc, writeArgsPipeline.writeRef()));
+
+    // Create output buffer to count how many threads ran for each dispatch
+    const int dispatchCount = 4;
+    BufferDesc outputBufferDesc = {};
+    outputBufferDesc.size = dispatchCount * sizeof(uint32_t);
+    outputBufferDesc.format = Format::Undefined;
+    outputBufferDesc.elementSize = sizeof(uint32_t);
+    outputBufferDesc.usage = BufferUsage::ShaderResource | BufferUsage::UnorderedAccess | BufferUsage::CopyDestination |
+                             BufferUsage::CopySource;
+    outputBufferDesc.defaultState = ResourceState::UnorderedAccess;
+    outputBufferDesc.memoryType = MemoryType::DeviceLocal;
+
+    uint32_t initialData[dispatchCount] = {0, 0, 0, 0};
+    ComPtr<IBuffer> outputBuffer;
+    REQUIRE_CALL(device->createBuffer(outputBufferDesc, initialData, outputBuffer.writeRef()));
+
+    // Create indirect argument buffer (uninitialized - will be written by GPU)
+    // Each IndirectDispatchArguments contains (threadGroupCountX, threadGroupCountY, threadGroupCountZ)
+    // With numthreads(4, 4, 1), each group has 16 threads
+    BufferDesc indirectBufferDesc = {};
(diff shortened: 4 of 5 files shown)


src/cuda/cuda-command.cpp
```diff
@@ -664,20 +665,25 @@ CommandQueueImpl::~CommandQueueImpl()
...
 
-    // Block on all events completing
+    // Block on all events completing (multi-stream case)
     for (const auto& ev : m_submitEvents)
     {
         SLANG_CUDA_ASSERT_ON_FAIL(cuEventSynchronize(ev.event));
     }
 
-    // Retire finished command buffers, which should be all of them
+    // Always synchronize the CUDA context before destruction.
```
ccummingsNV, line 700: I think this is fine to do on **queue** destruction - it's a one off. @skallweitNV do you agree.

src/cuda/cuda-command.cpp
```diff
@@ -664,20 +665,25 @@ CommandQueueImpl::~CommandQueueImpl()
 {
     SLANG_CUDA_CTX_SCOPE(getDevice<DeviceImpl>());
 
-    // Block on all events completing
+    // Block on all events completing (multi-stream case)
     for (const auto& ev : m_submitEvents)
     {
         SLANG_CUDA_ASSERT_ON_FAIL(cuEventSynchronize(ev.event));
```
ccummingsNV, line 695: Should we be destroying events as well?

Is it possible in NSight to track whether we leak these?

src/cuda/cuda-command.cpp
```diff
@@ -736,24 +758,64 @@ void CommandQueueImpl::retireCommandBuffer(CommandBufferImpl* commandBuffer)
...
+    }
+    else if (!m_commandBuffersInFlight.empty())
+    {
+        // Single-stream lazy mode: use cuStreamQuery (non-blocking)
+        CUresult result = cuStreamQuery(m_stream);
+        if (result == CUDA_SUCCESS)
+        {
+            // Stream is idle - all submitted work is complete
+            m_lastFinishedID = m_lastSubmittedID;
+        }
```
ccummingsNV, line 820: So if I understand, the idea here is that rather than free command buffers aggressively, we let them 'build up', but automatically free them if the stream finishes completely or they hit some peak at which GC is needed?  So this is a quick way of saying 'if theres no work left, we can set last finished to last submitted without worrying about events'.

src/cuda/cuda-command.cpp
```diff
@@ -763,6 +825,13 @@ Result CommandQueueImpl::createCommandEncoder(ICommandEncoder** outEncoder)
 {
     SLANG_CUDA_CTX_SCOPE(getDevice<DeviceImpl>());
 
+    // Set current stream on device for allocations during command encoding.
+    // This enables proper multi-stream tracking: when a page allocated on stream A
+    // is used for allocations during encoding for stream B, we record the cross-stream usage.
+    // All heaps query the device for the current stream, so we only need to set it once.
+    DeviceImpl* device = getDevice<DeviceImpl>();
```
ccummingsNV: I'm not entirely sure about this, due to the lifetime of a command encoder. One can be created multiple times and their lifetime could overlap. 

This concept of the device having a current stream feels like the most fragile aspect of the design

src/cuda/cuda-heap.cpp
```diff
@@ -5,28 +5,316 @@
...
+            if (!page->processEventsAndCheckReuse())
+                continue;
+
+            // Found a reusable page from same stream - best case
+            m_cachedPages.erase(it);
+            return page;
+        }
+    }
+
+    // Second pass: accept any reusable page (may need synchronization later)
```
ccummingsNV, line 233: Can you explain the concept of a 're-usable' page.

src/cuda/cuda-command.cpp
szihs, line 695: Good catch, let me spin Nsight with a test to confirm leak and fix it.

src/cuda/cuda-heap.cpp
szihs: Reusable page is a GPU memory page that has been completely freed. With caching enabled it is NOT deallocated and recycled (moved) to page cache (m_pageCache).

When we get new allocatePage request, we findResuable() page which removes it from cache and adds back to m_Pages (active page list)

src/cuda/cuda-command.cpp
szihs: I didn't think about it. Let me make it explicit so we don't have to store the stream..

src/cuda/cuda-command.cpp
szihs: Yes, that's right. We are making it lazy retirement.  In case of single stream it helps us to skip creating events entirely. 

The trigger for retirement occurs from pool exhaustion, explicit waitOnHost, destructor calls

ccummingsNV: So my concern here is that:
- command buffers have a list of objects they hold references to
- one of those objects is shader objects / resources that are bound during encoding
- those resources may be holding on to heap memory

i.e. in this case, some memory may actually be remaining allocated until the commmand buffers are freed, which I'm not sure is the behaviour you're after right?

szihs: Added `trackResourcesForCUDA` which skips tracking resources for  MemoryType::DeviceLocal and same stream use case.

src/cuda/cuda-command.cpp
skallweitNV: Yes, totally fine during destruction.

src/cuda/cuda-command.cpp
```diff
@@ -736,24 +762,64 @@ void CommandQueueImpl::retireCommandBuffer(CommandBufferImpl* commandBuffer)
     }
 }
 
+void CommandQueueImpl::retireCommandBufferLocked(CommandBufferImpl* commandBuffer)
+{
+    // NOTE: Caller must hold m_mutex!
```
skallweitNV, line 788: I've always wondered how to make these kind of functions that expect a lock to be hold safer. But the only thing I can think of is taking the `lock` as an argument as proof of the mutex being locked. But you could still happily pass in a lock on a different mutex, so meh.

src/cuda/cuda-command.cpp
```diff
@@ -736,24 +762,64 @@ void CommandQueueImpl::retireCommandBuffer(CommandBufferImpl* commandBuffer)
...
+    {
+        // Single-stream lazy mode: use cuStreamQuery (non-blocking)
+        CUresult result = cuStreamQuery(m_stream);
+        if (result == CUDA_SUCCESS)
+        {
+            // Stream is idle - all submitted work is complete
+            m_lastFinishedID = m_lastSubmittedID;
+        }
+        // CUDA_ERROR_NOT_READY means work is still pending - that's fine, we're async
+        // Other errors would indicate a more serious problem, but we don't fail here
```
skallweitNV: So what happens to errors other than `CUDA_ERROR_NOT_READY`, we just ignore them?

src/cuda/cuda-command.cpp
szihs: added proper cleanup

src/cuda/cuda-command.cpp
szihs: I can add SLANG_RETURN_ON_FAIL for defensive logic. However we should not be hitting those.
Maybe we should add SLANG_RHI_ASSERT - What do you think?

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+    allocDesc.size = 256 * 1024; // 256KB
+    allocDesc.alignment = 128;
+
+    // Perform many allocation/free cycles
+    for (int cycle = 0; cycle < 10; cycle++)
+    {
+        std::vector<HeapAlloc> allocations;
+
+        // Allocate multiple blocks
+        for (int i = 0; i < 5; i++)
```
skallweitNV: 5*10 allocations is hardly a "stress test" :)

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    HeapAllocDesc allocDesc;
+    allocDesc.size = allocSize;
+    allocDesc.alignment = 128;
+
+    // First allocation - write a pattern via GPU shader
+    HeapAlloc alloc1;
+    REQUIRE_CALL(heap->allocate(allocDesc, &alloc1));
+
+    runInitPointerShader(device, 0xDEADBEEF, alloc1.getDeviceAddress(), numElements);
```
skallweitNV, line 429: you do write but never check the results ..

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+    CHECK_EQ(report.totalAllocated, 0);
+    CHECK_EQ(report.numAllocations, 0);
+    CHECK_EQ(report.numPages, 1); // Page cached
+
+    // Second allocation - should reuse the cached page
+    HeapAlloc alloc2;
+    REQUIRE_CALL(heap->allocate(allocDesc, &alloc2));
+
+    // Write a different pattern
+    runInitPointerShader(device, 0xCAFEBABE, alloc2.getDeviceAddress(), numElements);
```
skallweitNV, line 450: another write without any reads .. 

it's not very obvious why would even write in this test, if success/failure of it doesn't affect the test's result. what exactly are we testing?

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+    allocDesc.alignment = 128;
+
+    // Run multiple iterations - each reuses memory from previous
+    for (int iteration = 0; iteration < 20; iteration++)
+    {
+        HeapAlloc alloc;
+        REQUIRE_CALL(heap->allocate(allocDesc, &alloc));
+
+        // Write a unique pattern for this iteration
+        uint32_t pattern = 0xABCD0000 | iteration;
```
skallweitNV, line 488: why write a different pattern if we never check anything? or am i missing something?

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+            HeapAlloc alloc;
+            REQUIRE_CALL(heap->allocate(allocDesc, &alloc));
+            allocations.push_back(alloc);
+        }
+
+        // Use all allocations in GPU work
+        for (size_t i = 0; i < allocations.size(); i++)
+        {
+            uint32_t pattern = (cycle << 16) | i;
+            runInitPointerShader(device, pattern, allocations[i].getDeviceAddress(), numElements);
```
skallweitNV, line 540: you get my point at this point :)

tests/test-caching-allocator.cpp
```diff
@@ -0,0 +1,511 @@
...
+    // Allocate source and destination
+    HeapAlloc srcAlloc, dstAlloc;
+    REQUIRE_CALL(heap->allocate(allocDesc, &srcAlloc));
+    REQUIRE_CALL(heap->allocate(allocDesc, &dstAlloc));
+
+    // Initialize source with pattern
+    runInitPointerShader(device, 0x12345678, srcAlloc.getDeviceAddress(), numElements);
+
+    // Copy from source to destination
+    runCopyPointerShader(device, srcAlloc.getDeviceAddress(), dstAlloc.getDeviceAddress(), numElements);
```
skallweitNV, line 594: and then?

tests/test-heap.cpp
```diff
@@ -155,7 +155,10 @@ GPU_TEST_CASE("heap-allocate", CUDA | Vulkan)
     CHECK_EQ(report.numPages, 0);
 }
 
-GPU_TEST_CASE("heap-submit", CUDA | Vulkan)
+// CUDA: PyTorch-style same-stream optimization - frees are IMMEDIATE
+// because CUDA stream FIFO ordering guarantees the GPU work using the
+// allocation will complete before any new work that might reuse the memory.
+GPU_TEST_CASE("heap-submit", CUDA)
```
skallweitNV: I don't know if it's better split the test into two or have a branch within the same test to check cuda vs. vulkan 

If split, we should find better names for these test. `heap-submit` suggests we test heap submit. What exactly is that?

tests/test-heap.cpp
```diff
@@ -503,28 +553,93 @@ GPU_TEST_CASE("heap-multiple-submits-pending-frees", CUDA | Vulkan)
...
+        allocDesc.size = 32 * 1024; // 32 KB (8192 uint32s)
+        allocDesc.alignment = 128;
+
+        HeapAlloc allocation;
+        REQUIRE_CALL(heap->allocate(allocDesc, &allocation));
+
+        // Actually USE the heap allocation - write a pattern to it via shader
+        // This creates GPU work that uses the allocation on the default stream
+        uint32_t numElements = (uint32_t)(allocDesc.size / sizeof(uint32_t));
+        runInitPointerShader(device, submitIndex + 1, allocation.getDeviceAddress(), numElements);
```
skallweitNV: all we do with `runInitPointerShader` is check that `allocation.getDeviceAddress()` is SOME valid address?!

include/slang-rhi.h
```diff
@@ -59,6 +59,11 @@ typedef size_t Offset;
 
 const uint64_t kTimeoutInfinite = 0xFFFFFFFFFFFFFFFF;
 
+/// Sentinel value indicating no stream context is available.
+/// Used in HeapAllocDesc::stream when allocating outside command encoding context.
+/// Note: nullptr/0 is NOT used as sentinel because it represents the default stream in CUDA.
+inline void* const kNoStream = reinterpret_cast<void*>(~uintptr_t(0));
```
ccummingsNV: I think we already have kInvalidCUDAStream in slang-rhi

src/cuda/cuda-command.cpp
ccummingsNV: Yeah - the 'correct' pattern is probably recursive mutexes, but its overkill here

src/cuda/cuda-command.cpp
```diff
@@ -920,17 +999,70 @@ Result CommandEncoderImpl::init()
...
-    rootObject->trackResources(m_commandBuffer->m_trackedObjects);
+    // Skip tracking device-local buffers - CUDA stream ordering guarantees safety
+    trackResourcesForCUDARoot(rootObject, m_commandBuffer->m_trackedObjects);
 
     BindingDataBuilder builder;
     builder.m_device = getDevice<DeviceImpl>();
     builder.m_bindingCache = &m_commandBuffer->m_bindingCache;
     builder.m_allocator = &m_commandBuffer->m_allocator;
     builder.m_constantBufferPool = &m_commandBuffer->m_constantBufferPool;
+    builder.m_stream = m_queue->m_stream; // Pass stream for multi-stream tracking
```
ccummingsNV: Does this work if the user is submitting on a requestedStream instead of the m_stream associated with the queue. (see cuda-command.cpp:899)

src/cuda/cuda-constant-buffer-pool.cpp
```diff
@@ -49,15 +49,15 @@ void ConstantBufferPool::reset()
     m_globalDataPool.reset(m_device);
 }
 
-Result ConstantBufferPool::allocate(size_t size, ConstantBufferMemType memType, Allocation& outAllocation)
+Result ConstantBufferPool::allocate(size_t size, ConstantBufferMemType memType, void* stream, Allocation& outAllocation)
```
ccummingsNV: If it's 1 constant buffer per command buffer, and 1 command buffer used per stream, does this need passing in every time or can it be stored?

tests/test-caching-allocator.cpp
szihs: Increased to 1M allocations

tests/test-caching-allocator.cpp
szihs: now writes patterns and verifies them to ensure the allocation is actually usable, not just a valid address.

tests/test-caching-allocator.cpp
szihs: Added verifyHeapPattern() helper that reads back GPU memory and verifies the pattern.

tests/test-heap.cpp
szihs: updated

ccummingsNV (approved):
LGTM

PR #627, opened 2026-01-22
Improve `static_vector` and `short_vector` containers

skallweitNV:
- Align implementations to `std::vector`
- Fix handling non-POD types among other things
- Add tests

Note: Most of this code was written and reviewed (through multiple iterations) by Claude 4.5 Opus. I've done a proper review at the end fixed small issues + added more cleanup.

### DIFF
--- a/src/core/short_vector.h
+++ b/src/core/short_vector.h
@@ -344,7 +344,7 @@ public:
     iterator erase(const_iterator pos)
     {
         SLANG_RHI_ASSERT(pos >= begin() && pos < end());
-        iterator it = begin() + (pos - cbegin());
+        iterator it = begin() + (pos - begin());
         if constexpr (std::is_trivially_copyable_v<T>)
         {
             std::memmove(it, it + 1, (end() - it - 1) * sizeof(T));
@@ -363,8 +363,8 @@ public:
     {
         SLANG_RHI_ASSERT(first >= begin() && first <= end());
         SLANG_RHI_ASSERT(last >= first && last <= end());
-        iterator it_first = begin() + (first - cbegin());
-        iterator it_last = begin() + (last - cbegin());
+        iterator it_first = begin() + (first - begin());
+        iterator it_last = begin() + (last - begin());
         if (it_first != it_last)
         {
             size_type num_erased = it_last - it_first;
@@ -388,7 +388,7 @@ public:
     iterator insert(const_iterator pos, const value_type& value)
     {
         SLANG_RHI_ASSERT(pos >= begin() && pos <= end());
-        size_type index = pos - cbegin();
+        size_type index = pos - begin();
 
         // Copy value before any modifications in case it references an element in this vector.
         value_type copy = value;
@@ -420,7 +420,7 @@ public:
     iterator insert(const_iterator pos, value_type&& value)
     {
         SLANG_RHI_ASSERT(pos >= begin() && pos <= end());
-        size_type index = pos - cbegin();
+        size_type index = pos - begin();
 
         // Move value before any modifications in case it references an element in this vector.
         value_type copy = std::move(value);
--- a/src/core/static_vector.h
+++ b/src/core/static_vector.h
@@ -260,7 +260,7 @@ public:
     iterator erase(const_iterator pos)
     {
         SLANG_RHI_ASSERT(pos >= begin() && pos < end());
-        iterator it = begin() + (pos - cbegin());
+        iterator it = begin() + (pos - begin());
         if constexpr (std::is_trivially_copyable_v<T>)
         {
             std::memmove(it, it + 1, (end() - it - 1) * sizeof(T));
@@ -279,8 +279,8 @@ public:
     {
         SLANG_RHI_ASSERT(first >= begin() && first <= end());
         SLANG_RHI_ASSERT(last >= first && last <= end());
-        iterator it_first = begin() + (first - cbegin());
-        iterator it_last = begin() + (last - cbegin());
+        iterator it_first = begin() + (first - begin());
+        iterator it_last = begin() + (last - begin());
         if (it_first != it_last)
         {
             size_type num_erased = it_last - it_first;
@@ -305,7 +305,7 @@ public:
     {
         SLANG_RHI_ASSERT(m_size < N);
         SLANG_RHI_ASSERT(pos >= begin() && pos <= end());
-        iterator it = begin() + (pos - cbegin());
+        iterator it = begin() + (pos - begin());
         if (it == end())
         {
             construct_at(m_size, value);
@@ -334,7 +334,7 @@ public:
     {
         SLANG_RHI_ASSERT(m_size < N);
         SLANG_RHI_ASSERT(pos >= begin() && pos <= end());
-        iterator it = begin() + (pos - cbegin());
+        iterator it = begin() + (pos - begin());
         if (it == end())
         {
             construct_at(m_size, std::move(value));


src/core/static_vector.h
```diff
@@ -22,89 +37,477 @@ class static_vector
...
-            push_back(value);
+        {
+            construct_at(m_size, value);
+            ++m_size;
+        }
     }
 
-    ~static_vector() = default;
+    /// Iterator range constructor.
+    template<typename InputIt, typename = std::enable_if_t<!std::is_integral<InputIt>::value>>
```
tdavidovicNV, line 80: Isn't this a weird condition to check? Why ban explicitly integral types? Is it a common pattern?

src/core/static_vector.h
```diff
@@ -22,89 +37,477 @@ class static_vector
...
+        return *ptr_at(m_size - 1);
+    }
+
+    [[nodiscard]] iterator begin() noexcept { return ptr_at(0); }
+    [[nodiscard]] const_iterator begin() const noexcept { return ptr_at(0); }
+    [[nodiscard]] const_iterator cbegin() const noexcept { return ptr_at(0); }
+
+    [[nodiscard]] iterator end() noexcept { return ptr_at(m_size); }
+    [[nodiscard]] const_iterator end() const noexcept { return ptr_at(m_size); }
+    [[nodiscard]] const_iterator cend() const noexcept { return ptr_at(m_size); }
```
tdavidovicNV, line 176: I assume we don't care about the reverse iterators?

src/core/static_vector.h
```diff
@@ -22,89 +37,477 @@ class static_vector
...
+    void construct_at(size_type index, Args&&... args)
+    {
+        ::new (static_cast<void*>(&m_storage[index])) T(std::forward<Args>(args)...);
+    }
+
+    void construct_range(size_type first, size_type last)
+    {
+        if constexpr (std::is_trivially_default_constructible_v<T>)
+        {
+            std::memset(ptr_at(first), 0, (last - first) * sizeof(T));
```
tdavidovicNV, line 454: Why even wipe the memory with 0?

src/core/static_vector.h
```diff
@@ -22,89 +37,477 @@ class static_vector
...
+
+    const_pointer ptr_at(size_type index) const noexcept
+    {
+        return std::launder(reinterpret_cast<const T*>(&m_storage[index]));
+    }
+
+    template<typename... Args>
+    void construct_at(size_type index, Args&&... args)
+    {
+        ::new (static_cast<void*>(&m_storage[index])) T(std::forward<Args>(args)...);
```
tdavidovicNV, line 447: Still not sure what exactly the `std::lauder` does, but assuming it is useful, we should probably do `::new (ptr_at(index)) T...` here to reap any benefits it has.

tests/test-short-vector.cpp
```diff
@@ -0,0 +1,981 @@
...
+        return *this;
+    }
+
+    ~LifetimeTracker() { ++s_destructCount; }
+};
+
+int LifetimeTracker::s_constructCount = 0;
+int LifetimeTracker::s_destructCount = 0;
+int LifetimeTracker::s_copyCount = 0;
+int LifetimeTracker::s_moveCount = 0;
```
tdavidovicNV: I think ints can be defined inside the struct?

src/core/short_vector.h
```diff
@@ -81,187 +32,515 @@ class short_vector
...
     short_vector(const short_vector& other)
-        : short_vector()
+        : m_size(0)
+        , m_capacity(N)
     {
-        grow(other.m_size);
-        detail::copy_range(other.m_data, other.m_data + other.m_size, m_data);
+        m_data = inline_buffer();
+        reserve(other.m_size);
+        copy_construct_range(other.m_data, other.m_data + other.m_size, m_data);
```
tdavidovicNV, line 100: I'd probably use `.data()` and `.size()` here, for clarity

src/core/short_vector.h
```diff
@@ -81,187 +32,515 @@ class short_vector
...
+            reserve(other.m_size);
+            copy_construct_range(other.m_data, other.m_data + other.m_size, m_data);
+            m_size = other.m_size;
         }
-        else
+        return *this;
+    }
+
+    /// Move assignment.
+    short_vector& operator=(short_vector&& other) noexcept(std::is_nothrow_move_constructible<T>::value)
```
tdavidovicNV, line 149: We only allow move/assignment from short_vectors of the same inline length. That might be an unnecessary restriction. (But LLVM implementation to not do that is rather crazy CRTP, so maybe not worth it)

src/core/short_vector.h
```diff
@@ -81,187 +32,515 @@ class short_vector
...
+            }
+            m_size -= num_erased;
+        }
+        return it_first;
+    }
+
+    iterator insert(const_iterator pos, const value_type& value)
+    {
+        SLANG_RHI_ASSERT(pos >= begin() && pos <= end());
+        size_type index = pos - cbegin();
```
tdavidovicNV: Here we subtract `cbegin` but compare with `begin`, feels inconsistent.

src/core/short_vector.h
```diff
@@ -81,187 +32,515 @@ class short_vector
...
+
+        // Copy value before any modifications in case it references an element in this vector.
+        value_type copy = value;
 
         if (m_size == m_capacity)
             grow(m_capacity * 2);
-        new (m_data + m_size) T(std::forward<Args>(args)...);
-        m_size++;
+
+        iterator it = m_data + index;
```
tdavidovicNV, line 399: And here we ignore the c/begin completely and get iterator from `m_data` rather than `begin() + index`

src/core/short_vector.h
```diff
@@ -270,22 +549,119 @@ class short_vector
...
     {
         if (new_capacity <= m_capacity)
             return;
-        m_capacity = new_capacity;
-        value_type* new_data = reinterpret_cast<value_type*>(std::malloc(m_capacity * sizeof(T)));
-        detail::move_range(m_data, m_data + m_size, new_data);
-        if ((void*)m_data != (void*)m_short_data)
-            std::free(m_data);
+
+        pointer new_data = static_cast<pointer>(std::malloc(new_capacity * sizeof(T)));
```
tdavidovicNV, line 581: Is this gonna correctly align with `T`? e.g., if the `T` contains SSE data

src/core/short_vector.h
```diff
@@ -270,22 +549,119 @@ class short_vector
...
+    static void construct_at(pointer p, Args&&... args)
+    {
+        ::new (static_cast<void*>(p)) T(std::forward<Args>(args)...);
+    }
+
+    static void construct_range(pointer first, pointer last)
+    {
+        if constexpr (std::is_trivially_default_constructible_v<T>)
+        {
+            std::memset(first, 0, (last - first) * sizeof(T));
```
tdavidovicNV, line 611: Not sure we need/want to clear memory here.

tests/test-short-vector.cpp
```diff
@@ -0,0 +1,981 @@
...
+        vec.push_back(1);
+        vec.push_back(2);
+        CHECK(vec.is_inline());
+
+        vec.push_back(3);
+        CHECK(!vec.is_inline());
+        CHECK(vec.size() == 3);
+        CHECK(vec[0] == 1);
+        CHECK(vec[1] == 2);
+        CHECK(vec[2] == 3);
```
tdavidovicNV, line 163: Maybe add capacity check as well?

src/core/static_vector.h
skallweitNV: It's to disambiguate from the constructor that takes a size and a default value, in case the default value is an integer type.

src/core/static_vector.h
skallweitNV: We could add those, good point.

src/core/static_vector.h
skallweitNV: It's what std::vector does

src/core/static_vector.h
skallweitNV: it's to avoid the compiler from certain optimizations in case that memory was used by a different object before casted to this type

src/core/short_vector.h
skallweitNV: we're dealing with implementation here, using `.data()` or `.size()` are meant for client side access. this is consistent with the rest of the code i would say.

src/core/short_vector.h
skallweitNV: thought about this too, but it's not worth it IMO

src/core/short_vector.h
skallweitNV: malloc AFAIK does align to SSE types these days

src/core/short_vector.h
skallweitNV: again, std::vector behavior

tests/test-short-vector.cpp
skallweitNV: only if const

src/core/static_vector.h
skallweitNV: the placement new does the "laundering" implicitly .. it knows that it the passed in pointer is just arbitrary data and that the thing it returns is an object constructed at this location. tbh i think the whole std::launder thing could just be left out and would probably not matter in any situation we deal with.

PR #633, opened 2026-01-29
Request fp8/bfloat16/coop_mat2 on vk device creation.

csyonghe:


include/slang-rhi.h
```diff
@@ -129,6 +129,7 @@ enum class DeviceType
     x(RealtimeClock,                            "realtime-clock"                                ) \
     x(CooperativeVector,                        "cooperative-vector"                            ) \
     x(CooperativeMatrix,                        "cooperative-matrix"                            ) \
+    x(CooperativeMatrix2,                        "cooperative-matrix-2"                         ) \
```
```suggestion
    x(CooperativeMatrix2,                       "cooperative-matrix-2"                          ) \
```

include/slang-rhi.h
```diff
@@ -150,6 +151,8 @@ enum class DeviceType
     x(WaveOps,                                  "wave-ops"                                      ) \
     x(MeshShader,                               "mesh-shader"                                   ) \
     x(Pointer,                                  "has-ptr"                                       ) \
+    x(Float8,                                   "fp8"                                           ) \
```
skallweitNV, line 155: Are these features available on D3D12 and/or CUDA? Will they ever be?
If not they probably shouldn't be device features, but just be queries as capabilities by clients.

skallweitNV (dismissed):
LGTM

include/slang-rhi.h
csyonghe: fp8/bfloat16 is available on cuda but fp8 with cooperative matrix on cuda is still being implemented in slang.

no update on d3d12 when they will have this.

tests/test-cooperative-vector.cpp
```diff
@@ -187,6 +187,15 @@ GPU_TEST_CASE("cooperative-vector-get-matrix-size", D3D12 | Vulkan | CUDA)
         for (CooperativeVectorComponentType type : componentTypes)
         {
             PRINT("  Component Type: %s\n", toString(type).c_str());
+            if (type == CooperativeVectorComponentType::FloatE4M3 || type == CooperativeVectorComponentType::FloatE5M2)
+            {
+                // Not all devices support fp8, skip if not available.
+                if (!device->hasFeature(Feature::Float8))
```
skallweitNV: This check will only work on Vulkan, right? so now we always skip these formats on D3D12 and OptiX which is not great. I'm surprised that `getCooperativeVectorProperties` which we use to determine the formats to test reports these to be supported while they are not?!

csyonghe: if you want we could hard code another `&& device->getType() == Vulkan` condition to this logic.

skallweitNV: Why are these formats advertised to be supported by Vulkan CoopVec then? Is that a driver bug?
It just feels wrong to skip this test based on these IMO unrelated extensions being available. If we have to, then let's just do it for Vulkan and add a comment as to why this is necessary, because I still don't really understand why it is needed.

PR #634, opened 2026-01-29
Add support for creating combined texture sampler descriptor handles

skallweitNV:


src/d3d12/d3d12-texture.cpp
```diff
@@ -398,6 +398,22 @@ Result TextureViewImpl::getDescriptorHandle(DescriptorHandleAccess access, Descr
...
+    SLANG_RETURN_ON_FAIL(getDescriptorHandle(DescriptorHandleAccess::Read, &textureHandle));
+    Sampler* sampler = m_sampler ? m_sampler : m_texture->m_sampler;
+    if (!sampler)
+    {
+        return SLANG_FAIL;
+    }
+    DescriptorHandle samplerHandle;
+    SLANG_RETURN_ON_FAIL(sampler->getDescriptorHandle(&samplerHandle));
+    outHandle->type = DescriptorHandleType::CombinedTextureSampler;
+    outHandle->value = textureHandle.value | (samplerHandle.value << 32);
```
tdavidovicNV, line 413: I am just always super carefuly with the types in shifts, so I'd probably do `UINT64_C(32)` here

src/vulkan/vk-texture.h
```diff
@@ -76,7 +76,7 @@ class TextureViewImpl : public TextureView
     virtual void makeInternal() override { m_texture.breakStrongReference(); }
 
     BreakableReference<TextureImpl> m_texture;
-    DescriptorHandle m_descriptorHandle[2] = {};
+    DescriptorHandle m_descriptorHandle[3] = {};
```
tdavidovicNV, line 80: It is not immediately obvious why this goes from 2 to 3

PR #639, opened 2026-02-02
Indirect compute dispatch tests + CUDA emulation

skallweitNV:


src/cuda/cuda-command.cpp
```diff
@@ -493,8 +493,30 @@ void CommandExecutor::cmdDispatchCompute(const commands::DispatchCompute& cmd)
...
+        return;
+
+    // Emulate indirect dispatch by reading the dispatch arguments from the buffer on the host.
+    // This requires a GPU->CPU synchronization and stalls the GPU pipeline, but CUDA does not provide
+    // a way to perform indirect dispatches natively.
+    // In the future, we might consider using CUDA dynamic parallelism to launch a helper kernel
+    // that performs the actual dispatch, but that would require more setup and complexity.
+    BufferImpl* argBuffer = checked_cast<BufferImpl*>(cmd.argBuffer.buffer);
+    CUdeviceptr srcPtr = (CUdeviceptr)argBuffer->m_cudaMemory + cmd.argBuffer.offset;
+
```
ccummingsNV, line 506: This would be faster if we use the constant buffer allocator if possible, as it allocates pinned host memory.

skallweitNV: Yeah I thought about this too. However, I think down the road the correct implementation is CUDA dynamic parallelism. We would have a trampoline kernel that reads the dispatch arguments from the buffer and then launch a kernel within the kernel. This is more involved though, so this emulation is probably good enough for now.

ccummingsNV (changes requested):
Just noticed a bug - I think you need to sync on the stream that's being used (or the whole context) before you can safely do this. Not sure exactly what the cuMemcpyDtoH does, but if it doesn't block execution on all streams, then the user could call this function  **before** the dynamic dispatch args have been written by the gpu.

That might be worth fixing - and maybe add a test that does actually generate the arguments from the gpu?

PR #641, opened 2026-02-03
 Fix #640: Add waitUntilCompleted() in retireCommandBuffers() to ensure memory   visibility

### LINKED ISSUE #640: Metal: Race condition in command buffer retirement can cause stale texture reads
## Summary

A race condition in the Metal backend's command buffer retirement logic can cause `readTexture()` to return stale/zero data when the GPU completes work quickly. This manifests as flaky test failures, particularly in CI environments.

## Root Cause

The `retireCommandBuffers()` function is called at the end of every `submit()`, which can remove command buffers from tracking based on `status()` alone - without ensuring memory visibility via `waitUntilCompleted()`. When a subsequent `waitOnHost()` runs, it may skip synchronization entirely if the tracking event has already reached the target value.

## Timeline of the Race (Fast GPU Scenario)

```
CPU Thread                          │ GPU
────────────────────────────────────┼────────────────────────────────
1. SlangPy submit(compute)          │
   - m_lastSubmittedID = N          │
   - buffer added to inFlight       │
   - commit()                       │──→ Start compute work
   - retireCommandBuffers()         │    │
     (buffer still executing)       │    │
                                    │    ↓
2. readTexture() submit(copy)       │ Compute COMPLETES (signals N)
   - m_lastSubmittedID = N+1        │    │
   - copy buffer added              │    │
   - commit()                       │──→ │ Start copy work
   - retireCommandBuffers()         │    │
     ⚠️ compute buffer status =     │    │
        Completed, so RETIRED!      │    │
                                    │    ↓
3. waitOnHost()                     │ Copy COMPLETES (signals N+1)
   - updateLastFinishedID()         │
     returns N+1                    │
   - Check: N+1 < N+1? NO!          │
     ⚠️ SKIP shared event wait!     │
   - Loop m_commandBuffersInFlight  │
     Only copy buffer is there      │
   - waitUntilCompleted(copy)       │
     (already done)                 │
                                    │
4. mapBuffer() for read             │
   - synchronizeResource()          │
   - waitUntilCompleted()           │
                                    │
⚠️ Problem: We never explicitly     │
   waited for compute buffer's      │
   memory writes to be visible!     │
```

## The Memory Visibility Gap

The issue is subtle: even though Metal executes command buffers in commit order, `waitUntilCompleted()` provides memory visibility guarantees, but we only call it on buffers still in `m_commandBuffersInFlight`. If a buffer was retired based on `status()` alone, we might miss the memory barrier.

```cpp
// Current code in retireCommandBuffers():
if (status == MTL::CommandBufferStatusCompleted)
    commandBuffer->reset();  // ← Just removes from tracking, no wait!

// The distinction:
// - status() == Completed means GPU finished executing commands
// - waitUntilCompleted() ensures memory operations are visible to CPU
// On managed storage mode buffers, this distinction matters for coherency.
```

## Observed Behavior

- Flaky `test_toy_restir[metal]` failures in CI
- Texture reads return zeros for portions that should have data
- Same code passes on retry (timing-dependent)
- Only affects Metal backend (CUDA uses hard `cuStreamSynchronize()`)

## References

- Failing CI: https://github.com/shader-slang/slangpy/actions/runs/21593175887/job/62218326179
- Retry that passed (same code): https://github.com/shader-slang/slangpy/actions/runs/21593175887/job/62249050159

szihs:
## Summary

  Adds `waitUntilCompleted()` call in `retireCommandBuffers()` to fix a race
  condition where command buffers can be retired based on `status()` alone, without
   ensuring memory visibility to the CPU.

  ## The Fix

  Add `waitUntilCompleted()` before retiring command buffers. This ensures all GPU
  writes are visible to the CPU (required for managed storage mode buffers).

  ```cpp
  if (status == MTL::CommandBufferStatusCompleted || status ==
  MTL::CommandBufferStatusError)
  {
      // Ensure memory visibility before retiring.
      commandBuffer->m_commandBuffer->waitUntilCompleted();
      commandBuffer->reset();
  }
  ```

  Alternatives Considered
  Approach: Remove retireCommandBuffers() from submit()
  Tradeoff: Delays cleanup until waitOnHost(), changes behavior

  Choosen Approach: Add waitUntilCompleted() in retireCommandBuffers() 
  Tradeoff: Chosen: ensures memory visibility whenever we retire, correct fix

### DIFF
--- a/src/metal/metal-command.cpp
+++ b/src/metal/metal-command.cpp
@@ -990,6 +990,10 @@ void CommandQueueImpl::retireCommandBuffers()
         auto status = commandBuffer->m_commandBuffer->status();
         if (status == MTL::CommandBufferStatusCompleted || status == MTL::CommandBufferStatusError)
         {
+            // Ensure memory visibility before retiring.
+            // status() only checks the signal value, but waitUntilCompleted() ensures
+            // all GPU writes are visible to the CPU (required for managed storage mode).
+            commandBuffer->m_commandBuffer->waitUntilCompleted();
             commandBuffer->reset();
         }
         else


src/metal/metal-command.cpp
```diff
@@ -990,6 +990,10 @@ void CommandQueueImpl::retireCommandBuffers()
         auto status = commandBuffer->m_commandBuffer->status();
         if (status == MTL::CommandBufferStatusCompleted || status == MTL::CommandBufferStatusError)
         {
+            // Ensure memory visibility before retiring.
+            // status() only checks the signal value, but waitUntilCompleted() ensures
+            // all GPU writes are visible to the CPU (required for managed storage mode).
+            commandBuffer->m_commandBuffer->waitUntilCompleted();
```
skallweitNV, line 996: Are we running any danger of blocking for a long time here? According to docs, `MTL::CommandBufferStatusCompleted` should only be set when the GPU work is done:

> MTLCommandBufferStatus.completed
> A command buffer’s successful, final state, which indicates the GPU finished running the command buffer’s commands without any problems.

The docs on `waitUntilCompleted()` say:

> waitUntilCompleted()
> Blocks the current thread until the GPU finishes executing the command buffer and all of its completion handlers.

So it's just the completion handlers that we wait for here? Can these block for a long time?

szihs:
I would need more analysis on this. Moving to draft

skallweitNV:
@szihs what is the state of this PR? Do you still plan on working on this or should we close it?

szihs:
I am not planning to work on this. Closing

PR #642, opened 2026-02-04
Reduce CUDA context push/pop overhead

szihs:
## Release Notes

* **New Features**
  * Added three new CUDA context management methods: `setCudaContextCurrent()`, `pushCudaContext()`, and `popCudaContext()` to enable explicit control of CUDA context state per thread; no-op for non-CUDA devices.

* **Refactor**
  * Restructured internal CUDA context handling from automatic per-operation scoping to explicit context management through new API methods.

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -163,8 +163,9 @@ enum class DeviceType
     /* Vulkan specific features */                                                                \
     x(ShaderResourceMinLod,                     "shader-resource-min-lod"                       ) \
     /* Metal specific features */                                                                 \
-    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        )
-// clang-format on
+    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        ) \
+    /* CUDA specific features */                                                                  \
+    x(AtomicBfloat16,                           "atomic-bfloat16"                               ) // clang-format on
 
 #define SLANG_RHI_FEATURE_X(e, _) e,
 enum class Feature
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -831,6 +831,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED)
         tests/test-acceleration-structure-creation-with-validation.cpp
         tests/test-aftermath.cpp
         tests/test-arena-allocator.cpp
+        tests/test-device-features.cpp
         tests/test-block-allocator.cpp
         tests/test-benchmark-command.cpp
         tests/test-bindless.cpp
--- a/src/cuda/cuda-device.cpp
+++ b/src/cuda/cuda-device.cpp
@@ -384,6 +384,12 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
                 addCapability(cc.capability);
             }
         }
+
+        // BFloat16 atomic operations require SM 9.0 (Hopper) or higher
+        if (major >= 9)
+        {
+            addFeature(Feature::AtomicBfloat16);
+        }
     }
 
     optix::ContextDesc optixContextDesc = {};
--- a/tests/test-device-features.cpp
+++ b/tests/test-device-features.cpp
@@ -1,0 +1,25 @@
+// test-device-features.cpp
+// Device feature tests (e.g. skip pattern when a feature is not available).
+
+#include "testing.h"
+
+using namespace rhi;
+using namespace rhi::testing;
+
+// Test that demonstrates skipping when AtomicBfloat16 is not available.
+// This pattern should be used by tests that require bfloat16 atomic operations.
+GPU_TEST_CASE("atomic-bfloat16", CUDA)
+{
+    REQUIRE(device);
+
+    if (!device->hasFeature(Feature::AtomicBfloat16))
+    {
+        SKIP("AtomicBfloat16 not supported (requires SM 9.0/Hopper or newer)");
+    }
+
+    // If we reach here, the device supports AtomicBfloat16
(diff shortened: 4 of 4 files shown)


src/cuda/cuda-surface.cpp
```diff
@@ -857,7 +842,8 @@ Result SurfaceImpl::acquireNextImage(ITexture** outTexture)
 
 Result SurfaceImpl::present()
 {
-    SLANG_CUDA_CTX_SCOPE(m_deviceImpl);
+    // Set CUDA context current for this thread (required for cuStreamSynchronize).
+    m_deviceImpl->setCudaContextCurrent();
```
ccummingsNV: This should be a push/pop

src/cuda/cuda-utils.cpp
```diff
@@ -10,34 +10,25 @@ static thread_local std::atomic<uint32_t> g_contextStackDepth = 0;
...
-    // seemingly has significant impacts on cpu and gpu. This should be
-    // removed in favour of just not pushing/popping context regularly,
-    // as cuCtxGetCurrent is not free.
-    CUcontext currentContext = nullptr;
-    cuCtxGetCurrent(&currentContext);
-    if (currentContext != device->m_ctx.context)
-    {
-        m_didPush = true;
-        SLANG_CUDA_ASSERT_ON_FAIL(cuCtxPushCurrent(device->m_ctx.context));
+    // Always push context - this is only used in destructors (cold path).
```
ccummingsNV: We should still use the GetCurrent approach - it's still a perf boost when possible

src/cuda/cuda-surface.cpp
szihs: updated

src/cuda/cuda-utils.cpp
szihs: Sure thing. Added it back

ccummingsNV (dismissed):
LGTM

src/cuda/cuda-utils.h
```diff
@@ -18,6 +18,7 @@ namespace rhi::cuda {
 class DeviceImpl;
 
 /// Helper class to push/pop CUDA context on the stack.
+/// Checks if the correct context is already current before pushing (optimization).
 class ContextScope
```
skallweitNV, line 22: I would have expected all of these context helpers would be obsolete now? Do we still need them somewhere?

src/cuda/cuda-surface.cpp
```diff
@@ -779,6 +767,7 @@ void SurfaceImpl::destroySharedTexture(SharedTexture& sharedTexture)
 
 Result SurfaceImpl::configure(const SurfaceConfig& config)
 {
+    // Push/pop CUDA context to restore caller's context (e.g., PyTorch) after configuration.
     SLANG_CUDA_CTX_SCOPE(m_deviceImpl);
```
skallweitNV, line 771: Oh I see we still use the automatic context scope here.

include/slang-rhi.h
```diff
@@ -3559,6 +3513,18 @@ class IDevice : public ISlangUnknown
     /// @param heapCount [in/out] On input: size of heapReports buffer (ignored if heapReports is null). On output:
     /// number of heaps available or written
     virtual SLANG_NO_THROW Result SLANG_MCALL reportHeaps(HeapReport* heapReports, uint32_t* heapCount) = 0;
+
+    /// Set the device's CUDA context as current on this thread.
+    /// For non-CUDA devices, this is a no-op.
+    virtual SLANG_NO_THROW Result SLANG_MCALL setCudaContextCurrent() = 0;
```
skallweitNV, line 3566: I guess we could think having a separate `ICudaDevice` interface that can be queried to access this interface, but it might also be overkill.

src/device.h
```diff
@@ -357,6 +344,11 @@ class Device : public IDevice, public ComObject
     // Provides a default implementation that reports heaps from m_globalHeaps.
     virtual SLANG_NO_THROW Result SLANG_MCALL reportHeaps(HeapReport* heapReports, uint32_t* heapCount) override;
 
+    // Default no-op implementations for CUDA context management (only meaningful for CUDA backend).
+    virtual SLANG_NO_THROW Result SLANG_MCALL setCudaContextCurrent() override { return SLANG_OK; }
```
skallweitNV, line 363: Should these return `SLANG_E_NOT_AVAILABLE` instead?

src/cuda/cuda-utils.h
szihs: Yes - They are still needed for destructor calls, and inter op calls etc.

include/slang-rhi.h
szihs: Agreed, the benefits are minimal - slangpy currently unconditionally call this which is no-op for non-coda.

src/device.h
szihs: I intentionally return `SLANG_OK` so these can be called unconditionally regardless of the backend. 

Returning error would need caller side to handle error case. Happy to change if you feel strongly about it.

skallweitNV: No, you can leave it. Good to merge then.

skallweitNV (dismissed):

csyonghe (commented):
I think this change is fundamentally problematic -- it forces every user of slang-rhi to explicitly call setCurrentCUDAContext() after creating a cuda device since it removed all the CTX_SCOPE that is necessary to preserve correctness without user setting current context explicitly. Also I doubt how much actual perf gain you get by removing these. This is causing a big regression in my codebase.

Can we revert this?

szihs:
Added https://github.com/shader-slang/slang-rhi/pull/666 to avoid breaking change by setting cuda context if not already set.

PR #649, opened 2026-02-06
Implement deferred deletion of GPU resources.

skallweitNV:
This PR implements deferred release of GPU resources.

The idea is to keep GPU resources alive after being released on the host because the GPU might still be accessing them.
This is achieved by overriding the `deleteThis` method for all resource types we want to extend lifetime for. We then record the last submitted ID and the resource into a deferred delete queue which periodically gets drained.

This ensure that resources will be kept alive for all previous submitted command buffers when deleted.

This PR implements deferred release on D3D12, Vulkan, CUDA and Metal. It's not needed for the CPU backend (dispatches are synchronous for now), D3D11 (already handles that internally) and WebGPU (same).

### DIFF
--- a/src/core/ring-queue.h
+++ b/src/core/ring-queue.h
@@ -49,12 +49,11 @@ public:
     {
         m_buffer.resize(other.m_buffer.size());
         // Copy elements in order, compacting to the beginning.
-        size_type src = other.m_head;
-        for (size_type dst = 0; dst < other.m_size; ++dst)
+        for (size_type dst = 0, src = other.m_head; dst < other.m_size; ++dst, ++src)
         {
-            m_buffer[dst] = other.m_buffer[src];
-            if (++src >= other.m_buffer.size())
+            if (src >= other.m_buffer.size())
                 src = 0;
+            m_buffer[dst] = other.m_buffer[src];
         }
     }
 
@@ -83,12 +82,11 @@ public:
             m_tail = other.m_size;
             m_size = other.m_size;
             // Copy elements in order, compacting to the beginning.
-            size_type src = other.m_head;
-            for (size_type dst = 0; dst < other.m_size; ++dst)
+            for (size_type dst = 0, src = other.m_head; dst < other.m_size; ++dst, ++src)
             {
-                m_buffer[dst] = other.m_buffer[src];
-                if (++src >= other.m_buffer.size())
+                if (src >= other.m_buffer.size())
                     src = 0;
+                m_buffer[dst] = other.m_buffer[src];
             }
         }
         return *this;
@@ -333,12 +331,11 @@ private:
         std::vector<value_type> new_buffer(new_capacity);
 
         // Compact elements to the beginning of the new buffer.
-        size_type src = m_head;
-        for (size_type dst = 0; dst < m_size; ++dst)
+        for (size_type dst = 0, src = m_head; dst < m_size; ++dst, ++src)
         {
-            new_buffer[dst] = std::move(m_buffer[src]);
-            if (++src >= m_buffer.size())
+            if (src >= m_buffer.size())
                 src = 0;
+            new_buffer[dst] = std::move(m_buffer[src]);
         }
 
         m_buffer = std::move(new_buffer);


src/core/ring-queue.h
```diff
@@ -0,0 +1,348 @@
...
+
+    static constexpr size_type default_initial_capacity = 64;
+
+    /// Default constructor with optional initial capacity.
+    explicit RingQueue(size_type initial_capacity = default_initial_capacity)
+        : m_head(0)
+        , m_tail(0)
+        , m_size(0)
+    {
+        m_buffer.resize(initial_capacity > 0 ? initial_capacity : 1);
```
tdavidovicNV, line 41: This is not using `default_initial_capacity` but instead starts at 1 (which isn't the most useful value)

src/core/ring-queue.h
```diff
@@ -0,0 +1,348 @@
...
+
+    /// Grow the buffer to at least the specified capacity and compact elements.
+    void grow(size_type new_capacity)
+    {
+        std::vector<value_type> new_buffer(new_capacity);
+
+        // Compact elements to the beginning of the new buffer.
+        for (size_type i = 0; i < m_size; ++i)
+        {
+            new_buffer[i] = std::move(m_buffer[(m_head + i) % m_buffer.size()]);
```
tdavidovicNV: I'd generally try to avoid the `%`, especially since the values on the `lhs` aren't really gonna be more than double of the `rhs`, right?

In my experience, this is often faster:
```
while (lhs >= rhs) lhs -= rhs;
```

src/d3d12/d3d12-command.cpp
```diff
@@ -1850,10 +1859,34 @@ void CommandQueueImpl::retireCommandBuffers()
...
+    // Delete deferred resources that are no longer in use by the GPU.
+    executeDeferredDeletes();
+
     // Flush all device heaps
     getDevice<DeviceImpl>()->flushHeaps();
 }
 
+void CommandQueueImpl::deferDelete(Resource* resource)
+{
+    std::lock_guard<std::mutex> lock(m_deferredDeleteQueueMutex);
```
tdavidovicNV, line 1869: You can do:
```
std::lock_guard lock(m_deferredDeleteQueueMutex);
```
(no need for the explicit template arg)

src/core/ring-queue.h
```diff
@@ -0,0 +1,348 @@
...
+    const_iterator cbegin() const { return const_iterator(this, 0); }
+    const_iterator cend() const { return const_iterator(this, m_size); }
+
+private:
+    /// Ensure there is capacity for at least one more element.
+    void ensure_capacity()
+    {
+        if (m_size >= m_buffer.size())
+        {
+            grow(m_buffer.size() * 2);
```
tdavidovicNV, line 324: fwiw, the current MSVC vector grows to `max(current_capacity + current_capacity / 2, required_capacity)`.

src/core/ring-queue.h
skallweitNV: It does use `default_initial_capacity` as the default argument value, but this not very useful. Changing this to
```
    explicit RingQueue(size_type initial_capacity = 64)
```

src/d3d12/d3d12-command.cpp
skallweitNV: Yes, but we use the explicit version through most of the code. It helps when you want to search for specific mutex types. This might come in handy when we do the multi-threading work someday.

src/core/ring-queue.h
skallweitNV: Good point but I think I prefer the readability. Smells a lot like premature optimization to me.

skallweitNV: loops are now optimized

src/core/ring-queue.h
```diff
@@ -0,0 +1,355 @@
...
+        , m_size(other.m_size)
+    {
+        m_buffer.resize(other.m_buffer.size());
+        // Copy elements in order, compacting to the beginning.
+        size_type src = other.m_head;
+        for (size_type dst = 0; dst < other.m_size; ++dst)
+        {
+            m_buffer[dst] = other.m_buffer[src];
+            if (++src >= other.m_buffer.size())
+                src = 0;
```
ccummingsNV: Could be a MOD :)

src/core/ring-queue.h
```diff
@@ -0,0 +1,355 @@
...
+            m_buffer.resize(other.m_buffer.size());
+            m_head = 0;
+            m_tail = other.m_size;
+            m_size = other.m_size;
+            // Copy elements in order, compacting to the beginning.
+            size_type src = other.m_head;
+            for (size_type dst = 0; dst < other.m_size; ++dst)
+            {
+                m_buffer[dst] = other.m_buffer[src];
+                if (++src >= other.m_buffer.size())
```
ccummingsNV: Could also be a MOD

src/core/ring-queue.h
```diff
@@ -327,9 +333,12 @@ class RingQueue
...
 
         // Compact elements to the beginning of the new buffer.
-        for (size_type i = 0; i < m_size; ++i)
+        size_type src = m_head;
+        for (size_type dst = 0; dst < m_size; ++dst)
         {
-            new_buffer[i] = std::move(m_buffer[(m_head + i) % m_buffer.size()]);
+            new_buffer[dst] = std::move(m_buffer[src]);
+            if (++src >= m_buffer.size())
+                src = 0;
```
tdavidovicNV, line 337: I'd write the loop like this, but it is matter of personal taste (mostly the ++src in the increment part of the loop, rather than the body, the init I am not precious about at all)

```
for (size_type dst = 0, src = m_head; dst < m_size; ++dst, ++src)
{
    if (src >= m_buffer.size())
        src = 0;
    new_buffer[dst] = std::move(m_buffer[src]);
}
```

src/cuda/cuda-command.cpp
```diff
@@ -848,6 +860,23 @@ Result CommandQueueImpl::retireCommandBuffersLocked()
...
+{
+    std::lock_guard<std::mutex> lock(m_deferredDeleteQueueMutex);
+    m_deferredDeleteQueue.push({m_lastSubmittedID, resource});
+}
+
+void CommandQueueImpl::executeDeferredDeletes()
+{
+    uint64_t lastFinishedID = m_lastFinishedID;
+    std::lock_guard<std::mutex> lock(m_deferredDeleteQueueMutex);
+    while (!m_deferredDeleteQueue.empty() && m_deferredDeleteQueue.front().submissionID <= lastFinishedID)
```
ccummingsNV, line 873: I've we're being picky, the queue should really only be locked whilst you extract the list of things that need deleting, rather than maintaining a lock during potentially costly deletes. Until we're actually submitting from multiple threads though probably not worth worrying about.

src/core/ring-queue.h
tdavidovicNV: I made him remove the MOD. MOD is like 50+ cycles. This is comparatively free. Don't do non-power-of-2 MOD if you don't have to.

skallweitNV: I can change this back to a mod which I had to begin with. I don't think the 50 cycles really matter and I think the mod is slightly more readable, but I'm totally fine keeping as is. Just let me know.

src/cuda/cuda-command.cpp
skallweitNV: Agreed, we should probably refactor this when we get to do proper locking.

src/core/ring-queue.h
tdavidovicNV: It's 2:1 and based on the Tuesday discussion about overlaps, I am quite a bit less precious about CPU cycles than I otherwise would be. Go for mod, sorry for the noise.

PR #656, opened 2026-02-10
Debug rework cleanup

skallweitNV:


tests/testing.cpp
```diff
@@ -162,8 +136,11 @@ static CaptureDebugCallback sCaptureDebugCallback;
 class DebugCallback : public IDebugCallback
 {
 public:
-    bool shouldIgnoreError(DebugMessageType type, DebugMessageSource source, const char* message)
+    bool shouldIgnoreMessage(DebugMessageType type, DebugMessageSource source, const char* message)
     {
+        if (type != DebugMessageType::Error)
+            return false;
+
```
16-Bit-Dog, line 143: ~Downside of this change is that we won't get device initialization warnings (for example: failing to enable a non-required debug setting) even if `options().verbose` is enabled. At a minimum I think we should allow `DebugMessageType::Warning` and `DebugMessageType::Info` if `options().verbose` is enabled.~

Edit: I misread the `return false` for `return true`. PR looks good.

skallweitNV:
merged as part of #622

PR #658, opened 2026-02-10
add atomic_bfloat16 feature request

kaizhangNV:
The main purpose for this feature request is if target system doesn't have atomicBFloat16 supported, the executable that relies on it will know it's unsupported.

So for slang-test, we can use it to ignore the tests relying on it.

## Release Notes

* **New Features**
  * Added atomic BFloat16 operation support for CUDA devices with compute capability 9.0 and newer, enabling enhanced 16-bit floating-point atomic operations on Hopper architecture and subsequent generations.

* **Tests**
  * Expanded test suite with new test cases to validate atomic BFloat16 feature availability and ensure proper device capability detection across supported CUDA hardware configurations.


include/slang-rhi.h
```diff
@@ -163,7 +163,8 @@ enum class DeviceType
     /* Vulkan specific features */                                                                \
     x(ShaderResourceMinLod,                     "shader-resource-min-lod"                       ) \
     /* Metal specific features */                                                                 \
-    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        )
+    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        ) \
+    x(AtomicBfloat16,                           "atomic-bfloat16"                               )
```
skallweitNV: Maybe add a `CUDA specific feature` comment to mark this new feature as CUDA specific, which it seems to be.

tests/test-atomic-bfloat16.cpp
```diff
@@ -0,0 +1,70 @@
+// test-atomic-bfloat16.cpp
```
skallweitNV: I don't think this test is very useful to have. We don't have similar tests for other features and it would be quite a list to have if we want to. If we really want to test features availability, we should probably have a `test-device-features` test that has `CHECK`s for more than just a single feature.

kaizhangNV: I added a `test-device-features` test file, but there is only this test. I guess you will need to complete the list in the future.

tests/test-device-features.cpp
```diff
@@ -0,0 +1,25 @@
...
+    if (!device->hasFeature(Feature::AtomicBfloat16))
+    {
+        SKIP("AtomicBfloat16 not supported (requires SM 9.0/Hopper or newer)");
+    }
+
+    // If we reach here, the device supports AtomicBfloat16
+    // Double-check that SM 9.0 capability is also present
+    CHECK(device->hasCapability(Capability::_cuda_sm_9_0));
+    MESSAGE("Running test on device with AtomicBfloat16 support (SM 9.0+)");
+    CHECK(true);
```
skallweitNV, line 24: Can we remove this?

include/slang-rhi.h
```diff
@@ -163,8 +163,9 @@ enum class DeviceType
     /* Vulkan specific features */                                                                \
     x(ShaderResourceMinLod,                     "shader-resource-min-lod"                       ) \
     /* Metal specific features */                                                                 \
-    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        )
-// clang-format on
+    x(ArgumentBufferTier2,                      "argument-buffer-tier-2"                        ) \
+    /* CUDA specific features */                                                                  \
+    x(AtomicBfloat16,                           "atomic-bfloat16"                               ) // clang-format on
```
skallweitNV, line 168: not sure why `clang-format on` moved to the end of the line here ..

PR #663, opened 2026-02-17
Fail D3D12 Device creation if `setupDebugLayer` fails, otherwise succeed

16-Bit-Dog:
Currently if `setupDebugLayer` succeeds, d3d12 device creation fails.
```
    if (SLANG_SUCCEEDED(setupDebugLayer(d3dModule)))
        return SLANG_FAIL;
```

This has been changed to fail when device creation fails: `SLANG_FAILED`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an initialization logic issue that could prematurely abort device startup when diagnostic setup succeeded. Device initialization now continues correctly on successful diagnostic enablement and only fails for actual setup errors, improving startup reliability and reducing false-negative failures.

### DIFF
--- a/src/d3d12/d3d12-device.cpp
+++ b/src/d3d12/d3d12-device.cpp
@@ -313,7 +313,7 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
     }
 
     SLANG_RETURN_ON_FAIL(setupDebugLayer(d3dModule));
-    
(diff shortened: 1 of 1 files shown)


ccummingsNV:
I've taken this fix + applied in another PR, along with fixes for the tests themselves

PR #674, opened 2026-02-23
Add threaded task pool

skallweitNV:
- Add `ThreadedTaskPool` implementation
- Add `IRHI::initThreadPool(workerCount)` to allow initializing the thread pool (both blocked and threaded)
- Allow reconfiguring task pool when no devices are alive
- Improve documentation of `ITaskPool` interface

## Summary by CodeRabbit

* **New Features**
  * Multithreaded task pool with configurable worker threads and a public initTaskPool API to initialize or replace the global pool.
  * Task dependency support for ordered execution and immediate eligibility when no dependencies.
  * Explicit payload lifetime management and APIs to retrieve and release task payloads.

* **Behavior / Documentation**
  * Clarified task lifecycle, submit/release/wait/is-done semantics and waitAll behavior.
  * Debug layer options and task-pool init/set now fail while devices are alive.

* **Tests**
  * Added threaded task-pool test suite mirroring existing scenarios.

### DIFF
--- a/src/core/task-pool.cpp
+++ b/src/core/task-pool.cpp
@@ -435,6 +435,9 @@ static std::mutex s_globalTaskPoolMutex;
 static ComPtr<ITaskPool> s_globalTaskPool;
 static std::atomic<ITaskPool*> s_cachedGlobalTaskPool{nullptr};
 
+// WARNING: setGlobalTaskPool must only be called when no devices are alive
+// and no other threads are using the global task pool. Calling it concurrently
+// with globalTaskPool() may result in use-after-free.
 Result setGlobalTaskPool(ITaskPool* taskPool)
 {
     std::lock_guard<std::mutex> lock(s_globalTaskPoolMutex);


include/slang-rhi.h
```diff
@@ -3621,25 +3649,43 @@ class ITaskPool : public ISlangUnknown
         size_t depsCount
```
tdavidovicNV, line 3649: I know we do data/size but here I'd somehow expect count/what

include/slang-rhi.h
```diff
@@ -3742,8 +3788,16 @@ class IRHI
...
+    /// Fails if any devices are currently alive.
     virtual SLANG_NO_THROW Result SLANG_MCALL setTaskPool(ITaskPool* taskPool) = 0;
+
+    /// Initialize the global task pool with the specified number of worker threads.
+    /// A value of -1 (default) creates a `ThreadedTaskPool` with `std::thread::hardware_concurrency()` worker threads.
+    /// A value of 0 creates a `BlockingTaskPool` (tasks execute synchronously in the calling thread).
+    /// A positive value creates a `ThreadedTaskPool` with that many worker threads.
+    /// If not called before first use, a `ThreadedTaskPool` with `std::thread::hardware_concurrency()` threads is
+    /// created by default. Fails if any devices are currently alive.
+    virtual SLANG_NO_THROW Result SLANG_MCALL initTaskPool(int workerCount = -1) = 0;
```
tdavidovicNV, line 3800: One pattern I like but cannot really be expressed here is that `-1` means you create `std::thread::hardware_concurrency() - 1` threads. Means "run at full speed, but leave me 1 thread". (Usually you do like -2, one for UI, one for main thread).

src/core/task-pool.cpp
```diff
@@ -76,35 +83,393 @@ bool BlockingTaskPool::isTaskDone(TaskHandle task)
...
+                    if (!dep->done.load(std::memory_order_acquire))
+                    {
+                        // Add an extra reference that will be released when the dependency finishes.
+                        retainTask(task);
+                        dep->children.push_back(task);
+                    }
+                    else
+                    {
+                        // Dependency is already done, decrement the counter.
+                        if (task->depsRemaining.fetch_sub(1, std::memory_order_acq_rel) == 1)
```
tdavidovicNV: Nothing is touching this value in parallel, you can do relaxed, I think.
Or just count it on the side in a normal int.

src/core/task-pool.cpp
```diff
@@ -76,35 +83,393 @@ bool BlockingTaskPool::isTaskDone(TaskHandle task)
...
+    {
+        SLANG_RHI_ASSERT(task);
+        SLANG_RHI_ASSERT(task->pool == this);
+
+        {
+            std::lock_guard<std::mutex> lock(m_queueMutex);
+            m_queue.push(task);
+        }
+
+        m_queueCV.notify_one();
```
tdavidovicNV: I believe this can lead to more expensive code than if you just called it while still holding the mutex.

include/slang-rhi.h
skallweitNV: Yeah in that case you have to be explicit and call `initTaskPool(std::thread::hardware_concurrency() - 2)`.
Being explicit feels right to me here.

src/core/task-pool.cpp
skallweitNV: yeah, it can actually be memory_order_relaxed in both places

src/core/task-pool.cpp
skallweitNV: yep, seems to be the preferred practice. but more so, i was doing this inconsistently across different notifies, now they are consistent

PR #676, opened 2026-02-24
Vulkan raygen entrypoint parameter support

skallweitNV:
- Minor cleanup in D3D12 backend
- Add support for passing entrypoint parameters to raygen shaders in Vulkan
- Optimize shader table creation (lookups in hash maps)

## Summary by CodeRabbit

* **New Features**
  * Per-entry-point ray-tracing parameters and dynamic shader-record sizing so each ray-generation entry can carry its own parameters.
  * Device-level buffer-update command exposed for more flexible runtime buffer updates.
* **Bug Fixes**
  * Bounds validation for ray-generation indices to prevent out-of-range dispatches.
  * SBT construction now uses per-pipeline shader-table data for correct addresses and sizes.
* **Tests**
  * Several ray-tracing tests will now run on Vulkan devices.

### DIFF
--- a/src/vulkan/vk-command.cpp
+++ b/src/vulkan/vk-command.cpp
@@ -1074,7 +1074,7 @@ void CommandRecorder::cmdSetRayTracingState(const commands::SetRayTracingState& 
 
     bool updatePipeline = !m_rayTracingStateValid || cmd.pipeline != m_rayTracingPipeline;
     bool updateBindings = updatePipeline || cmd.bindingData != m_bindingData;
-    bool updateShaderTable = !m_rayTracingStateValid || cmd.shaderTable != m_shaderTable;
+    bool updateShaderTable = updatePipeline || cmd.shaderTable != m_shaderTable;
 
     auto& api = m_device->m_api;
 


src/d3d12/d3d12-command.cpp
```diff
@@ -1152,6 +1159,9 @@ void CommandRecorder::cmdDispatchRays(const commands::DispatchRays& cmd)
     if (!m_rayTracingStateValid)
         return;
 
+    if (cmd.rayGenShaderIndex >= m_shaderTable->m_rayGenShaderCount)
```
ccummingsNV, line 1162: Is this actually an error we need to catch?

src/vulkan/vk-command.cpp
```diff
@@ -1094,29 +1095,29 @@ void CommandRecorder::cmdSetRayTracingState(const commands::SetRayTracingState&
...
-        m_raygenSBT.size = m_shaderTable->m_raygenRecordStride;
 
-        m_missSBT.deviceAddress = shaderTableAddr + m_shaderTable->m_raygenTableSize;
-        m_missSBT.stride = m_shaderTable->m_missRecordStride;
-        m_missSBT.size = m_shaderTable->m_missTableSize;
+        m_missSBT.deviceAddress = shaderTableAddr + m_shaderTablePipelineData->raygenTableSize;
+        m_missSBT.stride = m_shaderTablePipelineData->missRecordStride;
+        m_missSBT.size = m_shaderTablePipelineData->missTableSize;
 
         m_hitSBT.deviceAddress = m_missSBT.deviceAddress + m_missSBT.size;
```
ccummingsNV, line 1114: Do these incremented addressed need aligning to 256 bytes, or is that implicit in the size?

src/vulkan/vk-command.cpp
```diff
@@ -1131,7 +1132,54 @@ void CommandRecorder::cmdDispatchRays(const commands::DispatchRays& cmd)
     if (!m_rayTracingStateValid)
         return;
 
-    m_raygenSBT.deviceAddress = m_rayGenTableAddr + cmd.rayGenShaderIndex * m_raygenSBT.stride;
+    if (cmd.rayGenShaderIndex >= m_shaderTable->m_rayGenShaderCount)
```
ccummingsNV, line 1135: This feels like another of those errors we need to catch, either in debug layer, or add some mechanism to catch errors during command record.

src/vulkan/vk-command.cpp
```diff
@@ -1131,7 +1132,54 @@ void CommandRecorder::cmdDispatchRays(const commands::DispatchRays& cmd)
...
+                0,
+                nullptr
+            );
+        }
+    }
+
+    // Set raygen SBT address and size based on the selected raygen shader.
+    // Each raygen shader can have a different record size.
+    m_raygenSBT.deviceAddress = m_rayGenTableAddr + raygenInfo.recordOffset;
+    m_raygenSBT.stride = raygenInfo.recordSize;
```
ccummingsNV, line 1198: Is this recordSize -> m_raygenSBT.stride deliberate?

src/vulkan/vk-shader-object.cpp
```diff
@@ -268,6 +268,16 @@ Result BindingDataBuilder::bindAsRoot(
     m_bindingData->pushConstantData = m_allocator->allocate<void*>(m_pushConstantRanges.size());
     m_bindingData->pushConstantCount = 0;
 
+    // Allocate entry point data storage for ray tracing SBT.
+    size_t entryPointCount = specializedLayout->m_entryPoints.size();
```
ccummingsNV, line 272: Should this only be for RTP? Or can we not detect that at this point?

src/vulkan/vk-shader-object.cpp
```diff
@@ -321,44 +330,43 @@ Result BindingDataBuilder::bindAsRoot(
...
+        }
+        else
+        {
+            // For non-raygen entry points, ordinary data goes into push constants.
+            //
+            // The index of the push constant range to bind is passed down as part of the
+            // `offset`, and we increment it here so that any further recursively-contained
+            // push-constant ranges use the next index.
+            //
+            auto pushConstantRangeIndex = offset.pushConstantRange++;
```
ccummingsNV, line 365: Prior to this work, I hit a crash deep in slang-rhi due to getting here from a raygen shader. Is it worth catching the error case of this being a different RTP entry point but still somehow have been allowed to get this far down the pipeline, rather than crashing when dereferencing at pushConstantRangeIndex 0.

src/vulkan/vk-shader-table.cpp
```diff
@@ -13,43 +14,92 @@ ShaderTableImpl::ShaderTableImpl(Device* device, const ShaderTableDesc& desc)
 {
 }
 
-BufferImpl* ShaderTableImpl::getBuffer(RayTracingPipelineImpl* pipeline)
+/// Find the entry point index in the root object layout by name.
+static uint32_t findEntryPointIndexByName(
```
ccummingsNV: How often is this called and how big can the list get? It could end up being responsible for a depressing amount of overhead in my experience so far, if it's per dispatch.

src/vulkan/vk-shader-table.cpp
```diff
@@ -13,43 +14,92 @@ ShaderTableImpl::ShaderTableImpl(Device* device, const ShaderTableDesc& desc)
...
+    // Build raygen infos and calculate per-raygen record sizes based on entry point params.
+    // Each raygen shader gets its own record size based on its actual parameter requirements.
+    uint32_t raygenTableOffset = 0;
+    short_vector<RaygenInfo> raygenInfos;
+
+    for (uint32_t i = 0; i < m_rayGenShaderCount; i++)
+    {
+        // Get the entry point index and parameter size for this raygen shader, if it exists in the root layout.
+        const std::string& entryPointName = m_rayGenShaderEntryPointNames[i];
+        uint32_t entryPointIndex = findEntryPointIndexByName(rootLayout, programLayout, entryPointName);
```
ccummingsNV: This is effectively an N^2 loop, if the ray gen shader count is != 1

skallweitNV: yeah, but you really only a few raygen entrypoints at max

src/vulkan/vk-shader-table.cpp
skallweitNV: this is called once per pipeline/shadertable and there are usually only a few raygen shaders at max per pipeline. so while i agree, i don't think that's gonna hit us.

src/vulkan/vk-command.cpp
skallweitNV: yeah, we just point to a single record.

src/vulkan/vk-command.cpp
skallweitNV: yeah, we should add more thorough validation. i'll try get AI to write some of this

src/vulkan/vk-command.cpp
skallweitNV: the assumption is that they are aligned properly done when the shader table buffer is created in `vk-shader-table.cpp`:
```c++
    // Calculate record sizes (without alignment).
    uint32_t missRecordSize = max(handleSize, m_missRecordOverwriteMaxSize);
    uint32_t hitGroupRecordSize = max(handleSize, m_hitGroupRecordOverwriteMaxSize);
    uint32_t callableRecordSize = max(handleSize, m_callableRecordOverwriteMaxSize);

    // Align all record sizes to shaderGroupBaseAlignment.
    missRecordSize = (uint32_t)math::calcAligned2(missRecordSize, rtpProps.shaderGroupBaseAlignment);
    hitGroupRecordSize = (uint32_t)math::calcAligned2(hitGroupRecordSize, rtpProps.shaderGroupBaseAlignment);
    callableRecordSize = (uint32_t)math::calcAligned2(callableRecordSize, rtpProps.shaderGroupBaseAlignment);

    // Calculate table sizes.
    uint32_t raygenTableSize = raygenTableOffset;
    uint32_t missTableSize = m_missShaderCount * missRecordSize;
    uint32_t hitTableSize = m_hitGroupCount * hitGroupRecordSize;
    uint32_t callableTableSize = m_callableShaderCount * callableRecordSize;
```

src/d3d12/d3d12-command.cpp
skallweitNV: i thought it's better to not crash and just do a no-op. we should catch this in the validation layer to inform users.

src/vulkan/vk-shader-table.cpp
skallweitNV: optimized now

src/vulkan/vk-shader-table.cpp
skallweitNV: optimized now

src/vulkan/vk-shader-object.cpp
skallweitNV: done

PR #680, opened 2026-03-10
Add ASTC compressed format support across multiple backends.

guoxx:
Support ASTC compression format for Metal, Vulkan and WebGPU.

## Summary by CodeRabbit

* **New Features**
  * Added ASTC texture variants (4x4, 6x6, 8x8; Unorm and sRGB) across graphics backends and format tables; WebGPU feature flag and runtime gating respected.

* **Bug Fixes / Behavior**
  * Alignment logic refined: offsets use down-alignment and extents use consistent alignment.
  * Compressed texture sizes normalized to block dimensions.
  * Multi-layer/mip texture copies now handle per-subresource extents more precisely.

* **Tests**
  * New and expanded tests covering ASTC formats and compressed NPOT mip scenarios.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -163,6 +163,7 @@
 
 | API                                    | CPU | CUDA | D3D11 | D3D12 | Vulkan | Metal | WGPU |
 |----------------------------------------|-----|------|-------|-------|--------|-------|------|
+| `getDesc`                              | yes | yes  | yes   | yes   | yes    | yes   | yes  |
 | `beginRenderPass`                      | :x: | :x:  | yes   | yes   | yes    | yes   | yes  |
 | `beginComputePass`                     | yes | yes  | yes   | yes   | yes    | yes   | yes  |
 | `beginRayTracingPass`                  | :x: | yes  | :x:   | yes   | yes    | :x:   | :x:  |
@@ -236,6 +237,7 @@
 
 | API               | CPU | CUDA | D3D11 | D3D12 | Vulkan | Metal | WGPU |
 |-------------------|-----|------|-------|-------|--------|-------|------|
+| `getDesc`         | yes | yes  | yes   | yes   | yes    | yes   | yes  |
 | `getNativeHandle` | :x: | :x:  | :x:   | yes   | yes    | yes   | yes  |
 
 ## `ICommandQueue` interface
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -78,6 +78,9 @@ enum class StructType
     DeviceDesc,
     HeapDesc,
 
+    CommandEncoderDesc,
+    CommandBufferDesc,
+
     D3D12DeviceExtendedDesc,
     D3D12ExperimentalFeaturesDesc,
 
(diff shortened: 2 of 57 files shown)


tests/texture-test.cpp
```diff
@@ -1078,6 +1078,12 @@ Format kStandardFormats[] = {
     Format::RGB10A2Unorm,
     Format::BC1Unorm,
     Format::BC1UnormSrgb,
+    Format::ASTC4x4Unorm,
+    Format::ASTC4x4UnormSrgb,
+    Format::ASTC6x6Unorm,
+    Format::ASTC6x6UnormSrgb,
+    Format::ASTC8x8Unorm,
+    Format::ASTC8x8UnormSrgb,
```
skallweitNV: Our strategy so far has been to test a good representative subset of formats here, not the full set of formats. There is an extra switch to enable testing of all formats. So probably a good subset here would be `ASTC4x4Unorm` and `ASTC6x6UnormSrgb`

tests/test-cmd-copy-texture-to-buffer.cpp
```diff
@@ -279,8 +279,8 @@ GPU_TEST_CASE("cmd-copy-texture-to-buffer-offset", D3D12 | Vulkan | Metal | WGPU
             // Pick offset for the extra copy.
             Extent3D size = data.desc.size;
             Offset3D offset = {size.width / 4, size.height / 4, size.depth / 4};
-            offset.x = math::calcAligned2(offset.x, data.formatInfo.blockWidth);
-            offset.y = math::calcAligned2(offset.y, data.formatInfo.blockHeight);
+            offset.x = math::calcAlignedDown(offset.x, data.formatInfo.blockWidth);
```
ccummingsNV, line 282: This alignment function represents a change in behaviour - 'calcAligned2' is a power-of-2 specific round up. Is there a reason this has been changed to a round down in lots of areas?

tests/test-cmd-copy-texture.cpp
```diff
@@ -696,6 +696,54 @@ GPU_TEST_CASE("cmd-copy-texture-acrossmips", D3D11 | D3D12 | Vulkan | WGPU | CUD
     );
 }
 
+GPU_TEST_CASE("cmd-copy-texture-compressed-npot-mips", WGPU)
+{
+    TextureTestOptions options(device);
```
ccummingsNV, line 701: Note that the texture tests support 'both' for mips+arrays. 
- Do we want this to test none mips as well? 
- Do we want to test texture arrays?

If both are supported, we should probably test them

tests/test-cmd-copy-texture-to-buffer.cpp
guoxx: This change is intentional for two reasons.

First, calcAligned2 only works for power-of-two alignments, but ASTC 6x6 is not power-of-two.

Second,  this is an offset, not an extent, so it should round down to the previous valid block boundary. Rounding up moves the copy origin forward and changes the tested region; for test cases the size is already rounded up separately, so rounding the origin up as well can overshoot on small/NPOT mips

tests/test-cmd-copy-texture.cpp
guoxx: More tests are added, cmdCopyTexture for metal are properly handle texture arrays and mipmap now.

PR #684, opened 2026-03-16
Take care of a CADrawable leak

Geenz:
Looking through - there was a bit of a problem where CADrawable would have a really bad tendency to leak from the swapchain.

For reference, I was seeing memory consumption balloon by the gigabytes over seconds of rendering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory and resource cleanup during rendering to reduce memory spikes and improve stability and responsiveness.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ set(SLANG_RHI_FETCH_SLANG_VERSION "2026.4.1" CACHE STRING "Slang version to fetc
 
 # DirectXShaderCompiler
 if(NOT DEFINED SLANG_RHI_DXC_URL)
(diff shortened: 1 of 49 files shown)


src/metal/metal-surface.cpp
```diff
@@ -97,10 +101,12 @@ Result SurfaceImpl::present()
         return SLANG_FAIL;
     }
 
+    AUTORELEASEPOOL
+
     MTL::CommandBuffer* commandBuffer = m_device->m_commandQueue->commandBuffer();
     commandBuffer->presentDrawable(m_currentDrawable.get());
     commandBuffer->commit();
-    commandBuffer->release();
+    // Note: you don't want to do a release here - just use the auto release pool.
```
skallweitNV: You can remove this comment.

Geenz: Should be good!

PR #697, opened 2026-03-25
Support CoopVec feature when using AgilitySDK preview

jkwak-work:
This PR is to support the Cooperative Vector feature when the AgilitySDK "PREVIEW" is used.

## Summary by CodeRabbit

* **New Features**
  * Added detection of cooperative vector support on compatible D3D12 devices, enabling applications to discover and utilize this hardware capability when available.

### DIFF
--- a/src/d3d12/d3d12-device.cpp
+++ b/src/d3d12/d3d12-device.cpp
@@ -848,6 +848,21 @@ Result DeviceImpl::initialize(const DeviceDesc& desc)
         }
     }
 
+#if SLANG_RHI_AGILITY_SDK_VERSION == D3D12_PREVIEW_SDK_VERSION
+#if defined(D3D12_PREVIEW_SDK_VERSION) && (D3D12_PREVIEW_SDK_VERSION >= 713)
+    {
+        // Check for cooperative vector support via native D3D12 Agility PREVIEW SDK.
+        D3D12_FEATURE_DATA_COOPERATIVE_VECTOR coopVecData = {};
+        if (SLANG_SUCCEEDED(
+                m_device->CheckFeatureSupport(D3D12_FEATURE_COOPERATIVE_VECTOR, &coopVecData, sizeof(coopVecData))
+            ))
+        {
+            addFeature(Feature::CooperativeVector);
+        }
+    }
+#endif
+#endif
+
     // Initialize NVAPI
 #if SLANG_RHI_ENABLE_NVAPI
     {


skallweitNV:
This feels incomplete. We're detecting coopvec support in D3D12 Agility SDK but then don't use any of it's APIs?!

jkwak-work:
> This feels incomplete. We're detecting coopvec support in D3D12 Agility SDK but then don't use any of it's APIs?!

Currently all of slang-test tests under `tests/cooperative-vector` directory with `-render-feature cooperative-vector` are not running at all, because the `render-feature` checking is failing.

Here is an example that shows dx12 test is ignored:
```
 build/Debug/bin/slang-test.exe tests/cooperative-vector/add.slang
Supported backends: fxc dxc glslang spirv-dis clang visualstudio genericcpp nvrtc llvm spirv-opt metal tint
Check vk,vulkan: Supported
Check dx12,d3d12: Supported
Check dx11,d3d11: Supported
Check cpu: Supported
Check cuda: Supported
Check wgpu,webgpu: Supported
Check llvm: Supported
passed test: 'tests/cooperative-vector/add.slang (vk)'
ignored test: 'tests/cooperative-vector/add.slang.1 (dx12)'   // <======= IGNORED!!!
passed test: 'tests/cooperative-vector/add.slang.2 (cpu)'
passed test: 'tests/cooperative-vector/add.slang.3'
passed test: 'tests/cooperative-vector/add.slang.4'
passed test: 'tests/cooperative-vector/add.slang.5 syn (cuda)'

===
100% of tests passed (5/5), 1 tests ignored
===
```

And here is how the test is setup:
```
//TEST(compute):COMPARE_COMPUTE(filecheck-buffer=CHECK):-dx12 -render-feature cooperative-vector -dx12-experimental -output-using-type -profile cs_6_9 -Xslang... -Xdxc -Vd -X.
```
https://github.com/shader-slang/slang/blob/425eced3fa7e5b5a34b54537aedab77064adf3da/tests/cooperative-vector/add.slang#L2

This PR is to allow them to run; but `cooperative-vector' is available only with `PREVIEW` of AgilitySDK.

And its version should be 713+; that's why there is a guard with `#if defined(D3D12_PREVIEW_SDK_VERSION) && (D3D12_PREVIEW_SDK_VERSION >= 713)`.

And the enum used in this PR, `D3D12_FEATURE_COOPERATIVE_VECTOR`, is available only with the PREVIEW variant; that's why the code is wrapped with `#if SLANG_RHI_AGILITY_SDK_VERSION == D3D12_PREVIEW_SDK_VERSION`.

I don't think there is anything specific to be done with API side to enable or disable; maybe we can query but it is not needed for now.

skallweitNV:
@jkwak-work what is the state of this? Should we update this PR or close it?

PR #700, opened 2026-04-03
Fix CUDA library discovery on Linux/WSL

csyonghe:
## Summary
- try `libcuda.so.1` before `libcuda.so` on Linux so CUDA device creation works on WSL
- search CUDA 12+ `targets/x86_64-linux/lib` locations when discovering NVRTC
- mention the additional Linux search location in the NVRTC error message

## Why
On WSL, the CUDA driver is typically exposed as `libcuda.so.1` rather than `libcuda.so`. Recent CUDA toolkit installs also place `libnvrtc.so` under `targets/x86_64-linux/lib`, which was not part of the Linux search path.

Without these two path updates, CUDA device creation fails even when CUDA on WSL is installed correctly.

## Validation
- rebuilt `slangpy` editable package on WSL
- created a CUDA `slangpy.Device(...)` successfully after the change

## Summary by CodeRabbit

* **Improvements**
  * Enhanced CUDA driver initialization on Linux to support additional library search paths including versioned shared library names.
  * Extended CUDA NVRTC library discovery on Linux to search additional target directories for improved library resolution.

### DIFF
--- a/src/cuda-driver-api.cpp
+++ b/src/cuda-driver-api.cpp
@@ -32,6 +32,7 @@ extern "C" bool rhiCudaDriverApiInit()
     };
 #elif SLANG_LINUX_FAMILY
     const char* cudaPaths[] = {
+        "libcuda.so.1",
         "libcuda.so",
         nullptr,
     };
--- a/src/cuda/cuda-nvrtc.cpp
+++ b/src/cuda/cuda-nvrtc.cpp
@@ -77,6 +77,7 @@ inline void findNVRTCPaths(std::vector<std::filesystem::path>& outPaths)
         {
             outPaths.push_back(std::filesystem::path(path) / "lib64");
             outPaths.push_back(std::filesystem::path(path) / "lib");
+            outPaths.push_back(std::filesystem::path(path) / "targets" / "x86_64-linux" / "lib");
         }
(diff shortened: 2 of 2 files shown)


src/cuda-driver-api.cpp
```diff
@@ -32,6 +32,7 @@ extern "C" bool rhiCudaDriverApiInit()
     };
 #elif SLANG_LINUX_FAMILY
     const char* cudaPaths[] = {
+        "libcuda.so.1",
         "libcuda.so",
```
jhelferty-nv, line 36: I would expect `libcuda.so` to come first in this list, since it's the more 'generic' library name, and is generally a symlink to `libcuda.so.1` anyways. If it fails, it seems reasonable to fall back to `libcuda.so.1`.

I'll approve for now, assuming there's some esoteric reason it needs to be this way, but if we can do it in the other order, I think that'd be more correct.

jhelferty-nv (approved):

src/cuda-driver-api.cpp
csyonghe: libcuda.so does existing in wsl and it points to an invalid library with a lot of functions missing, causing the device creation failure.

PR #701, opened 2026-04-04
Fix varLayout assertion failure when importing module with global ConstantBuffer

EvanLuo42:
## Problem

`IDevice::createShaderProgram` crashes with `assert failure: varLayout` when a shader imports a module that declares a global `ConstantBuffer<T>` variable. The same shader code compiles successfully with `slangc`.

**Reproduction:**
```slang
// MyData.slang (imported module)
public struct MyData { public float4 value; };
public ConstantBuffer<MyData> gData;

// Main.slang
import MyData;
[shader("compute")] [numthreads(1,1,1)]
void computeMain(uint3 tid : SV_DispatchThreadID) { ... gData.value ... }
```

```cpp
ShaderProgramDesc desc = {};
desc.slangGlobalScope = module;
desc.slangEntryPoints = entryPoints;
desc.slangEntryPointCount = 1;
device->createShaderProgram(desc, program.writeRef()); // crash
```

```
FATAL ERROR: [Error] [Slang] error[E99999]: Slang compilation
aborted due to an exception of class Slang::InternalError assert failure: varLayout
For assistance, file an issue on GitHub (https://github.com/shader-slang/slang/issues) or join the Slang Discord (https:
//khr.io/slangdiscord)

FATAL ERROR: REQUIRE( !((device->createCo
mputePipeline(pipelineDesc, pipeline.writeRef())) < 0) ) THREW exception: "unknown exception"
```

## Root Cause

`ShaderProgram::init()` creates a composite component type from the module and entry points via `createCompositeComponentType`, but never calls `link()` on it. When the Slang compiler later performs layout reflection or code generation on this unlinked composite, it hits an internal assertion because variable layout information has not been resolved.

## Fix

**`src/shader.cpp`**: Call `link()` on the composite program after `createCompositeComponentType` in the `SingleProgram` linking path.

## Test

Added `tests/test-imported-constant-buffer.cpp` with corresponding `.slang` files that reproduce the original crash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shader program linking and handling of composite programs to avoid linking/reflect issues.
  * CPU pipeline creation now uses the linked program for entry-point layout and host callable lookup.

* **Tests**
  * Added GPU tests for imported constant-buffer behavior, including a compute shader that validates buffer binding and results.

* **Chores**
  * Updated .gitignore to exclude JetBrains IDE metadata.

### DIFF
--- a/src/shader.cpp
+++ b/src/shader.cpp
@@ -56,11 +56,7 @@ Result ShaderProgram::init()
     }
 
     auto session = m_desc.slangGlobalScope ? m_desc.slangGlobalScope->getSession() : nullptr;
-    if (m_desc.linkingStyle == LinkingStyle::SingleProgram && m_desc.slangEntryPointCount == 0)
-    {
-        linkedProgram = m_desc.slangGlobalScope;
-    }
-    else if (m_desc.linkingStyle == LinkingStyle::SingleProgram)
+    if (m_desc.linkingStyle == LinkingStyle::SingleProgram)
     {
         std::vector<slang::IComponentType*> components;
         if (m_desc.slangGlobalScope)
--- a/src/shader.cpp
+++ b/src/shader.cpp
@@ -56,7 +56,11 @@ Result ShaderProgram::init()
     }
 
     auto session = m_desc.slangGlobalScope ? m_desc.slangGlobalScope->getSession() : nullptr;
-    if (m_desc.linkingStyle == LinkingStyle::SingleProgram)
+    if (m_desc.linkingStyle == LinkingStyle::SingleProgram && m_desc.slangEntryPointCount == 0)
+    {
+        linkedProgram = m_desc.slangGlobalScope;
+    }
+    else if (m_desc.linkingStyle == LinkingStyle::SingleProgram)
     {
         std::vector<slang::IComponentType*> components;
         if (m_desc.slangGlobalScope)
--- a/src/shader.cpp
+++ b/src/shader.cpp
@@ -71,9 +71,11 @@ Result ShaderProgram::init()
             }
             components.push_back(m_desc.slangEntryPoints[i]);
         }
+        ComPtr<slang::IComponentType> composedProgram;
         SLANG_RETURN_ON_FAIL(
-            session->createCompositeComponentType(components.data(), components.size(), linkedProgram.writeRef())
+            session->createCompositeComponentType(components.data(), components.size(), composedProgram.writeRef())
         );
+        SLANG_RETURN_ON_FAIL(composedProgram->link(linkedProgram.writeRef()));
     }
     else
     {
--- a/tests/test-imported-constant-buffer.cpp
+++ b/tests/test-imported-constant-buffer.cpp
@@ -3,7 +3,7 @@
 using namespace rhi;
 using namespace rhi::testing;
 
-GPU_TEST_CASE("imported-constant-buffer", ALL)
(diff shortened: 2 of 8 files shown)


src/shader.cpp
```diff
@@ -71,9 +71,11 @@ Result ShaderProgram::init()
             }
             components.push_back(m_desc.slangEntryPoints[i]);
         }
+        ComPtr<slang::IComponentType> composedProgram;
         SLANG_RETURN_ON_FAIL(
-            session->createCompositeComponentType(components.data(), components.size(), linkedProgram.writeRef())
+            session->createCompositeComponentType(components.data(), components.size(), composedProgram.writeRef())
         );
+        SLANG_RETURN_ON_FAIL(composedProgram->link(linkedProgram.writeRef()));
```
skallweitNV, line 78: I think this is generally OK, but we should probably omit the link step (and ideally the composition step as well) when there are no entry points specified at all. If only a `slangGlobalScope` is provided this can be used directly without calling `createCompositeComponentType` or `link`.

tests/test-imported-constant-buffer.cpp
```diff
@@ -0,0 +1,89 @@
...
+    desc.slangEntryPoints = entryPoints;
+    desc.slangEntryPointCount = 1;
+
+    ComPtr<IShaderProgram> shaderProgram;
+    REQUIRE_CALL(device->createShaderProgram(desc, shaderProgram.writeRef()));
+
+    ComputePipelineDesc pipelineDesc = {};
+    pipelineDesc.program = shaderProgram.get();
+    ComPtr<IComputePipeline> pipeline;
+    REQUIRE_CALL(device->createComputePipeline(pipelineDesc, pipeline.writeRef()));
```
skallweitNV: Don't we have testing helpers to reduce this boilerplate?

skallweitNV (commented):
Thanks for the PR!
Can we not change the file mode of `.claude/hooks/stop.py` or is there a reason for this?

src/shader.cpp
EvanLuo42: I tried the fast-path but it re-triggers the `varLayout` assertion that 3903c4a was fixing. The fast-path assumes `slangGlobalScope` is directly usable when no entry points are provided. But `loadProgram`  in `tests/testing.cpp` calls `createCompositeComponentType` without `link()` and passes that unlinked composite as `slangGlobalScope` (with no entry points). It relies on `ShaderProgram::init()` linking it internally. So I revert the fast-path for now.

skallweitNV: Oh interesting. I was just worried about extra cost of `link()`. We could fix the code in testing.cpp instead. But this feels like more of an API contract question at this point,

EvanLuo42: Shall i do the fix in this PR?

skallweitNV: Sorry for the delay. Yeah, we should probably fix the testing code so we don't have to call `createCompositeComponentType` and `link` if only `slangGlobalScope` is passed.

EvanLuo42: I have fixed the tesing code and also in `cpu-pipeline.cpp`

jkwak-work:
@EvanLuo42 , is this PR ready for another review?

EvanLuo42:
@jkwak-work Yes it is.

skallweitNV (approved):
LGTM

PR #702, opened 2026-04-10
Make block allocator safe for multiple static initialization

skiminki-nv:
The static initializers and deinitializers may be invoked multiple times for the same object when the code is in a DSO, which is opened in the RTLD_GLOBAL mode. This mode shares the symbols between different dlopen() instances. In practice, this means that constructors and destructors of static objects may be invoked multiple times.

Protect BlockAllocator<T> from multiple initialization and deinitialization by adding a wrapper with reference counting.

Issue https://github.com/shader-slang/slang/issues/10785

## Summary by CodeRabbit

* **Refactor**
  * Per-class allocators now use a process-wide reference-counted wrapper to manage initialization and teardown, with serialized construction/destruction and stronger lifetime checks.
* **Tests**
  * Block-allocator tests updated to access the underlying allocator instance, aligning test access with the new allocator wrapper behavior.

### DIFF
--- a/src/core/block-allocator.h
+++ b/src/core/block-allocator.h
@@ -246,7 +246,7 @@ public:
     }
 
     /// Return the underlying allocator
-    BlockAllocator<T> &getUnderlyingAllocator()
+    BlockAllocator<T>& getUnderlyingAllocator()
     {
         SLANG_RHI_ASSERT(s_refcount > 0);
         return *s_allocator;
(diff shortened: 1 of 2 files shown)


skallweitNV:
#710 should fix this issue (plus a lot of other ones) for this scenario.

skiminki-nv:
PR #710 makes this PR moot. I'll retract this PR after 710 has been merged.

skiminki-nv:
Retracting the PR. The underlying issue was fixed by #710.

PR #713, opened 2026-04-21
Metal: enable GPU pointer support with untracked hazard mode

jhelferty-nv:
# Metal Backend: Untracked Hazard Mode with MTLResidencySet

Part of implementing shader-slang/slang#10841.

## Why change the Metal backend?

The previous Metal backend used `MTLHazardTrackingModeTracked` (the default), where Metal automatically inserts barriers between command encoders that access the same resources. While convenient, this model has fundamental limitations for slang-rhi's GPU pointer support:

1. **Tracked mode cannot follow pointers.** Metal's automatic hazard tracking operates on resources that are directly bound to a pipeline via argument buffers or `useResource` calls. When a shader accesses a buffer through a GPU pointer (device address), Metal has no way to know that resource is in use. This means tracked mode provides zero synchronization benefit for pointer-accessed resources -- the very resources slang-rhi's pointer binding system is designed to support.

2. **Tracked mode has per-resource overhead.** Each `useResource` call that enrolls a resource in hazard tracking carries CPU overhead that scales with resource count. For workloads binding large numbers of resources through argument buffers and pointer indirection, this overhead adds up.

3. **Tracked mode conflates residency and synchronization.** The `useResource` API serves double duty: it both declares resource residency for a command encoder and enrolls the resource in hazard tracking. This coupling makes it impossible to manage residency efficiently (batched, at queue level) while opting out of per-resource tracking overhead.

## Why not just add useResources for pointer-accessed buffers?

The natural question is whether we could keep tracked mode and simply call `useResources` for buffers accessed via GPU pointers. This would avoid the untracked mode migration entirely. Unfortunately, the slang-rhi API makes this impractical.

Pointer values are passed to the shader object as raw `uint64_t` device addresses via `setData()`:

```cpp
shaderCursor["src"].setData(src->getDeviceAddress());
```

This is a plain `memcpy` of 8 bytes into the shader object's uniform data blob -- no `IBuffer*` reference is retained. But `useResources` requires `MTL::Resource*` pointers, not addresses. So to call `useResources` for pointer targets, we would need:

1. **A device-level address-to-buffer registry** to reverse-lookup `uint64_t -> MTL::Buffer*` (populated at buffer creation, cleared on destruction).
2. **A type-layout reflection walk** at bind time to identify which byte offsets in the uniform data contain pointer values (using Slang's type system to find pointer-kind fields and their offsets).
3. **Per-bind resolution logic** to extract addresses from those offsets, look them up in the registry, and add the resulting `MTL::Buffer*` to the `useResources` list.

The untracked mode approach avoids all of this. Because `memoryBarrier(scope: buffers | textures)` operates on **all buffer and texture memory** accessible to the encoder -- not on individually tracked resources -- it inherently covers pointer-accessed buffers without needing to know which buffers are in use. There is no address resolution, no type-layout walk, and no per-bind overhead.

## What replaces it?

The new backend uses three orthogonal mechanisms, each with a single well-defined responsibility:

### Residency: MTLResidencySet (macOS 15+)

All GPU-accessible allocations (buffers and textures) are registered in a single `MTLResidencySet` attached to the command queue. This replaces per-encoder `useResource` calls as the primary residency mechanism. `useResources` is still called for resources referenced inside argument buffers (parameter blocks), since Metal requires these declarations to properly resolve argument buffer resource references -- but residency itself is managed by the residency set, not by these per-encoder calls.

- Allocations are added at creation time via `registerAllocation()` and removed at destruction time via `unregisterAllocation()`.
- Changes are batched: `addAllocation`/`removeAllocation` stage changes, and a single `commit()` call in `CommandQueueImpl::submit()` applies them before any command buffer executes.
- The residency set is attached to the command queue, so every command buffer automatically inherits the full set of resident allocations.
- Access is protected by `m_residencySetMutex` since Apple documents `MTLResidencySet` methods as not thread-safe.

### Synchronization: Explicit fence and barriers

With tracked mode disabled, all GPU synchronization is the application's responsibility. The synchronization model is documented in detail in `metal-command.h` (above `CommandQueueImpl`), targeted at readers familiar with Vulkan/D3D12.

| Scope | Mechanism | When it fires |
|-------|-----------|---------------|
| Within a compute/render pass | `memoryBarrier(scope:)` | On pipeline rebind (`cmdSetComputeState`) after a prior dispatch |
| Between encoders (same or different command buffers) | `MTL::Fence` (`m_queueFence`) | `updateFence` at encoder end, `waitForFence` at next encoder start |
| CPU-side tracking | `MTL::SharedEvent` (`m_trackingEvent`) | Signaled on last CB of each `submit()`; used for deferred deletion and `waitOnHost()` |

A single `MTL::Fence` owned by `CommandQueueImpl` handles all GPU-side synchronization -- both within a command buffer (between encoder transitions) and across command buffer submissions on the same queue. Unlike `VkFence` (which is GPU->CPU only), `MTL::Fence` is GPU->GPU and provides both execution ordering and memory visibility for untracked resources. Its `waitForFence` resolves at execution time (not recording time), so recording order is irrelevant; only commit order matters.

`MTL::SharedEvent` is used solely for CPU-side concerns: deferred resource deletion and host-side waiting. It does not participate in GPU-side memory visibility.

This explicit model is directly analogous to how Vulkan and D3D12 handle synchronization, and it works correctly for pointer-accessed resources because the barriers operate on memory scopes (all buffers, all textures) rather than individually tracked resources.

### Storage: Private + Shared only

`MTLStorageModeManaged` is forbidden because its `didModifyRange`/`synchronizeResource` model creates implicit per-resource tracking that conflicts with the explicit synchronization strategy. On Apple Silicon's UMA architecture, `Shared` mode provides CPU/GPU coherent access without explicit cache management:

| RHI MemoryType | Metal Storage Mode | Rationale |
|----------------|--------------------|-----------
| `DeviceLocal` | `Private` | GPU-only, highest bandwidth |
| `Upload` | `Shared` + `WriteCombined` | UMA coherent, CPU write-optimized |
| `ReadBack` | `Shared` + `DefaultCache` | UMA coherent, CPU read-optimized |

## Requirements

- **macOS 15.0+** (runtime requirement for `MTLResidencySet` API)
- **Apple GPU Family 6+** (Apple Silicon; enforced at device initialization)
- **metal-cpp macOS15.2/iOS18.2 headers** (compile-time requirement for `MTLResidencySet` type definitions)

## Trade-offs

**What we gain:**
- Correct synchronization for GPU pointer workloads (the primary motivation)
- Lower per-frame CPU overhead by replacing per-encoder residency management with batched residency via `MTLResidencySet`
- Simpler mental model: synchronization is explicit and auditable, matching Vulkan/D3D12 patterns
- Batched residency management with minimal API call overhead

**What we give up:**
- Intel Mac support (MTLResidencySet requires Apple Silicon)
- macOS < 15 support
- Automatic protection against missing barriers (developer must insert barriers correctly; synchronization model documented in `metal-command.h`)

---

## Summary

- Fix Metal argument buffer layout and uniform data copy for structs containing pointer fields, enabling GPU pointer support on Metal.
- Migrate Metal backend to use `MTLResidencySet` for all resource residency management.
- Enforce `MTLHazardTrackingModeUntracked` on all resource allocations with explicit manual synchronization:
  - **Intra-encoder**: `memoryBarrier` for compute/render passes
  - **Inter-encoder and inter-command-buffer**: Single queue-owned `MTL::Fence` (`waitForFence`/`updateFence`) between all encoder transitions
  - **CPU-side tracking**: `MTL::SharedEvent` for deferred deletion and `waitOnHost()`
- Replace `MTLStorageModeManaged` with `MTLStorageModeShared` for CPU-accessible resources (UMA coherent on Apple Silicon).
- Require Apple GPU Family 6+ (Apple Silicon) at device initialization.
- Add new pointer binding tests: struct float copy, global barrier intermediate copy, intra-pass rebind, parameter blocks with pointers.
- Document the Metal synchronization model in `metal-command.h`, explaining the design for readers familiar with Vulkan/D3D12.
- Improve error handling and thread safety (residency set mutex).

## Test plan

- [x] All 933 `slang-rhi-tests` pass (0 failures, macOS/Metal)
- [x] All Metal-specific tests pass including new `bind-pointers-intra-pass-rebind`
- [x] slangpy `test_pointers.py` Metal results unchanged (18 pass, 3 known failures, 2 skipped)
- [x] clang-format-20 clean on all modified files
- [ ] CI passes on Linux (Vulkan/CUDA -- no Metal regressions expected)

## Summary by CodeRabbit

* **New Features**
  * Metal residency tracking and queue-scoped fence support for stronger GPU resource ordering
  * Device-pointer support for parameter blocks and pointer-bearing structs on Metal

* **Bug Fixes**
  * Improved command-buffer/encoder synchronization, error reporting, and memory-barrier behavior
  * Safer buffer/texture allocation, initialization, staging, and CPU-mapping semantics

* **Tests**
  * Added/expanded pointer-binding tests and new shader cases covering Metal

* **Chores**
  * Updated Metal C++ headers and include handling for builds


jhelferty-nv:
## slangpy Metal test status

The slangpy `test_pointers.py` suite was recently enabled for Metal (slangpy commit `0651eca5`, April 17) — previously Metal was excluded with a TODO referencing [shader-slang/slang#7605](https://github.com/shader-slang/slang/issues/7605). That commit added `DeviceType.metal` to `POINTER_DEVICE_TYPES`, enabling all 23 pointer tests on Metal at once.

**Results: 18 passed, 3 failed, 2 skipped** — identical between this PR and the baseline (no regressions).

### The 3 failures are all newly-enabled tests, not regressions

All 3 were known-broken at the time Metal was enabled. They fail due to Slang compiler limitations unrelated to this slang-rhi PR:

| Test | Error | Root cause |
|------|-------|------------|
| `test_copy_pointer` | MSL compile error: `device int *device *` cannot appear in `[[buffer]]` pointee type | Slang emits pointer-to-pointer fields in structs used as `constant*` kernel parameters. MSL forbids nested device pointers in buffer-backed structs. |
| `test_raw_buffer_of_pointers` | Same MSL compile error (`StructuredBuffer<int*>` → `device int *device *`) | Same root cause — `StructuredBuffer<int*>` produces nested device pointers in the CallData struct. |
| `test_atomic_float_in_struct_access` | Slang frontend error `E36107`: `operator&` unavailable for Metal target | `operator&` (address-of) is capability-gated to `cpp | cuda | spirv | llvm` in `core.meta.slang:2940` — Metal is excluded. This operator is being removed from Slang in the near future, so no fix is planned. |

### What needs to be done

- **`test_atomic_float_in_struct_access`**: Skip on Metal in slangpy. The `&` operator is going away — no Slang compiler fix needed.
- **`test_copy_pointer` / `test_raw_buffer_of_pointers`**: Requires a Slang compiler fix — either an IR legalization pass to lower `device T*` fields to `uint64_t` in buffer-backed structs for Metal, or changes to how slangpy constructs `CallData` structs. This is tracked as a known Metal pointer limitation.

None of these require changes in slang-rhi.

src/metal/metal-buffer.cpp
```diff
@@ -9,11 +9,16 @@ BufferImpl::BufferImpl(Device* device, const BufferDesc& desc)
...
+    if (auto* device = getDevice<DeviceImpl>())
+        device->unregisterAllocation(m_buffer.get());
+}
 
 void BufferImpl::deleteThis()
 {
-    getDevice<DeviceImpl>()->deferDelete(this);
+    if (auto* device = getDevice<DeviceImpl>())
+        device->deferDelete(this);
 }
```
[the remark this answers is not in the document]
jhelferty-nv: Good catch, but this is a pre-existing issue, not introduced by this PR. On `origin/main`, `deleteThis()` unconditionally dereferences `getDevice<DeviceImpl>()` without any null check:

```cpp
void BufferImpl::deleteThis()
{
    getDevice<DeviceImpl>()->deferDelete(this);
}
```

Our change added the null guard to prevent a crash on detached/shutdown paths, which turned a potential null-pointer dereference into a potential leak. The `else delete this;` fallback is a reasonable further improvement, but it's out of scope for this PR since we'd want to apply it consistently across all backends (the same pattern exists in Vulkan and D3D12). Filed as a follow-up.

skallweitNV:
@jhelferty-nv before I dive into this myself, here is one additional quick AI review. Maybe worth addressing the issues it found before I take thorough look.

## Review: Metal Backend — Manual Resource Tracking & Pointer Support

### Overall Assessment

The refactor is well-structured. The core design — switching to `MTLHazardTrackingModeUntracked` with explicit fences between encoders, a `ResidencySet` for allocation management, and `StorageModeShared` replacing `StorageModeManaged` — is sound and appropriate for Apple Silicon. The test coverage for pointer scenarios is excellent. That said, there are several issues worth addressing.

---

### Issues

**1. `hasPointerFields()` is computed but never used — pointer-target buffers aren't made resident via `useResource`**

`_hasPointerFields()` in metal-shader-object-layout.cpp correctly detects pointer fields, and the `m_hasPointerFields` flag is stored + exposed via an accessor, but **no code ever calls `hasPointerFields()`**. When a shader dereferences a `Ptr<T>`, the GPU address is written into the argument buffer as raw uniform data, but the buffer that address points to is never passed to `encoder->useResources(...)`.

This currently works because:
- All buffers are registered in the `ResidencySet` (via `registerAllocation`), and
- On Apple Silicon, residency set membership is sufficient for GPU access.

However, this relies on the Apple GPU's behavior where `useResource` is not strictly required for residency-set-registered resources. If Apple ever tightens the Metal validation layer or this code runs on a non-UMA path, it could break silently. Consider either:
- Removing `_hasPointerFields`/`m_hasPointerFields` entirely (dead code), or
- Actually wiring it up to register pointer-targeted buffers via `useResource`.

**2. Residency set not removed from command queue on shutdown**

The `DeviceImpl` destructor doesn't call `m_commandQueue->removeResidencySet(m_residencySet.get())` before releasing the command queue. The Metal spec says a residency set should be removed from the queue before being released. The `NS::SharedPtr` will drop the reference, but the queue may still reference the stale set during teardown. This is likely benign in practice, but could trigger Metal validation layer warnings.

**3. `m_lastCpuAccessMode` is dead state**

In metal-buffer.h, `m_lastCpuAccessMode` is still declared and written in `mapBuffer()`, but `unmapBuffer()` is now a no-op. This is vestigial code from the `StorageModeManaged` era. Safe to remove to avoid confusion.

**4. `m_computeEncoderHasDispatched` not set by indirect dispatch — latent bug**

The flag is only set in `cmdDispatchCompute`. When `cmdDispatchComputeIndirect` gets implemented (currently stubbed as `NOT_SUPPORTED`), it will also need to set this flag, or the barrier-on-rebind in `cmdSetComputeState` won't fire after an indirect dispatch. Worth adding a `// TODO` or setting the flag preemptively in the stub.

---

### Minor Observations

**5. `cmdGlobalBarrier` else-branch with no active encoder is a no-op.** If called when no encoder is open, `endCommandEncoder()` does nothing. This seems intentional but worth a comment — the caller might expect barrier semantics and get none.

**6. Staging resources are not registered with the residency set.** The diff comments explain this is safe because "all allocations are GPU-accessible regardless of residency set membership" on Apple Silicon GPUFamilyApple6+. This is correct today, but the assumption is repeated in 4+ places. If it ever changes, multiple sites need updating. Consider a named helper or a single-source comment.

**7. `makeResourceOptions` assert won't catch texture paths.** The `SLANG_RHI_ASSERT(storageMode != MTL::ResourceStorageModeManaged)` in metal-utils.h only guards buffer creation. Texture creation sets `setHazardTrackingMode`/`setStorageMode` directly on the descriptor and bypasses `makeResourceOptions`. If someone creates a texture with managed mode in the future, this assertion won't catch it.

**8. The `writeOrdinaryDataIntoArgumentBuffer` field-count mismatch assertion calls `std::abort()` in release builds** (via `SLANG_RHI_ASSERT_FAILURE`). This is deliberate (prevents data corruption), but worth noting — it's a hard crash, not a graceful error. The `return SLANG_FAIL` after it only runs if asserts are disabled via `ScopedDisableAssert`.

**9. Test `"bind-pointers-intra-pass-rebind"` excludes Vulkan/D3D12.** The test is marked `CUDA | Metal` only. The comment says Metal serializes via the barrier-on-rebind mechanism. The exclusion of Vulkan/D3D12 makes sense (no automatic inter-dispatch barriers), but it might be worth adding a `// Note:` explaining why, since the nosync test already covers the "expect race" case for those APIs.

---

### What Looks Good

- **Fence-based inter-encoder synchronization** is clean: `updateFence` on end, `waitForFence` on create, with the `m_needsFenceWait` flag preventing unnecessary waits.
- **`encodeWait` on the tracking event** in both `record()` and `encodeWaitForPreviousSubmission()` correctly handles cross-command-buffer ordering for untracked resources. The initial event value of 1 matching `m_lastSubmittedID` ensures the first wait completes immediately.
- **Residency set commit timing** in `submit()` is correct — committed before command buffers execute.
- **Error handling** is improved: staging buffer/command buffer/encoder creation now checks for null individually rather than batching into a single `if (!a || !b || !c)`.
- **Layout fix** in `setElementTypeLayout` using the `m_parameterBlockTypeLayout` for Metal's Tier 2 argument buffer layout is correct — prevents overlapping offsets between value and pointer fields.
- **Test coverage** is thorough: parameter blocks, mixed value+pointer fields, struct-embedded float pointers, inter-pass sync, intra-pass rebind, and the global barrier path.

src/metal/metal-command.cpp
```diff
@@ -111,14 +133,16 @@ class CommandRecorder
...
-    for (const auto& buffer : commandBuffer->m_bindingCache.buffers)
+    // Wait for all previous submissions to complete before executing this command buffer.
+    // Required for untracked resources: Metal does not automatically synchronize
+    // memory across command buffer boundaries for untracked resources.
+    // Must be encoded BEFORE any commands so the GPU waits before executing work.
     {
-        getBlitCommandEncoder()->synchronizeResource(buffer->m_buffer.get());
+        auto* queue = commandBuffer->m_queue;
+        m_commandBuffer->encodeWait(queue->m_trackingEvent.get(), queue->m_lastSubmittedID);
     }
```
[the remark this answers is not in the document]
jhelferty-nv: From looking at slang-rhi usage, the command queue should only ever be called from a single thread, so it doesn't look like this should be an issue.

Note: slang-rhi could really use some documentation explaining the usage and threading expectations/contract with the user. I'm having to assume what these are based on the existing slangpy code and test cases, as well as leaning heavily on the threading expectations of D3D/VK targets that would be violated if multithreaded use were to occur.

jhelferty-nv:
I think I've addressed the actionable items. It's interesting that some of these issues didn't show up with my own local review prompt. Agent-assisted response follows.

Thanks for the thorough review, Simon. Addressed all items in `feca5c3`:

**Issues:**

1. **`hasPointerFields()` dead code** -- Removed entirely (`_hasPointerFields`, `m_hasPointerFields`, accessor). Residency set covers pointer-target buffers; no per-encoder `useResource` gating needed.

2. **Residency set not removed on shutdown** -- Added `m_commandQueue->removeResidencySet()` in `~DeviceImpl()` after `shutdown()` (GPU is provably idle at that point).

3. **`m_lastCpuAccessMode` dead state** -- Removed field and write site. Vestigial from managed-mode era.

4. **`m_computeEncoderHasDispatched` + indirect dispatch** -- Added TODO comment in the `NOT_SUPPORTED` stub noting this must be set when implemented.

**Minor observations:**

5. **`cmdGlobalBarrier` no-encoder else-branch** -- Expanded comment to clarify: for blit/AS encoders it ends the encoder (fence handles ordering), for no encoder it's a no-op (vacuously satisfied).

6. **Staging residency assumption repeated** -- Centralized the rationale in `metal-utils.h` near `makeResourceOptions`. Individual sites now reference it with a one-liner.

7. **`makeResourceOptions` doesn't guard textures** -- Added `SLANG_RHI_ASSERT(storageMode != Managed)` in `createTexture()` after the storage mode switch, and a note in `makeResourceOptions` about the texture gap.

8. **`SLANG_RHI_ASSERT_FAILURE` hard crash** -- Added comment documenting the intent: field-count mismatch means the argument buffer layout has diverged from shader reflection, continuing would silently corrupt GPU memory.

9. **`bind-pointers-intra-pass-rebind` excludes Vulkan/D3D12** -- Corrected comment: Vulkan/D3D12 *do* emit barriers on rebind (via `requireBindingStates` + `commitBarriers`), but only for descriptor-bound resources. This test's buffers are accessed via raw device addresses which bypass the state tracker.

src/metal/metal-buffer.cpp
```diff
@@ -9,11 +9,16 @@ BufferImpl::BufferImpl(Device* device, const BufferDesc& desc)
 {
 }
 
-BufferImpl::~BufferImpl() {}
+BufferImpl::~BufferImpl()
+{
+    if (auto* device = getDevice<DeviceImpl>())
```
skallweitNV: This check is obsolete, the device lifetime is tied to the buffer.

src/metal/metal-buffer.cpp
```diff
@@ -9,11 +9,16 @@ BufferImpl::BufferImpl(Device* device, const BufferDesc& desc)
...
+        device->unregisterAllocation(m_buffer.get());
+}
 
 void BufferImpl::deleteThis()
 {
-    getDevice<DeviceImpl>()->deferDelete(this);
+    if (auto* device = getDevice<DeviceImpl>())
```
skallweitNV: Same, no need for the check IMO.

src/metal/metal-buffer.cpp
```diff
@@ -61,23 +71,42 @@ Result DeviceImpl::createBuffer(const BufferDesc& desc_, const void* initData, I
...
-            return SLANG_FAIL;
+            auto stagingOpts = makeResourceOptions(MTL::ResourceStorageModeShared);
+            NS::SharedPtr<MTL::Buffer> stagingBuffer =
+                NS::TransferPtr(m_device->newBuffer(initData, bufferSize, stagingOpts));
+            if (!stagingBuffer)
+                return SLANG_FAIL;
+            MTL::CommandBuffer* commandBuffer = m_commandQueue->commandBuffer();
+            if (!commandBuffer)
+                return SLANG_FAIL;
+            encodeWaitForPreviousSubmission(commandBuffer);
```
skallweitNV: Why do we have to wait for previous submission to finish? We're copying from a new staging buffer to a new buffer, there shouldn't be any dependency, no?

src/metal/metal-buffer.cpp
```diff
@@ -94,13 +123,12 @@ Result DeviceImpl::createBufferFromNativeHandle(NativeHandle handle, const Buffe
 Result DeviceImpl::mapBuffer(IBuffer* buffer, CpuAccessMode mode, void** outData)
 {
     BufferImpl* bufferImpl = checked_cast<BufferImpl*>(buffer);
-    bufferImpl->m_lastCpuAccessMode = mode;
     if (mode == CpuAccessMode::Read)
     {
         MTL::CommandBuffer* commandBuffer = m_commandQueue->commandBuffer();
```
skallweitNV: I don't think slang-rhi needs to synchronize here. In fact, mapBuffer probably becomes a no-op on Metal.

src/metal/metal-texture.cpp
```diff
@@ -12,6 +12,11 @@ TextureImpl::TextureImpl(Device* device, const TextureDesc& desc)
 TextureImpl::~TextureImpl()
 {
     m_defaultView.setNull();
+    if (m_texture && !m_isSwapchainTexture)
+    {
+        if (auto* device = getDevice<DeviceImpl>())
```
skallweitNV: Check not necessary.

src/metal/metal-texture.cpp
```diff
@@ -23,7 +28,8 @@ void TextureImpl::deleteThis()
     }
     m_defaultView.setNull();
     m_sampler.setNull();
-    getDevice<DeviceImpl>()->deferDelete(this);
+    if (auto* device = getDevice<DeviceImpl>())
```
skallweitNV: Check not necessary.

src/metal/metal-texture.cpp
```diff
@@ -151,23 +159,31 @@ Result DeviceImpl::createTexture(const TextureDesc& desc_, const SubresourceData
...
         textureDesc->setCpuCacheMode(MTL::CPUCacheModeDefaultCache);
+        textureDesc->setHazardTrackingMode(MTL::HazardTrackingModeUntracked);
         NS::SharedPtr<MTL::Texture> stagingTexture = NS::TransferPtr(m_device->newTexture(textureDesc.get()));
 
+        if (!stagingTexture)
+            return SLANG_FAIL;
         MTL::CommandBuffer* commandBuffer = m_commandQueue->commandBuffer();
+        if (!commandBuffer)
+            return SLANG_FAIL;
+        encodeWaitForPreviousSubmission(commandBuffer);
```
skallweitNV: Same question here, do we have any dependency to wait for?

src/metal/metal-utils.h
```diff
@@ -75,4 +75,27 @@ MTL::TriangleFillMode translateTriangleFillMode(FillMode mode);
...
+    MTL::ResourceOptions storageMode,
+    MTL::ResourceOptions cpuCacheMode = MTL::ResourceCPUCacheModeDefaultCache
+)
+{
+    SLANG_RHI_ASSERT(storageMode != MTL::ResourceStorageModeManaged);
+
+    return storageMode | cpuCacheMode | MTL::ResourceHazardTrackingModeUntracked;
+}
+
+// Staging resources (short-lived upload/readback buffers and textures) are
```
skallweitNV: This comment feels fairly out-of-place here.

src/metal/metal-utils.h
```diff
@@ -75,4 +75,27 @@ MTL::TriangleFillMode translateTriangleFillMode(FillMode mode);
...
 MTL::StoreAction translateStoreOp(StoreOp storeOp, bool resolve);
 
+/// Build MTL::ResourceOptions with enforced restrictions:
+/// - MTLStorageModeManaged is forbidden (R3)
+/// - MTLHazardTrackingModeTracked is forbidden (R1)
+/// All resources are created with untracked hazard tracking mode.
+/// Note: textures use MTL::TextureDescriptor setters directly and bypass
+/// this helper. A separate SLANG_RHI_ASSERT in createTexture() guards
+/// against managed-mode textures.
+inline MTL::ResourceOptions makeResourceOptions(
```
skallweitNV, line 87: Not sure about the usefulness of this helper TBH. But not too fussy.

tests/test-bind-pointers.cpp
```diff
@@ -139,15 +314,165 @@ GPU_TEST_CASE("bind-pointers-intermediate-copy-nosync", Vulkan | CUDA)
...
+        // The next compute pass's getComputeCommandEncoder() calls waitForFence(), ensuring
+        // writes from the first encoder are visible before the second encoder executes.
         compareComputeResult(device, dst, std::span<uint8_t>(data));
     }
     else
     {
-        // GFX APIs like Vulkan and D3D12 require explicit synchronization between dispatches, which
-        // isn't done automatically for pointers, so we'd expect race conditions
+        // GFX APIs like Vulkan and D3D12 require explicit synchronization between dispatches,
+        // which isn't done automatically for pointer-accessed resources, so we'd expect race
```
skallweitNV: Do not write tests for race-conditions. They are not deterministic by definition. We've had multiple of these tests in the past that we had to disable.

src/metal/metal-device.cpp
```diff
@@ -77,6 +81,30 @@ Result DeviceImpl::initialize(const DeviceDesc& desc, BackendImpl* backend)
     {
         return SLANG_FAIL;
     }
+
+    // R5: Require Apple Silicon (GPU Family 6+) for untracked mode + MTLResidencySet.
+    if (!m_device->supportsFamily(MTL::GPUFamilyApple6))
+    {
+        return SLANG_FAIL;
```
skallweitNV, line 93: We should probably also emit an error through the message callback here.

src/metal/metal-device.cpp
```diff
@@ -454,4 +486,35 @@ Result DeviceImpl::createQueryPool(const QueryPoolDesc& desc, IQueryPool** outPo
     return SLANG_OK;
 }
 
+void DeviceImpl::registerAllocation(MTL::Allocation* allocation)
+{
+    if (m_residencySet && allocation)
```
skallweitNV: This check feels dangerous. If the device was successfully created, `m_residencySet` must be available. If `allocation` is `null` then that's probably an error. So a `SLANG_RHI_ASSERT(allocation != nullptr)` should probably be enough here. Same below.

src/metal/metal-device.cpp
```diff
@@ -454,4 +486,35 @@ Result DeviceImpl::createQueryPool(const QueryPoolDesc& desc, IQueryPool** outPo
...
+    {
+        std::lock_guard<std::mutex> lock(m_residencySetMutex);
+        m_residencySet->addAllocation(allocation);
+        m_residencySetDirty = true;
+    }
+}
+
+void DeviceImpl::unregisterAllocation(MTL::Allocation* allocation)
+{
+    if (m_residencySet && allocation)
```
skallweitNV: Assert instead?

src/metal/metal-command.cpp
```diff
@@ -10,8 +10,26 @@
...
+
 namespace rhi::metal {
 
+static void addErrorHandler(MTL::CommandBuffer* commandBuffer)
+{
+    commandBuffer->addCompletedHandler(^(MTL::CommandBuffer* cb) {
+      if (cb->status() == MTL::CommandBufferStatusError)
+      {
+          NS::Error* error = cb->error();
+          std::fprintf(
```
skallweitNV, line 23: Would be nice to report that through the callback. But if this is a fatal error we can't recover from, this is fine.

src/metal/metal-command.cpp
```diff
@@ -930,6 +991,15 @@ MTL::RenderCommandEncoder* CommandRecorder::getRenderCommandEncoder(MTL::RenderP
     {
         endCommandEncoder();
         m_renderCommandEncoder = NS::RetainPtr(m_commandBuffer->renderCommandEncoder(renderPassDesc));
+        MTL::Fence* fence = m_commandBufferImpl ? m_commandBufferImpl->m_encoderFence.get() : nullptr;
+        if (fence && m_needsFenceWait)
+        {
+            m_renderCommandEncoder->waitForFence(
+                fence,
+                MTL::RenderStages(MTL::RenderStageVertex | MTL::RenderStageFragment)
```
skallweitNV: What about geometry shaders?

src/metal/metal-command.cpp
```diff
@@ -1153,9 +1262,21 @@ Result CommandQueueImpl::submit(const SubmitDesc& desc)
             FenceImpl* fence = checked_cast<FenceImpl*>(desc.waitFences[i]);
             commandBuffer->encodeWait(fence->m_event.get(), desc.waitFenceValues[i]);
         }
+        addErrorHandler(commandBuffer);
         commandBuffer->commit();
     }
 
+    // Commit any pending residency set changes.
+    if (auto* device = getDevice<DeviceImpl>())
```
skallweitNV: Unnecessary check.

skallweitNV (commented):
Overall looks good. A few comments maybe worth addressing.

fangjunzhou:
I've added a test case in slangpy for this update: https://github.com/shader-slang/slangpy/pull/954

src/metal/metal-buffer.cpp
```diff
@@ -94,13 +123,12 @@ Result DeviceImpl::createBufferFromNativeHandle(NativeHandle handle, const Buffe
...
-        MTL::BlitCommandEncoder* encoder = commandBuffer->blitCommandEncoder();
-        encoder->synchronizeResource(bufferImpl->m_buffer.get());
-        encoder->endEncoding();
+        if (!commandBuffer)
+            return SLANG_FAIL;
+        encodeWaitForPreviousSubmission(commandBuffer);
```
kaizhangNV: why there is a wait for previous submission?

src/metal/metal-buffer.cpp
```diff
@@ -111,16 +139,7 @@ Result DeviceImpl::mapBuffer(IBuffer* buffer, CpuAccessMode mode, void** outData
 Result DeviceImpl::unmapBuffer(IBuffer* buffer)
 {
     BufferImpl* bufferImpl = checked_cast<BufferImpl*>(buffer);
```
kaizhangNV, line 135: Can you just leave return OK in this function?

kaizhangNV (dismissed):
It overall looks good to me

src/metal/metal-buffer.cpp
jhelferty-nv: Removed.

src/metal/metal-buffer.cpp
jhelferty-nv: Removed.

src/metal/metal-buffer.cpp
jhelferty-nv: Good question. The `encodeWaitForPreviousSubmission` has been replaced with a single queue-owned `MTL::Fence` that all encoders participate in (`waitForFence` at start, `updateFence` at end). For this specific case you're right that there's no true data dependency -- both buffers are new. The `waitForFence` here completes immediately since no prior encoder has modified these resources. The `updateFence` at the end is the important part: it ensures subsequent encoders (from user command buffers) will see this init copy as complete before they access the buffer. The cost of the no-op wait is negligible compared to the blit itself.

src/metal/metal-buffer.cpp
jhelferty-nv: Agreed, and it already is -- `mapBuffer` just returns `m_buffer->contents()` with no synchronization. All Metal buffers that are mappable use `MTLStorageModeShared` on Apple Silicon UMA, so `contents()` gives a coherent CPU pointer directly.

src/metal/metal-texture.cpp
jhelferty-nv: Removed.

src/metal/metal-texture.cpp
jhelferty-nv: Removed.

src/metal/metal-texture.cpp
jhelferty-nv: Same situation as the buffer init -- the `waitForFence` completes immediately since these are new resources with no prior dependency. The `updateFence` ensures subsequent work sees the init data copy. Replaced `encodeWaitForPreviousSubmission` with the queue-owned `MTL::Fence` pattern.

src/metal/metal-utils.h
jhelferty-nv: Trimmed it down to a brief cross-reference to the synchronization model documentation in `metal-command.h`.

src/metal/metal-utils.h
jhelferty-nv: It centralizes the two invariants (no managed mode, always untracked) in one place so they can't be forgotten at individual call sites. Happy to inline if you prefer, but since you're not fussy I'll leave it.

tests/test-bind-pointers.cpp
jhelferty-nv: This test (`bind-pointers-intra-pass-rebind`) is deterministic, not a race condition test. It verifies that two dispatches within a single compute pass, with a rebind between them, correctly synchronize via `memoryBarrier`. The comment mentioning "nosync" was referencing a test that doesn't exist -- I've removed that reference. The test is excluded from Vulkan/D3D12 because their rebind barriers only cover descriptor-bound resources, not pointer-accessed ones.

src/metal/metal-device.cpp
jhelferty-nv: Already done -- the block calls `handleMessage(DebugMessageType::Error, DebugMessageSource::Driver, ...)` with the localized error description before returning `SLANG_FAIL`.

src/metal/metal-device.cpp
jhelferty-nv: Changed to `SLANG_RHI_ASSERT(allocation)`.

src/metal/metal-device.cpp
jhelferty-nv: Changed to `SLANG_RHI_ASSERT(allocation)`.

src/metal/metal-command.cpp
jhelferty-nv: Right, it's a fatal error -- `SLANG_RHI_ASSERT_FAILURE` fires immediately after. The `fprintf` to stderr ensures the error message is visible even if the assert handler doesn't surface it.

src/metal/metal-command.cpp
jhelferty-nv: Metal does not support geometry shaders. Mesh shaders are the replacement, and `drawMeshTasks` is already listed as `NOT_SUPPORTED` (Metal's mesh shader API uses a different dispatch model via `drawMeshThreadgroups` which we haven't implemented yet).

src/metal/metal-command.cpp
jhelferty-nv: Removed.

src/metal/metal-buffer.cpp
jhelferty-nv: The `encodeWaitForPreviousSubmission` call has been removed entirely. `mapBuffer` is now a no-op on Metal -- it just returns `m_buffer->contents()`. All mappable buffers use `MTLStorageModeShared` (UMA coherent), so no synchronization is needed.

src/metal/metal-buffer.cpp
jhelferty-nv: It's essentially that now. The `(void)bufferImpl` is just to suppress the unused variable warning from `checked_cast`. No synchronization or other work.

src/metal/metal-command.cpp
```diff
@@ -1153,9 +1262,21 @@ Result CommandQueueImpl::submit(const SubmitDesc& desc)
...
+    {
+        std::lock_guard<std::mutex> lock(device->m_residencySetMutex);
+        if (device->m_residencySet && device->m_residencySetDirty)
+        {
+            device->m_residencySet->commit();
+            device->m_residencySetDirty = false;
+        }
+    }
```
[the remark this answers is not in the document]
jhelferty-nv: The only `commit()` paths outside of `submit()` are the device-level init helpers (`createBuffer`, `createTexture`, `readBuffer`). These operate on either new resources (already added to the residency set via `registerAllocation`) or existing resources (already resident from a prior `submit`). The residency set is committed at the next `submit()` before any user command buffer executes. Adding a redundant `commit()` in each init helper would add mutex contention and API calls for no correctness benefit.

src/metal/metal-device.cpp
```diff
@@ -454,4 +486,35 @@ Result DeviceImpl::createQueryPool(const QueryPoolDesc& desc, IQueryPool** outPo
...
+void DeviceImpl::encodeWaitForPreviousSubmission(MTL::CommandBuffer* commandBuffer)
+{
+    if (m_queue && commandBuffer)
+    {
+        // On the first call (no prior submission), this wait completes immediately
+        // because CommandQueueImpl::init sets both m_lastSubmittedID and the
+        // tracking event's signaled value to 1.
+        commandBuffer->encodeWait(m_queue->m_trackingEvent.get(), m_queue->m_lastSubmittedID);
+    }
+}
```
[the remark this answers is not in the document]
jhelferty-nv: This matches the pattern used by both the D3D12 and Vulkan backends, which also use a plain `uint64_t` for `m_lastSubmittedID` with the mutex only protecting the deferred-delete queue push. The implicit contract across all backends is single-threaded queue access (or externally serialized). Making this atomic alone wouldn't be sufficient -- `submit()` is a multi-step sequence touching `m_commandBuffersInFlight`, event signaling, etc. Any thread-safety improvement here should be a cross-backend effort.

src/metal/metal-command.cpp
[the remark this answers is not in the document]
jhelferty-nv: This is no longer applicable. The `encodeWait`/`MTLSharedEvent` inter-command-buffer synchronization has been replaced entirely with a single queue-owned `MTL::Fence`. The fence's `waitForFence` resolves at GPU execution time (not recording time), so there is no stale ID problem -- recording order is irrelevant, only commit order matters.

src/metal/metal-buffer.cpp
```diff
@@ -9,7 +10,10 @@ BufferImpl::BufferImpl(Device* device, const BufferDesc& desc)
 {
 }
 
-BufferImpl::~BufferImpl() {}
+BufferImpl::~BufferImpl()
+{
+    getDevice<DeviceImpl>()->unregisterAllocation(m_buffer.get());
+}
```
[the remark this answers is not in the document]
jhelferty-nv: Good catch. Fixed -- added the null guard to match `TextureImpl::~TextureImpl`.

src/metal/metal-buffer.cpp
```diff
@@ -93,34 +124,16 @@ Result DeviceImpl::createBufferFromNativeHandle(NativeHandle handle, const Buffe
...
-    {
-        bufferImpl->m_buffer->didModifyRange(NS::Range(0, bufferImpl->m_desc.size));
-        MTL::CommandBuffer* commandBuffer = m_commandQueue->commandBuffer();
-        MTL::BlitCommandEncoder* encoder = commandBuffer->blitCommandEncoder();
-        encoder->synchronizeResource(bufferImpl->m_buffer.get());
-        encoder->endEncoding();
-        commandBuffer->commit();
-        commandBuffer->waitUntilCompleted();
-    }
+    (void)bufferImpl;
```
skallweitNV: We typically use the `SLANG_UNUSED` macro for this.

skallweitNV (dismissed):
LGTM

skallweitNV (approved):

PR #726, opened 2026-04-29
Add ExtraDescriptorBinding API for compiler-synthesized resources

jvepsalainen-nv:
## Summary

Add a generic mechanism for declaring and binding descriptor slots
that exist in the compiled shader binary but are not surfaced in
Slang's public reflection — for example, the coverage buffer
synthesized by Slang's `-trace-coverage` instrumentation feature.
Without this, slang-rhi's pipeline-layout builder walks reflection
alone and the resulting `VkPipelineLayout` rejects descriptor
references that the SPIR-V binary still issues, causing dispatch to
fail at pipeline creation.

## Motivation

The motivating use case is shader coverage instrumentation
(shader-slang/slang#10886, shader-slang/slang#10897). With that
feature enabled, the Slang compiler synthesizes a counter buffer
directly in the IR and surfaces its `(set, binding)` via
`slang::ICoverageTracingMetadata` — intentionally not via public
reflection, since coverage is a runtime-instrumentation concern
that should not pollute IDE / language-server / reflection-viewer
surfaces. Hosts on slang-rhi need a way to declare such slots so
that `vkCreatePipelineLayout` accepts the resulting layout.

The same pattern generalizes to any future compiler-synthesized
binding (debug-shadow buffers, profile probes, etc.); slang-rhi
gains no coverage-specific knowledge from this change.

## API additions

```cpp
namespace rhi {

// New struct.
struct ExtraDescriptorBinding {
    uint32_t set = 0;
    uint32_t binding = 0;
    slang::BindingType bindingType = slang::BindingType::Unknown;
};

// New optional fields on existing ShaderProgramDesc.
struct ShaderProgramDesc {
    // ... existing fields ...
    const ExtraDescriptorBinding* extraDescriptorBindings = nullptr;
    uint32_t extraDescriptorBindingCount = 0;
};

// New virtual on existing IShaderObject (added at end).
class IShaderObject : public ISlangUnknown {
    // ... existing methods ...
    virtual Result setExtraBinding(
        uint32_t set,
        uint32_t binding,
        const Binding& resource) = 0;
};

}
```

The host populates `extraDescriptorBindings` at program creation
based on whatever side-channel metadata interface it consults
(\`ICoverageTracingMetadata\` for the coverage case), and binds
resources to those slots via \`setExtraBinding\` at command-
recording time. Both APIs are additive; no existing slang-rhi
caller needs to change.

## Implementation scope

This PR implements the Vulkan path only:

- \`vk-shader-object-layout.cpp\` — \`Builder::addExtraDescriptorBinding\` merges the slot into the appropriate \`DescriptorSetInfo\` before \`vkCreateDescriptorSetLayout\`, so the pipeline layout includes the binding.
- \`vk-shader-object.cpp\` — \`bindAsRoot\` walks the root shader object's \`m_extraBindings\` after the reflection-driven binding loop and writes a descriptor for each, resolving \`VkDescriptorType\` from the layout's per-set binding list.
- \`vk-device.cpp\` — \`createShaderProgram\` plumbs the extras from desc → \`RootShaderObjectLayoutImpl::create\`.
- \`shader-object.{h,cpp}\` — base \`ShaderObject::setExtraBinding\` stores extras alongside reflection-keyed slots; a shared \`populateResourceSlot\` helper factors the per-binding-type validation logic.
- \`debug-layer/debug-shader-object.{h,cpp}\` — forwarding override.

D3D12, Metal, CUDA, CPU, WGPU follow the same shape and are scoped as follow-up commits/PRs.

## Verification

End-to-end verified via shader-slang/slang's \`examples/shader-coverage-demo\` on Vulkan (MoltenVK on macOS arm64). The demo:

1. Compiles a compute shader with \`-trace-coverage\` (counter ops + IR-time-synthesized coverage buffer).
2. Reads \`(set, binding)\` from \`slang::ICoverageTracingMetadata\`.
3. Sets \`ShaderProgramDesc::extraDescriptorBindings\` with the slot.
4. Calls \`setExtraBinding\` to bind the host's counter buffer.
5. Dispatches 8 frames × 64 particles.
6. Reads the buffer back; produces real per-line LCOV.

Sample LCOV output (verifies counters land in the right buffer):

\`\`\`
SF:examples/shader-coverage-demo/simulate.slang
DA:31,512        ← 64 × 8 = 512 hits (always-executed line) ✓
SF:examples/shader-coverage-demo/physics.slang
DA:45,0          ← unreachable error branch (dead-code signal) ✓
DA:46,0
\`\`\`

Without this PR the same dispatch fails at \`vkCreateComputePipelines\` with \"\`SPIR-V uses descriptor [Set 0, Binding 2, "__slang_coverage"] but the binding was not declared in the VkPipelineLayoutCreateInfo::pSetLayouts[0]\`\".

## Test plan

- [x] Vulkan dispatch via shader-coverage-demo verified locally on macOS / MoltenVK
- [ ] CI green
- [ ] Vulkan validation layer messages reviewed (pre-existing pipeline-barrier warnings unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -267,6 +267,25 @@ enum class LinkingStyle
     SeparateEntryPointCompilation
 };
 
+// Describes one descriptor binding that is present in the compiled
+// shader binary but not surfaced in Slang's public reflection. The
+// host supplies these via `ShaderProgramDesc::extraDescriptorBindings`
+// so slang-rhi includes the slot in pipeline layouts.
+//
+// The canonical motivation is compiler-synthesized resources whose
+// existence is reported through a side-channel metadata interface
+// (e.g. `slang::ICoverageTracingMetadata` for `-trace-coverage`).
+//
+// `bindingType` uses `slang::BindingType` so callers can pass through
+// values returned by Slang's reflection / metadata APIs without an
+// extra translation step.
+struct ExtraDescriptorBinding
+{
+    uint32_t set = 0;
+    uint32_t binding = 0;
+    slang::BindingType bindingType = slang::BindingType::Unknown;
+};
+
 struct ShaderProgramDesc
 {
     StructType type = StructType::ShaderProgramDesc;
@@ -287,6 +306,16 @@ struct ShaderProgramDesc
     // If set to 0, then `slangGlobalScope` must contain Slang EntryPoint components.
     // If not 0, then `slangGlobalScope` must not contain any EntryPoint components.
     uint32_t slangEntryPointCount = 0;
+
+    // Bindings that are present in the compiled shader binary but not
+    // in Slang's reflection (e.g. compiler-synthesized resources).
+    // slang-rhi will include these in the constructed pipeline
+    // layout so that subsequent `setBinding` calls can target them.
+    // The host obtains the (set, binding, bindingType) values from
+    // the relevant metadata interface (e.g.
+    // `slang::ICoverageTracingMetadata`).
+    const ExtraDescriptorBinding* extraDescriptorBindings = nullptr;
+    uint32_t extraDescriptorBindingCount = 0;
 
     const char* label = nullptr;
 };
@@ -1745,6 +1774,20 @@ public:
     virtual SLANG_NO_THROW Result SLANG_MCALL getObject(const ShaderOffset& offset, IShaderObject** outObject) = 0;
     virtual SLANG_NO_THROW Result SLANG_MCALL setObject(const ShaderOffset& offset, IShaderObject* object) = 0;
     virtual SLANG_NO_THROW Result SLANG_MCALL setBinding(const ShaderOffset& offset, const Binding& binding) = 0;
+
+    // Bind a resource at a raw `(set, binding)` slot declared via
+    // `ShaderProgramDesc::extraDescriptorBindings`. Used for
+    // resources that exist in the compiled shader binary but are not
+    // surfaced in Slang's reflection (e.g. compiler-synthesized
+    // bindings from `-trace-coverage`). The slot must have been
+    // declared on the program — otherwise the call returns
+    // `SLANG_E_INVALID_ARG`.
+    virtual SLANG_NO_THROW Result SLANG_MCALL setExtraBinding(
+        uint32_t set,
+        uint32_t binding,
+        const Binding& resource
+    ) = 0;
+
     virtual SLANG_NO_THROW Result SLANG_MCALL setDescriptorHandle(
         const ShaderOffset& offset,
         const DescriptorHandle& handle
--- a/src/debug-layer/debug-shader-object.cpp
+++ b/src/debug-layer/debug-shader-object.cpp
@@ -158,6 +158,13 @@ Result DebugShaderObject::setBinding(const ShaderOffset& offset, const Binding& 
     return baseObject->setBinding(offset, binding);
 }
 
+Result DebugShaderObject::setExtraBinding(uint32_t set, uint32_t binding, const Binding& resource)
+{
+    SLANG_RHI_DEBUG_API(IShaderObject, setExtraBinding);
+    SLANG_RETURN_ON_FAIL(checkNotFinalized());
+    return baseObject->setExtraBinding(set, binding, resource);
+}
+
 Result DebugShaderObject::setDescriptorHandle(const ShaderOffset& offset, const DescriptorHandle& handle)
 {
     SLANG_RHI_DEBUG_API(IShaderObject, setDescriptorHandle);
--- a/src/debug-layer/debug-shader-object.h
+++ b/src/debug-layer/debug-shader-object.h
@@ -57,6 +57,11 @@ public:
     virtual SLANG_NO_THROW Result SLANG_MCALL getObject(const ShaderOffset& offset, IShaderObject** outObject) override;
     virtual SLANG_NO_THROW Result SLANG_MCALL setObject(const ShaderOffset& offset, IShaderObject* object) override;
     virtual SLANG_NO_THROW Result SLANG_MCALL setBinding(const ShaderOffset& offset, const Binding& binding) override;
+    virtual SLANG_NO_THROW Result SLANG_MCALL setExtraBinding(
+        uint32_t set,
+        uint32_t binding,
+        const Binding& resource
+    ) override;
     virtual SLANG_NO_THROW Result SLANG_MCALL setDescriptorHandle(
         const ShaderOffset& offset,
         const DescriptorHandle& handle
--- a/src/shader-object.cpp
+++ b/src/shader-object.cpp
@@ -377,6 +377,132 @@ Result ShaderObject::setBinding(const ShaderOffset& offset, const Binding& bindi
     return SLANG_OK;
 }
 
+namespace {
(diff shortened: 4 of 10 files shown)


skallweitNV:
Quick AI review. I think the point on having these bindings properly exposed through Slang reflection might be a much better approach.



### Does it make sense?

**The motivation is sound.** Slang's coverage instrumentation (`-trace-coverage`) synthesizes a buffer binding at the IR level that intentionally doesn't appear in public reflection. Without a mechanism to declare these "hidden" bindings, `vkCreatePipelineLayout` rightfully rejects the SPIR-V because it references undeclared descriptors. The PR solves a real problem.

**However, there are design concerns:**

### 1. API Surface Uses `(set, binding)` — This is Vulkan-Centric

The `ExtraDescriptorBinding` struct exposes raw `(set, binding)` pairs, which map 1:1 to Vulkan's descriptor set/binding model. This abstraction leaks through the RHI boundary:

- **D3D12** uses root signatures with descriptor tables, root descriptors, and root constants — organized by `register` and `space`, not `(set, binding)`. The mapping isn't trivial; D3D12 doesn't even have a concept of "descriptor sets" in the same way.
- **Metal** uses argument buffers with indices per buffer/texture/sampler type.
- **WebGPU** uses bind groups and bindings, which are *similar* to Vulkan sets/bindings but not identical in semantics.
- **CUDA/CPU** don't have descriptor sets at all — they use direct memory pointers.

The PR claims other backends "follow the same shape" as follow-ups, but the `(set, binding)` model fundamentally doesn't map cleanly to D3D12's root signature mechanism. For D3D12, you'd need `(register, space, registerType)` instead.

### 2. `setExtraBinding` Bypasses the Existing Hook Mechanism

The existing `setBinding()` invokes `m_setBindingHook` after populating a slot — this is the mechanism CUDA and CPU backends use to perform side effects at bind time. The new `setExtraBinding()` does NOT invoke this hook. This means:
- CUDA/CPU backends would silently ignore extra bindings unless they add custom handling in their `bindAsRoot` equivalent.
- The patterns diverge: normal bindings use hooks, extra bindings don't.

### 3. Linear Search in `setExtraBinding` and `bindAsRoot`

```cpp
for (auto& entry : m_extraBindings) {
    if (entry.set == set && entry.binding == binding) { ... }
}
```

And in the Vulkan binding path:
```cpp
for (const auto& vkBinding : descriptorSetInfo.vkBindings) {
    if (vkBinding.binding == extra.binding) { ... }
}
```

These are O(n) linear scans. For the motivating use case (1 coverage buffer), this is fine. But the PR explicitly mentions generalizing to "debug-shadow buffers, profile probes, etc." — if many extras accumulate, this becomes quadratic per dispatch.

### 4. No Validation That the Slot Was Declared

`setExtraBinding()` in `ShaderObject` will happily store bindings to *any* `(set, binding)` pair. The validation only happens at bind time in Vulkan's `bindAsRoot` where undeclared sets are silently `continue`d. The PR description says "The slot must have been declared on the program — otherwise the call returns `SLANG_E_INVALID_ARG`" but the code silently drops unknown slots rather than returning an error.

### 5. `populateResourceSlot` Duplicates `setBinding` Logic

The PR extracts 126 lines of binding-type switching into a `populateResourceSlot` helper in an anonymous namespace — but doesn't refactor `setBinding` to use it. This creates two diverging copies of the same logic. When a new `BindingType` is added, both paths need updating.

### 6. Vulkan-Only Implementation Ships; Only `Buffer` Case Handled in `bindAsRoot`

The `bindAsRoot` code has:
```cpp
default:
    // Other resource kinds (textures, samplers, AS) can
    // be added on demand.
    break;
```

Combined with `populateResourceSlot` handling all types, this creates a confusing mismatch — you can *set* a texture as an extra binding (no error), but it will be silently ignored at dispatch time. This violates the principle of least surprise.

---

### Alternative Approaches

1. **Extend Slang reflection instead.** If the coverage buffer's `(set, binding)` is known to the compiler via `ICoverageTracingMetadata`, Slang could expose it through the *existing* reflection API (even if marked as "synthetic" or "internal"). This would require zero RHI changes — the existing pipeline layout builder would pick it up automatically.

2. **Use an opaque `PipelineLayoutExtension` mechanism.** Instead of baking `(set, binding)` semantics into the public API, allow backends to accept opaque extension data on `ShaderProgramDesc`. Each backend would interpret its own extension type (e.g., VulkanExtraBindings, D3D12ExtraRegisters). This is more verbose but avoids the Vulkan-centric abstraction leak.

3. **Use the existing `setBinding` path with a synthetic reflection range.** The RHI could internally create a synthetic binding range when processing `extraDescriptorBindings`, assigning them `bindingRangeIndex` values at the end of the reflection-derived ranges. This would unify the binding path entirely — no new virtual method needed.

---

### Will it Rot?

**High rot risk:**
- The `populateResourceSlot` helper duplicates `setBinding` and will diverge over time.
- Only the Vulkan path is implemented; other backends will either need follow-up PRs (that may never land) or will silently ignore the API.
- Adding `setExtraBinding` to `IShaderObject` (a COM interface) is a permanent commitment. If Slang reflection is later extended to expose synthesized bindings normally, this API becomes dead weight that can't be removed without a breaking change.
- The `vkBinding.stageFlags = VK_SHADER_STAGE_ALL` is overly broad — it works but wastes descriptor pool resources on stages that don't use the binding. It also means validation layers won't catch binding-to-wrong-stage errors.

---

### Additional Overhead for Rarely-Used API

The overhead is minimal for the common case (no extra bindings):
- `ShaderObject` gains a `short_vector<ExtraBindingSlot> m_extraBindings` member — this is heap-allocated only if used (short_vector inline storage).
- `ShaderProgramDesc` gains two new fields (pointer + count) defaulting to nullptr/0 — negligible.
- The `bindAsRoot` loop iterates `m_extraBindings` which is empty in the common case — effectively free.
- The layout builder's `for` loop over `extraDescriptorBindingCount` is also zero-cost when count is 0.

**Verdict: no meaningful runtime cost when unused.** Memory cost is one extra `short_vector` (likely 24 bytes inline) per shader object.

---

### Summary

| Aspect | Assessment |
|--------|-----------|
| Solves a real problem | Yes |
| API design | Vulkan-centric; `(set, binding)` doesn't map to D3D12/Metal cleanly |
| Cross-backend viability | Questionable without significant backend-specific re-interpretation |
| Code duplication | `populateResourceSlot` duplicates `setBinding` without refactoring either |
| Silent failures | Extra bindings to textures/samplers silently dropped; undeclared slots silently ignored |
| Rot risk | High — COM interface extension, duplicated code paths, partial implementation |
| Performance overhead | Negligible when unused |
| Alternative approaches | Extending Slang reflection or using synthetic binding ranges would be cleaner |

**Recommendation:** The motivating use case is valid, but the API would benefit from being designed around Slang's own abstraction layer (register/space or binding range indices) rather than raw Vulkan `(set, binding)` pairs. The cleanest long-term solution would be to have Slang expose these bindings via reflection (even with a "synthetic" flag) so the existing pipeline layout machinery handles them automatically — zero new API surface needed.

jvepsalainen-nv:
Closing as debug and pipecleaning only. Actual implementation continues in https://github.com/shader-slang/slang-rhi/pull/739

PR #731, opened 2026-04-30
Use emscripten 5.0.6

skallweitNV:
Unfortunately the emscripten build depends on a specific range of versions to work (due to webgpu dependency). We manually set the version to `5.0.6` instead of `latest` until this can be fixed in a different way.


skallweitNV:
@j8asic for visibility

ccummingsNV (approved):

PR #733, opened 2026-05-01
Add D3D12 work-graph support

jkwak-work:
This PR is to support D3D12 WorkGraph that is available with SM 6.8.
This PR follows the pattern of the code change when RayTracingPipelineDesc was added.

Add IWorkGraphPipeline, IWorkGraphPassEncoder, and IDevice::createWorkGraphPipeline to the public API, with a full D3D12 implementation and no-op stubs for all other backends and the debug layer.

## New Features

* Added Work Graph pipeline support, enabling advanced compute execution patterns on compatible hardware
* Extended the API with pipeline creation for work graphs and new command encoding operations for work graph pass management
* Implemented full Work Graph support for Direct3D 12 backend including state management and graph dispatch


skallweitNV:
@jkwak-work do you plan to work on this PR or should we close it?

PR #739, opened 2026-05-07
Add synthetic resource bindings for Vulkan and CUDA

jvepsalainen-nv:
## Summary

This PR adds an opt-in synthetic-resource binding path to `slang-rhi` for compiler-synthesized resources that are not part of the normal reflected shader interface.

Related to Slang issue https://github.com/shader-slang/slang/issues/10794.

Depends on Slang release v2026.9.1.

The first consumer is shader coverage, where Slang exposes a hidden writable counter buffer through metadata and `slang-rhi` needs to merge that resource into program layout and runtime binding without adding a second backend-specific binding model.

## API shape

The optional API is declared in `<slang-rhi/synthetic-bindings.h>` and documented in `docs/synthetic-bindings.md`.

- `ShaderProgramSyntheticResourcesDesc`
- `SyntheticResourceBindingDesc`
- `SyntheticBindingLocation`
- `ISyntheticShaderProgram`
- `bindSyntheticResource(...)`

Runtime binding still goes through the existing `ShaderOffset` / `IShaderObject::setBinding()` path. Synthetic resources are merged into the program's internal layout model, and ordinary programs do not expose `ISyntheticShaderProgram`.

## Implementation

- The main `slang-rhi.h` header only keeps the `StructType` enum value needed for `ShaderProgramDesc.next`; optional synthetic binding declarations live in the separate header.
- `ShaderProgram` stores synthetic-resource data behind a nullable state object, so ordinary programs do not carry synthetic-resource vectors or owned strings.
- Common parsing, validation, descriptor copying, and resolved-location storage live in `src/synthetic-resource-bindings.*`.
- Vulkan merges synthetic resources into the root descriptor layout and preserves explicit Vulkan set numbers only when synthetic resources are present.
- CUDA merges synthetic resources into the root layout using the existing uniform-offset marshaling path.
- Unsupported backends keep the normal path unchanged when no synthetic descriptor is provided and return `SLANG_E_NOT_IMPLEMENTED` for non-empty synthetic descriptors.
- The Emscripten package still links the CMark archive shipped with the Slang release because the static Slang compiler archive references it.

## Tests

Added focused synthetic-resource tests covering:

- opt-in interface behavior for ordinary programs
- unsupported-backend rejection behavior
- invalid descriptor validation
- `ISyntheticShaderProgram` location queries
- `bindSyntheticResource(...)`
- end-to-end Slang metadata consumption on Vulkan
- CUDA layout and binding-path coverage

## Intended Flow

1. Slang exposes hidden-resource metadata for the compiled shader.
2. The host translates that metadata into `SyntheticResourceBindingDesc` records.
3. The host passes those records into `createShaderProgram()` via `ShaderProgramDesc.next`.
4. `slang-rhi` merges the synthetic resources into backend layout state only on supported backends.
5. The host binds by synthetic resource id using `bindSyntheticResource(...)`, or queries `ISyntheticShaderProgram` and calls `setBinding()` directly.

## Backend Coverage

Implemented in this PR:

- Vulkan
- CUDA

Follow-up backend work can add D3D12, Metal, and CPU support without changing the public API shape.


tests/test-synthetic-resource-bindings.cpp
```diff
@@ -0,0 +1,381 @@
...
+    case slang::SyntheticResourceAccess::Read:
+        return SyntheticResourceAccess::Read;
+    case slang::SyntheticResourceAccess::Write:
+        return SyntheticResourceAccess::Write;
+    case slang::SyntheticResourceAccess::ReadWrite:
+        return SyntheticResourceAccess::ReadWrite;
+    default:
+        SLANG_RHI_ASSERT_FAILURE("Unhandled synthetic resource access");
+        return SyntheticResourceAccess::Read;
+    }
```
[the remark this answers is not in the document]
jvepsalainen-nv: This integration test intentionally depends on the Slang-side synthetic resource metadata/query API from slang#11080. That companion PR is planned to land first, so I kept the real end-to-end metadata path here instead of adding compatibility shims for the older pinned Slang snapshot.

tests/test-synthetic-resource-bindings.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (device->getDeviceType() == DeviceType::CUDA)
+    {
+        CHECK_EQ(location.offset.uniformOffset, 16u);
+    }
+
+    ComPtr<IShaderObject> rootObject;
+    REQUIRE_CALL(device->createRootShaderObject(shaderProgram.get(), rootObject.writeRef()));
+
+    auto buffer = createTestBuffer(device);
+    REQUIRE_CALL(bindSyntheticResource(shaderProgram.get(), rootObject.get(), kSyntheticResourceID, Binding(buffer)));
```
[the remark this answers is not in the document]
jvepsalainen-nv: Added an end-to-end Vulkan coverage case here. It now binds a normal output buffer plus the synthetic coverage buffer, dispatches once, verifies `outBuffer[0] == 10`, and reads the synthetic counter buffer back to confirm that the coverage instrumentation actually consumed the hidden binding. I kept the smaller direct synthetic-binding test as an API wiring check for the non-metadata path.

jvepsalainen-nv:
Addressed the outside-diff ABI comment in 00389bd: `StructType::ShaderProgramSyntheticResourcesDesc` is now appended after the existing enum values instead of inserted after `ShaderProgramDesc`, so existing public `StructType` numeric values stay stable.

jvepsalainen-nv:
Candidates to split into separate PRs. There are required for shader coverage feature to work but they are not directly related to synthetic resource bindings:

  - 1c07dfe Fix Vulkan bring-up on macOS: separate Vulkan platform bring-up PR. It is useful for running on this Mac, but it is not conceptually part of synthetic resource binding.
  - a974fc4 Fix Vulkan barrier stages for unsupported features: separate Vulkan correctness PR. It fixes barrier stage masks based on supported device features, unrelated to the synthetic binding API.
  - Part of 35028aa: the vk-api.cpp / vk-backend.cpp cleanup belongs with 1c07dfe if we split macOS Vulkan bring-up. The rest of 35028aa is synthetic-resource validation and should stay.

src/vulkan/vk-api.cpp
```diff
@@ -112,6 +132,24 @@ Result VulkanApi::initGlobalProcs(const VulkanModule& module)
...
+{
+    SLANG_RHI_ASSERT(instance && vkGetInstanceProcAddr != nullptr);
+
+    vkEnumeratePhysicalDevices =
+        (PFN_vkEnumeratePhysicalDevices)vkGetInstanceProcAddr(instance, "vkEnumeratePhysicalDevices");
+    vkGetPhysicalDeviceProperties2 =
+        (PFN_vkGetPhysicalDeviceProperties2)vkGetInstanceProcAddr(instance, "vkGetPhysicalDeviceProperties2");
+
+    if (!(vkEnumeratePhysicalDevices && vkGetPhysicalDeviceProperties2))
+    {
```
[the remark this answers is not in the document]
jvepsalainen-nv: Addressed in 44a0138. Vulkan enumeration proc init now falls back to `vkGetPhysicalDeviceProperties2KHR`, and instance proc init falls back to both `vkGetPhysicalDeviceProperties2KHR` and `vkGetPhysicalDeviceFeatures2KHR` when the core names are unavailable.

src/shader.cpp
```diff
@@ -204,6 +208,159 @@ const ShaderProgramDesc& ShaderProgram::getDesc()
...
+        return SLANG_E_INVALID_ARG;
+    if (record.arraySize == 0)
+        return SLANG_E_INVALID_ARG;
+    if (record.scope == SyntheticResourceScope::EntryPoint)
+    {
+        if (record.entryPointIndex < 0)
+            return SLANG_E_INVALID_ARG;
+    }
+    return SLANG_OK;
+}
```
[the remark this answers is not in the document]
jvepsalainen-nv: Addressed in 44a0138. `_validateSyntheticResourceRecord` now rejects global-scope records with `entryPointIndex != -1` and rejects `uniformStride != 0` when `uniformOffset == -1`.

src/vulkan/vk-shader-object-layout.cpp
```diff
@@ -974,4 +1046,81 @@ void RootShaderObjectLayoutImpl::Builder::addEntryPoint(EntryPointLayout* entryP
...
+        auto descriptorSetIndex = findOrAddDescriptorSet((uint32_t)resource.space);
+        if (SLANG_FAILED(m_result))
+            return m_result;
+
+        VkDescriptorSetLayoutBinding vkBindingRangeDesc = {};
+        vkBindingRangeDesc.binding = (uint32_t)resource.binding;
+        vkBindingRangeDesc.descriptorCount = resource.arraySize;
+        vkBindingRangeDesc.descriptorType = _mapDescriptorType(resource.bindingType);
+        vkBindingRangeDesc.stageFlags = VK_SHADER_STAGE_ALL;
+        if (!addDescriptorSetBinding(
```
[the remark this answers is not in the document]
jvepsalainen-nv: Addressed in 44a0138. Vulkan synthetic resource layout creation now rejects unmapped descriptor types before adding the descriptor set binding, so unsupported binding types fail with `SLANG_E_INVALID_ARG` earlier.

jvepsalainen-nv:
Change back to draft. We need a new Slang release with the synthetic metadata API.

tests/test-synthetic-resource-bindings.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    buffer.setNull();
+    rootObject.setNull();
+    syntheticProgram.setNull();
+    shaderProgram.setNull();
+}
```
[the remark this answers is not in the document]
jvepsalainen-nv: CUDA runtime consumption is now covered by `synthetic-resource-bindings-from-slang-metadata`, which is declared `Vulkan | CUDA` and dispatches/readbacks both the output buffer and coverage counter buffer. I am intentionally not adding EntryPoint-scope runtime coverage in this PR because both synthetic-resource backends currently reject non-Global scopes while building layout (`SLANG_E_NOT_IMPLEMENTED`) before any binding location exists. Exercising the `bindSyntheticResource()` entry-point branch requires backend support for entry-point-scoped synthetic resources, which is separate from consuming the Slang coverage metadata this PR needs.

jvepsalainen-nv:
CI is passing using Slang v2026.9

CMakeLists.txt
```diff
@@ -158,6 +158,11 @@ if(NOT DEFINED SLANG_RHI_DXC_URL)
     set(SLANG_RHI_DXC_URL "https://github.com/microsoft/DirectXShaderCompiler/releases/download/v1.9.2602/dxc_2026_02_20.zip")
 endif()
 
+# Swift CMark is a Slang compiler dependency in static wasm packages.
```
jvepsalainen-nv: This needs to be fixed in Slang release packaging. Slang has added n libcmark-gfm.a dependency in 2026.5.

skallweitNV:
@jvepsalainen-nv splitting off the parts of this PR that can go in standalone before the synthetic resources would be very welcome as it brings down the scope of the main PR. I wonder if "synthetic" is the correct term, but I might have a better idea once I dug through the code a bit.

skallweitNV:
Here is an initial (AI generated) review. I will take another look tomorrow, but the clean approach of having Slang expose synthetic bindings through reflection data is obviously not feasible. I like the idea of using extended descriptors, but I'm also a bit afraid of some of the complexity this currently adds.


### Summary

This PR introduces a mechanism for the host application to inject "synthetic" (compiler-generated, non-reflected) resource bindings into a `ShaderProgram`'s internal layout. The first use case is shader coverage tracing, where the Slang compiler emits a hidden RWBuffer for hit counters that isn't part of normal shader reflection.

---

### 1. Architectural Impact Assessment

**Scope of intrusion:**

| Layer | Impact |
|-------|--------|
| Public API (`slang-rhi.h`) | +174 lines: new enums, structs, interface, free function |
| `ShaderProgram` base class | Now multiply-inherits `ISyntheticShaderProgram`, stores synthetic records/locations |
| `ShaderObjectLayout` base | New virtual `getMinUniformBufferSize()` |
| `ShaderObject` base (`init()`) | Consults new virtual for buffer sizing |
| Vulkan `ShaderObjectLayoutImpl::Builder` | Refactored from map-based to direct-indexed descriptor sets; new `addDescriptorSetBinding()` with collision detection; `m_result` error propagation threaded throughout |
| Vulkan `RootShaderObjectLayoutImpl` | Now receives `ShaderProgram*`, calls `addSyntheticResources()` |
| CUDA `RootShaderObjectLayoutImpl` | Factory method, new `_addSyntheticResources()`, `m_minUniformBufferSize` |
| `vk-utils.cpp` | `calcPipelineStageFlags` signature changed (now takes `const VulkanApi&`), all call sites updated |
| `vk-backend.cpp` / `vk-api.cpp` | macOS Vulkan bring-up, enumeration proc refactor (unrelated to synthetic resources) |

The PR modifies **core architectural layers** (`ShaderProgram`, `ShaderObjectLayout`, `ShaderObject`) — these are classes that every backend inherits from or interacts with.

---

### 2. Key Concerns

#### 2.1 Layout class now aware of `ShaderProgram` (circular dependency smell)

Previously, the layout construction was a pure function of Slang reflection data. Now `RootShaderObjectLayoutImpl::Builder` holds a `ShaderProgram*` and writes back to it (`addResolvedSyntheticBindingLocation`). This creates a bidirectional coupling: `ShaderProgram` owns the layout, and the layout builder writes back into `ShaderProgram`.

This isn't necessarily wrong, but it breaks the clean unidirectional data flow (`desc → layout → object`) that existed before. Future developers need to reason about when `m_syntheticBindingLocations` is populated relative to other initialization.

#### 2.2 Vulkan descriptor set model change (map → direct-index)

The old model used `std::map<uint32_t, uint32_t>` allowing arbitrary space indices to map to sequential vector positions. The new model uses `space` as a direct index into the vector, creating potentially empty descriptor sets for intermediate spaces. This is a **semantic behavior change** even for programs with zero synthetic resources:

- If Slang emits spaces `{0, 3}`, the old code created a 2-element vector (`[space0, space3]`). The new code creates a 4-element vector with empty sets at indices 1 and 2.
- Those empty descriptor sets will get `VkDescriptorSetLayout` objects created (with zero bindings), and will flow into `vkCreatePipelineLayout`.

While Vulkan technically allows empty descriptor set layouts, this changes the **descriptor set indexing** that `vkCmdBindDescriptorSets` expects. The `addAllDescriptorSetsRec()` traversal now needs to match this new layout. The TODO comment about contiguous sets was removed — suggesting this is intentional — but the compatibility implications for existing shaders with non-contiguous spaces aren't tested.

#### 2.3 Error propagation via `m_result` member

The builder now uses a deferred `m_result` field that's checked at multiple points. This replaces direct `return` values with a "poison-bit" pattern (once `m_result` is failed, all subsequent operations are no-ops). While functional, this pattern:
- Makes control flow harder to reason about (any prior operation could have silently failed)
- Requires discipline to check `m_result` at every entry point
- The existing code already uses `SLANG_RETURN_ON_FAIL` extensively — mixing both patterns is fragile

#### 2.4 `ShaderProgram` multiply-inherits from `ISyntheticShaderProgram`

The `getInterface()` return type changes from `IShaderProgram*` to `void*`, and now responds to two GUIDs. This is fine from a COM perspective, but it means **every program object** carries the `ISyntheticShaderProgram` vtable whether it has synthetic resources or not. The `queryInterface` call will succeed even with zero synthetic bindings (returning a count of 0), which could confuse consumers.

#### 2.5 Backend coverage: only Vulkan + CUDA

D3D12, Metal, and WebGPU aren't implemented. If a user passes synthetic resource descriptors on those backends, program creation will **silently succeed** (the records are stored but never resolved into binding locations). The `bindSyntheticResource()` helper would then fail at `findSyntheticBindingLocationByID` time since no locations were populated. This should probably fail earlier at `createShaderProgram` time if synthetic resources are requested on an unsupported backend.

#### 2.6 Maintenance burden for future backends/changes

Every future backend that implements shader program creation now needs to:
1. Accept `ShaderProgram*` in its layout builder
2. Iterate `m_syntheticResourceInputs`
3. Map binding types to native descriptor concepts
4. Call `addResolvedSyntheticBindingLocation` back into the program

This is a non-trivial maintenance tax, and it's not enforced at compile time (no pure virtual requiring implementation).

---

### 3. Is "Synthetic" the Right Term?

"Synthetic" conveys "compiler-generated / not authored by the user", which is accurate. However, several alternative terms might be clearer:

| Term | Pros | Cons |
|------|------|------|
| **Synthetic** | Familiar in compiler theory | Sounds artificial/fake; not obvious it means "hidden from reflection" |
| **Injected** | Clearly states the host is injecting something extra | Could imply runtime modification |
| **Implicit** | Used in Slang already for implicit parameters | Conflicts with C++ `implicit` keyword meaning |
| **Extra** / **ExtraDescriptor** | PR #726 (closed predecessor) used this | Generic, doesn't convey why |
| **InternalBinding** | Clear about visibility | Could confuse with RHI internals |
| **CompilerBinding** / **CompilerResource** | Most precise: the compiler emitted it | Longer, but very descriptive |

**Recommendation:** "Synthetic" is acceptable, but **"CompilerResource"** or **"CompilerBinding"** would be more self-documenting for a public API. The term should make clear that the *compiler* generated the resource and the *host* is supplying the binding.

---

### 4. Alternative Approaches (Less Intrusive)

#### 4.1 Post-hoc descriptor set patching (most minimal)

Instead of modifying the layout builder, provide a lightweight "extra descriptor write" at bind time:

```cpp
struct ExtraDescriptorWrite {
    uint32_t set;
    uint32_t binding;
    VkDescriptorType type;
    // ...
};
```

The root shader object could accept extra descriptor writes that get applied alongside the reflected bindings during `bindRootShaderObject`. This avoids modifying the layout builder entirely, at the cost of skipping validation that the binding slot is actually in the layout.

**Trade-off:** Less safe (no collision detection), but zero impact on layout construction.

#### 4.2 Extend `slang::ProgramLayout` reflection instead

If Slang's reflection API could expose these "synthetic" bindings as normal reflected parameters (perhaps with a flag marking them as compiler-internal), then slang-rhi wouldn't need *any* special handling — the existing layout builder would pick them up naturally. This pushes complexity into Slang but keeps slang-rhi clean.

**Trade-off:** Requires Slang compiler changes (which are already happening in [shader-slang/slang#11080](https://github.com/shader-slang/slang/pull/11080)), but would eliminate the entire synthetic resource concept from the RHI.

#### 4.3 Dedicated "system binding" descriptor set

Reserve one well-known descriptor set (e.g., set 31) for all compiler-injected resources. The RHI always allocates this set, and `bindSyntheticResource` just writes into it directly without modifying the program layout. This is how many engines handle debug/profiling resources.

**Trade-off:** Wastes a descriptor set slot on every pipeline, but completely decouples synthetic resources from the reflection-driven layout system.

#### 4.4 Composition via wrapper ShaderObject

Create a `SyntheticResourceShaderObject` that wraps the root shader object and intercepts `bindRootShaderObject` to append synthetic descriptor writes. No changes to ShaderProgram or layout needed.

**Trade-off:** Another layer of indirection, but zero intrusion into core.

---

### 5. Unrelated Changes Bundled In

As the author noted, these should be split out:
- **`vk-api.cpp` macOS Vulkan bring-up** (library path candidates, `initEnumerationProcs`)
- **`vk-utils.cpp` / `vk-command.cpp` barrier stage flags** (feature-aware stage masks) — this is a real correctness fix but unrelated
- **`cuda-device.cpp` feature detection refactor** (removing compute-capability-based feature flags) — also unrelated

These represent ~30% of the diff volume and muddy the review.

---

### 6. Verdict

**Disruptiveness: Medium-High.** The PR touches core abstractions (`ShaderProgram`, `ShaderObjectLayout`, `ShaderObject`) that every backend inherits from. The Vulkan descriptor set model was materially changed (map→direct-index). The bidirectional coupling between program and layout builder is a new pattern.

**Maintenance burden: Moderate.** Each new backend needs manual synthetic resource handling. The `m_result` poison-bit error propagation adds cognitive overhead. The empty-descriptor-set behavior for sparse spaces needs ongoing awareness.

**My recommendation:** Approach **4.2** (making Slang's reflection expose these bindings natively) would be the least intrusive long-term solution. If that's not feasible in the near term, **4.3** (dedicated system descriptor set) would keep the RHI layout code clean while still supporting the use case. The current approach works but creates coupling and maintenance burden that will compound as more synthetic resource types are added.

jvepsalainen-nv:
Thanks for the feedback. I'll first create a separate PRs for the unrelated but required changes.
This PR will also require Slang version update. I'll create a separate PR for version update as well when we have it fixed.

jvepsalainen-nv:
Split the unrelated Vulkan fixes out of this PR:

- #748 fixes Vulkan barrier stage masks / readback barrier access.
- #749 fixes macOS Vulkan library loading and adapter-enumeration setup.

This PR branch has been updated so the synthetic-resource diff no longer includes vk-api, vk-backend, vk-command, or vk-utils; it keeps only the Vulkan shader-program/layout changes needed for synthetic resources.

jvepsalainen-nv:
Updated this PR based on the review feedback. 
- Made synthetic resources opt-in:
      - `ISyntheticShaderProgram` is only exposed when the program was created with non-empty synthetic resource descriptors.
      - Normal shader programs no longer expose the synthetic interface or run synthetic layout work.
      - This isolates the functionality to coverage and debug builds that contain non-empty synthetic resource data
  - Isolated unsupported backend behavior:
      - CPU, D3D11, D3D12, Metal, and WGPU keep the normal path unchanged when no synthetic resources are present.
      - If synthetic resources are provided, unsupported backends fail explicitly with `SLANG_E_NOT_IMPLEMENTED`.
  - Cleaned up Vulkan layout builder error handling:
      - Removed the shared `m_result` state / poison-bit pattern.
      - Builder functions now return Result directly and propagate errors with `SLANG_RETURN_ON_FAIL`.
  - Fixed resolved binding metadata details:
      - Stored `SyntheticBindingLocation::structSize` correctly.
      - Kept debug-name lifetime owned by the shader program.
  - Added targeted tests:
      - Ordinary programs return `SLANG_E_NO_INTERFACE` for `ISyntheticShaderProgram`.
      - Unsupported backends reject synthetic resources explicitly.
      - Existing Vulkan/CUDA synthetic binding tests remain in place.
  - Updated docs and PR description:
      - README, docs/api.md, and include/slang-rhi.h now describe opt-in behavior, supported backends, and fail-fast behavior.

CMakeLists.txt
jvepsalainen-nv: Update to Slang v2026.09.1 and remove this

README.md
```diff
@@ -9,6 +9,51 @@ The `slang-rhi` library provides a render hardware interface for the Slang shadi
 It is based on the "gfx" layer originally developed in the Slang repository.
 This library is under active refactoring and development, and is not yet ready for general use.
 
+## Synthetic resource bindings
```
skallweitNV: This feels extremely out of place. You can add a new `synthetic-bindings.md` or similar in the docs folder, but putting it here feels wrong without anything else documented here.

include/slang-rhi.h
```diff
@@ -268,6 +270,86 @@ enum class LinkingStyle
     SeparateEntryPointCompilation
 };
 
+enum class SyntheticResourceScope
```
skallweitNV: I would consider putting all the synthetic API into a separate header file, because 99% of users will not need this.

CMakeLists.txt
```diff
@@ -353,11 +353,16 @@ if(NOT SLANG_RHI_BUILD_FROM_SLANG_REPO)
                 INTERFACE_INCLUDE_DIRECTORIES ${SLANG_RHI_SLANG_INCLUDE_DIR}
                 IMPORTED_LOCATION ${SLANG_RHI_SLANG_BINARY_DIR}/lib/lib${SLANG_LIB_NAME}.a
             )
+            set(SLANG_RHI_SLANG_CMARK_LIB ${SLANG_RHI_SLANG_BINARY_DIR}/lib/libcmark-gfm.a)
```
skallweitNV, line 413: Why does slang-rhi need to deal with the cmark library?!

docs/api.md
```diff
@@ -121,6 +121,57 @@
 | `finalize`                  | yes | yes  | yes   | yes   | yes    | yes   | yes  |
 | `isFinalized`               | yes | yes  | yes   | yes   | yes    | yes   | yes  |
 
+## `ISyntheticShaderProgram` interface
+
+`ISyntheticShaderProgram` is an opt-in interface. A shader program only
```
skallweitNV: The main point of this `api.md` document was to have an overview of what part of the slang-rhi API is implemented for what backend. This is not meant for API documentation.

src/shader.h
```diff
@@ -74,14 +94,35 @@ class ShaderProgram : public IShaderProgram, public DeviceChild
...
+    {
+        return m_syntheticResourceInputs;
+    }
+    Result setResolvedSyntheticBindingLocations(const std::vector<SyntheticBindingLocation>& locations);
+
     // IShaderProgram interface
     virtual SLANG_NO_THROW const ShaderProgramDesc& SLANG_MCALL getDesc() override;
     virtual SLANG_NO_THROW Result SLANG_MCALL getCompilationReport(ISlangBlob** outReportBlob) override;
     virtual SLANG_NO_THROW slang::TypeReflection* SLANG_MCALL findTypeByName(const char* name) override;
+    virtual SLANG_NO_THROW uint32_t SLANG_MCALL getSyntheticBindingCount() override;
```
skallweitNV, line 95: These are not part of the `IShaderProgram` interface. Add a new section with a comment for clarity.

src/shader.h
```diff
@@ -61,6 +79,8 @@ class ShaderProgram : public IShaderProgram, public DeviceChild
     bool m_compiledShaders = false;
 
     std::unordered_map<SpecializationKey, RefPtr<ShaderProgram>, SpecializationKey::Hasher> m_specializedPrograms;
+    std::vector<SyntheticResourceBindingRecord> m_syntheticResourceInputs;
+    std::vector<SyntheticBindingLocation> m_syntheticBindingLocations;
```
skallweitNV: Every program now pays for the extra data. Not by much I guess but it's not really obvious that these are optional and not used by default.

Maybe we could group the data for the synthetic resources in a separate struct and just keep a single pointer here which can be null if there is no synthetic resources?

skallweitNV: I guess some of the helpers could then also move to this new struct/class instead of living on `ShaderProgram`.

src/shader.h
```diff
@@ -9,10 +9,27 @@
 #include "rhi-shared-fwd.h"
 #include "device-child.h"
 
+#include <string>
 #include <unordered_map>
+#include <vector>
 
 namespace rhi {
 
+struct SyntheticResourceBindingRecord
```
skallweitNV: This could be moved together with the struct for holding all the synthetic resource binding stuff into a new header/cpp pair. This would then also be a great place for documenting the system.

src/cuda/cuda-shader-object.cpp
```diff
@@ -138,6 +138,12 @@ Result BindingDataBuilder::writeObjectData(
 )
 {
     size_t size = specializedLayout->getElementTypeLayout()->getSize();
+    // Widen the allocation to satisfy any synthesized resources whose
+    // uniform offset lies outside the Slang element-type layout. Mirrors
+    // the sizing performed by ShaderObject::init for `m_data`.
+    const size_t minUniformSize = specializedLayout->getMinUniformBufferSize();
+    if (minUniformSize > size)
+        size = minUniformSize;
```
skallweitNV: Should we instead have a `getUniformBufferSize()` function instead that returns `getElementTypeLayout()->getSize()` by default and is overriden to add the space for synthetic bindings where needed? Something to think about.

skallweitNV (changes requested):
I would like to split the synthetic resource binding API into a separate header file if possible because it's not something the average user should probably see that prominently. The main change in `slang-rhi.h` would just be the added `ShaderProgramSyntheticResourcesDesc` enum value.

I would also like to move as much of the code into a new header/cpp for the implementation. I think that would help the reader to understand the boundaries better. This might also help reduce memory cost when the feature is not in use.

The CUDA side looks reasonable to me. I haven't looked at the Vulkan part in enough detail, it is more involved of course. The important thing is that there is zero changes/cost to the descriptor set handling if the feature is not in use.

Let me know if you agree with these change requests. I do want to take a better look at the Vulkan side but that's my comments so far.

CMakeLists.txt
jvepsalainen-nv: This should not be needed with Slang 2026.09.1. This was a cleanup mistake after removing workarounds before 2026.09.1

README.md
jvepsalainen-nv: Moved the long-form documentation to docs/synthetic-bindings.md and removed the API section from README.

docs/api.md
jvepsalainen-nv: Trimmed docs/api.md back to backend support tables and moved API usage documentation to docs/synthetic-bindings.md.

CMakeLists.txt
jvepsalainen-nv: I rechecked this with Slang 2026.9.1. The Emscripten static slang-compiler archive still references cmark symbols through slang-markdown.cpp, so the wasm link fails without linking the packaged libcmark-gfm.a. I kept this as package link glue rather than a fallback/workaround; we can remove it once the Slang package no longer requires the separate archive.

include/slang-rhi.h
jvepsalainen-nv: Moved the optional synthetic-resource API to <slang-rhi/synthetic-bindings.h>. The main header now only keeps the StructType enum value needed for ShaderProgramDesc.next.

src/shader.h
jvepsalainen-nv: Moved the storage, validation, descriptor copying, and resolved-location handling into SyntheticResourceBindingState in src/synthetic-resource-bindings.*. ShaderProgram now only keeps a nullable pointer for optional state.

src/shader.h
jvepsalainen-nv: Moved the shared synthetic-resource implementation into src/synthetic-resource-bindings.h/.cpp. The optional public API is documented separately in docs/synthetic-bindings.md and declared in <slang-rhi/synthetic-bindings.h>.

src/cuda/cuda-shader-object.cpp
jvepsalainen-nv: Changed this to getUniformBufferSize(), with the base implementation returning the Slang element-type size and CUDA overriding it when synthetic uniform marshaling extends the required storage.

src/shader.h
jvepsalainen-nv: Split the declarations into separate IShaderProgram and ISyntheticShaderProgram sections for clarity.

docs/api.md
```diff
@@ -121,6 +121,30 @@
...
+
+| Descriptor path                         | CPU | CUDA | D3D11 | D3D12 | Vulkan | Metal | WGPU |
+|-----------------------------------------|-----|------|-------|-------|--------|-------|------|
+| `ShaderProgramSyntheticResourcesDesc`   | :x: | yes  | :x:   | :x:   | yes    | :x:   | :x:  |
+
+## Synthetic binding helper
+
+| Helper API                   | CPU | CUDA | D3D11 | D3D12 | Vulkan | Metal | WGPU |
+|-----------------------------|-----|------|-------|-------|--------|-------|------|
+| `bindSyntheticResource(...)`| :x: | yes  | :x:   | :x:   | yes    | :x:   | :x:  |
```
skallweitNV: minor, but these tables are not aligned as the rest are

docs/synthetic-bindings.md
```diff
@@ -0,0 +1,67 @@
...
+The intended flow is:
+
+1. Query hidden-resource metadata from Slang, typically through
+   `slang::ISyntheticResourceMetadata`.
+2. Translate each metadata record into `SyntheticResourceBindingDesc`.
+3. Chain `ShaderProgramSyntheticResourcesDesc` through
+   `ShaderProgramDesc.next` when creating the shader program.
+4. Bind the resource through `bindSyntheticResource(...)`, or query
+   `ISyntheticShaderProgram` and call `IShaderObject::setBinding()`
+   with the resolved `SyntheticBindingLocation`.
```
skallweitNV, line 44: If Slang provides interfaces to query these synthetic resource bindings, should we have helpers in here that can directly operate on these and make using them more convenient? could be a follow-up as well

CMakeLists.txt
skallweitNV: ok, can you create a ticket for this so we don't forget?

src/shader.h
```diff
@@ -74,14 +84,32 @@ class ShaderProgram : public IShaderProgram, public DeviceChild
 
     virtual Result createShaderModule(slang::EntryPointReflection* entryPointInfo, ComPtr<ISlangBlob> kernelCode);
 
+    bool hasSyntheticResourceInputs() const;
+    const std::vector<SyntheticResourceBindingRecord>& getSyntheticResourceInputs() const;
+    Result setResolvedSyntheticBindingLocations(const std::vector<SyntheticBindingLocation>& locations);
```
skallweitNV: instead of these 3 members, can we just have a single accessors `const SyntheticResourceBindingState* getSyntheticResourceBindingState() const` and then operate on that directly?

src/cuda/cuda-device.cpp
```diff
@@ -444,7 +445,20 @@ Result DeviceImpl::createShaderProgram(
...
+        RootShaderObjectLayoutImpl::create(
+            this,
+            shaderProgram->linkedProgram->getLayout(),
+            syntheticResources,
+            syntheticResources ? &syntheticLocations : nullptr,
+            shaderProgram->m_rootObjectLayout.writeRef()
+        )
+    );
+    if (syntheticResources)
+        SLANG_RETURN_ON_FAIL(shaderProgram->setResolvedSyntheticBindingLocations(syntheticLocations));
```
skallweitNV: if you have an accessor for the synthetic resource state, you can just pass that into the `RootShaderObjectLayoutImpl::create` and call the `setResolvedSyntheticBindingLocations` directly on the state.

src/shader.cpp
```diff
@@ -204,6 +209,80 @@ const ShaderProgramDesc& ShaderProgram::getDesc()
...
+    SLANG_RHI_ASSERT(m_syntheticResources);
+    return m_syntheticResources->getInputs();
+}
+
+Result ShaderProgram::_initSyntheticResourceDescs()
+{
+    if (!m_desc.next)
+        return SLANG_OK;
+
+    std::unique_ptr<SyntheticResourceBindingState> syntheticResources =
```
skallweitNV, line 253: these feels wrong, you create the SyntheticResourceBindingState unconditionally when `m_desc.next` is not nullptr? this really should only happen if the desc for synthetic resource bindings is found. i would probably search for that desc and then only initialize if it is found.

src/shader.h
```diff
@@ -9,10 +9,16 @@
 #include "rhi-shared-fwd.h"
 #include "device-child.h"
 
+#include <string>
+#include <memory>
 #include <unordered_map>
+#include <vector>
```
skallweitNV, line 15: these still all needed?

tests/test-synthetic-resource-bindings.cpp
```diff
@@ -0,0 +1,678 @@
+#include "testing.h"
```
skallweitNV, line 1: AI tends to write questionable tests sometimes. Did you verify that these all make sense?

src/shader-object.cpp
```diff
@@ -498,6 +498,10 @@ Result ShaderObject::init(Device* device, ShaderObjectLayout* layout)
         uniformSize = layout->getElementTypeLayout()->getSize();
     }
 
+    const size_t layoutUniformSize = layout->getUniformBufferSize();
+    if (layoutUniformSize > uniformSize)
+        uniformSize = layoutUniformSize;
+
```
skallweitNV: shouldn't the `else` case just use `layout->getUniformBufferSize()` directly and then the rest down here is not needed?!

src/vulkan/vk-shader-object.cpp
```diff
@@ -670,7 +670,6 @@ Result BindingDataBuilder::allocateDescriptorSets(
     ShaderObjectLayoutImpl* specializedLayout
 )
 {
-    SLANG_RHI_ASSERT(specializedLayout->getOwnDescriptorSets().size() <= 1);
```
skallweitNV, line 673: why is this assert removed?

skallweitNV (commented):
This moves into the right direction now. Left a few comments. Still need to dig through the Vulkan changes though.

docs/api.md
jvepsalainen-nv: Fixed

docs/synthetic-bindings.md
jvepsalainen-nv: Yes, helper APIs that consume Slang metadata directly would make sense. I suggest to keep this PR focused on the RHI-side binding and add the helpers as a follow-up.

src/shader.h
jvepsalainen-nv: Fixed by using getSyntheticResourceBindingState() and moving resolved-location updates onto SyntheticResourceBindingState.

src/cuda/cuda-device.cpp
jvepsalainen-nv: Fixed as suggested. CUDA now passes SyntheticResourceBindingState into root layout creation and the layout writes resolved locations back to the state.

src/shader.cpp
jvepsalainen-nv: This was incorrect. ShaderProgram now first scans the ShaderProgramDesc.next chain for ShaderProgramSyntheticResourcesDesc and creates SyntheticResourceBindingState only if that desc is present.

src/shader.h
jvepsalainen-nv: Removed the unused SyntheticResourceBindingRecord forward declaration. The remaining includes are still needed; core/timer.h provides TimePoint used by ShaderCompilationReporter declarations in this header.

src/shader-object.cpp
jvepsalainen-nv: Fixed. The non-ParameterBlock path now uses `layout->getUniformBufferSize()` directly and the redundant max block is removed.

src/vulkan/vk-shader-object.cpp
jvepsalainen-nv: Synthetic resources can preserve explicit Vulkan set numbers from compiler metadata, so a root object can now own multiple descriptor sets. Added a comment at the allocation site to document that this is expected.

PR #748, opened 2026-05-19
Fix Vulkan barrier stage masks

jvepsalainen-nv:
## Summary
- make Vulkan barrier stage masks depend on enabled device features instead of always including optional shader stages
- include mesh/task shader stages when supported
- include acceleration-structure build stage for acceleration-structure read/query/copy paths
- fix readBuffer transition dst access to CopySource

## Notes
Split out from shader-slang/slang-rhi#739 to keep the synthetic-resource binding PR focused.

### DIFF
--- a/src/vulkan/vk-utils.cpp
+++ b/src/vulkan/vk-utils.cpp
@@ -328,7 +328,28 @@ VkAccessFlagBits calcAccessFlags(ResourceState state)
     }
 }
 
-VkPipelineStageFlagBits calcPipelineStageFlags(ResourceState state, bool src)
+static VkPipelineStageFlags getSupportedShaderStageFlags(const VulkanApi& api)
+{
+    VkPipelineStageFlags result = VK_PIPELINE_STAGE_VERTEX_SHADER_BIT |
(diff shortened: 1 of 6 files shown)


src/vulkan/vk-utils.cpp
```diff
@@ -328,7 +328,32 @@ VkAccessFlagBits calcAccessFlags(ResourceState state)
     }
 }
 
-VkPipelineStageFlagBits calcPipelineStageFlags(ResourceState state, bool src)
+static VkPipelineStageFlags getSupportedShaderStageFlags(const VulkanApi& api)
```
skallweitNV: Feels rather expensive to call this for every `calcPipelineStageFlags`. We should probably pass the supported stage flags to `calcPipelineStageFlags` and precompute them for the device once.

jvepsalainen-nv: I reworked this so that the supported shader stage mask is now computed once during Vulkan device initialization and cached on `VulkanApi`.
`calcPipelineStageFlags` now takes the cached mask directly instead of checking device features on every barrier calculation.

skallweitNV (approved):

PR #749, opened 2026-05-19
Fix Vulkan adapter enumeration on macOS

jvepsalainen-nv:
## Summary
- load Vulkan from common MoltenVK/Homebrew library locations on Apple platforms
- initialize the smaller Vulkan instance-proc set needed for adapter enumeration instead of requiring full device-instance proc loading
- use KHR fallbacks for physical-device feature/property queries
- enable the external-memory-capabilities instance extension needed for physical-device ID properties on Vulkan 1.0

## Notes
Split out from shader-slang/slang-rhi#739 to keep the synthetic-resource binding PR focused.
This PR depends on shader-slang/slang-rhi#748.

### DIFF
--- a/src/vulkan/vk-api.cpp
+++ b/src/vulkan/vk-api.cpp
@@ -140,13 +140,13 @@ Result VulkanApi::initEnumerationProcs(VkInstance instance)
         (PFN_vkEnumeratePhysicalDevices)vkGetInstanceProcAddr(instance, "vkEnumeratePhysicalDevices");
     vkGetPhysicalDeviceProperties2 =
         (PFN_vkGetPhysicalDeviceProperties2)vkGetInstanceProcAddr(instance, "vkGetPhysicalDeviceProperties2");
-    vkEnumerateDeviceExtensionProperties =
-        (PFN_vkEnumerateDeviceExtensionProperties)vkGetInstanceProcAddr(
-            instance,
-            "vkEnumerateDeviceExtensionProperties"
-        );
+    if (!vkGetPhysicalDeviceProperties2)
+    {
+        vkGetPhysicalDeviceProperties2 =
+            (PFN_vkGetPhysicalDeviceProperties2)vkGetInstanceProcAddr(instance, "vkGetPhysicalDeviceProperties2KHR");
+    }
 
-    if (!(vkEnumeratePhysicalDevices && vkGetPhysicalDeviceProperties2 && vkEnumerateDeviceExtensionProperties))
+    if (!(vkEnumeratePhysicalDevices && vkGetPhysicalDeviceProperties2))
     {
         return SLANG_FAIL;
     }
@@ -165,6 +165,16 @@ Result VulkanApi::initInstanceProcs(VkInstance instance)
 
     // Get optional
     VK_API_INSTANCE_PROCS_OPT(VK_API_GET_INSTANCE_PROC)
+    if (!vkGetPhysicalDeviceFeatures2)
+    {
+        vkGetPhysicalDeviceFeatures2 =
+            (PFN_vkGetPhysicalDeviceFeatures2)vkGetInstanceProcAddr(instance, "vkGetPhysicalDeviceFeatures2KHR");
+    }
+    if (!vkGetPhysicalDeviceProperties2)
+    {
+        vkGetPhysicalDeviceProperties2 =
+            (PFN_vkGetPhysicalDeviceProperties2)vkGetInstanceProcAddr(instance, "vkGetPhysicalDeviceProperties2KHR");
+    }
 
     if (!areDefined(ProcType::Instance))
     {
--- a/src/vulkan/vk-api.cpp
+++ b/src/vulkan/vk-api.cpp
@@ -13,18 +13,6 @@
 
 namespace rhi::vk {
 
-#if SLANG_APPLE_FAMILY
-static void* _tryLoadVulkanModule(const char* const* candidates, size_t candidateCount)
-{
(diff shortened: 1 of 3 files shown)


src/vulkan/vk-backend.cpp
```diff
@@ -31,15 +31,21 @@ Result BackendImpl::createDevice(const DeviceDesc& desc, IDevice** outDevice)
...
+    {
+        return SLANG_FAIL;
+    }
 
     VulkanApi api;
-    SLANG_RETURN_ON_FAIL(api.initGlobalProcs(module));
+    if (SLANG_FAILED(api.initGlobalProcs(module)))
+    {
+        return SLANG_FAIL;
+    }
```
skallweitNV: What is wrong with using `SLANG_RETURN_ON_FAIL`?!
Also why do we remove the deferred destruction of the module?

src/vulkan/vk-backend.cpp
```diff
@@ -50,14 +56,13 @@ Result BackendImpl::enumerateAdapters()
     instanceCreateInfo.flags = VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
 #endif
     VkInstance instance;
-    SLANG_VK_RETURN_ON_FAIL(api.vkCreateInstance(&instanceCreateInfo, nullptr, &instance));
+    if (api.vkCreateInstance(&instanceCreateInfo, nullptr, &instance) != VK_SUCCESS)
```
skallweitNV: What's wrong with using `SLANG_VK_RETURN_ON_FAIL`?

src/vulkan/vk-backend.cpp
```diff
@@ -50,14 +56,13 @@ Result BackendImpl::enumerateAdapters()
...
+        return SLANG_FAIL;
+    }
     SLANG_RHI_DEFERRED({ api.vkDestroyInstance(instance, nullptr); });
 
-    // This will fail due to not loading any extensions.
-    api.initInstanceProcs(instance);
-
-    if (!(api.vkEnumeratePhysicalDevices && api.vkGetPhysicalDeviceProperties2 &&
-          api.vkEnumerateDeviceExtensionProperties))
+    if (SLANG_FAILED(api.initEnumerationProcs(instance)))
```
skallweitNV: Should probably use `SLANG_RETURN_ON_FAIL` as well.

src/vulkan/vk-backend.cpp
```diff
@@ -31,15 +31,21 @@ Result BackendImpl::createDevice(const DeviceDesc& desc, IDevice** outDevice)
...
-    SLANG_RETURN_ON_FAIL(api.initGlobalProcs(module));
+    if (SLANG_FAILED(api.initGlobalProcs(module)))
+    {
+        return SLANG_FAIL;
+    }
 
     VkInstanceCreateInfo instanceCreateInfo = {VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO};
     const char* instanceExtensions[] = {
         VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME,
+        VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME,
```
skallweitNV: Can we expect this to be always available?

src/vulkan/vk-api.cpp
```diff
@@ -32,7 +44,15 @@ Result VulkanModule::init()
...
-    m_module = dlopen("libvulkan.1.dylib", RTLD_NOW | RTLD_GLOBAL);
+    static const char* const kVulkanModuleCandidates[] = {
+        "libvulkan.1.dylib",
+        "libvulkan.dylib",
+        "/opt/homebrew/lib/libvulkan.1.dylib",
+        "/opt/homebrew/lib/libvulkan.dylib",
+        "/usr/local/lib/libvulkan.1.dylib",
+        "/usr/local/lib/libvulkan.dylib",
+    };
+    m_module = _tryLoadVulkanModule(kVulkanModuleCandidates, SLANG_COUNT_OF(kVulkanModuleCandidates));
```
skallweitNV: could do the loading inline using something like
```c++
for (size_t i = 0; i <= SLANG_COUNT_OF(kVulkanModuleCandidates) && !m_module; ++i)
{
    m_module = dlopen(candidates[i], RTLD_NOW | RTLD_GLOBAL));
}
```

src/vulkan/vk-api.cpp
```diff
@@ -112,6 +132,29 @@ Result VulkanApi::initGlobalProcs(const VulkanModule& module)
     return SLANG_OK;
 }
 
+Result VulkanApi::initEnumerationProcs(VkInstance instance)
```
skallweitNV: Do we really need to treat these separately. We have the concept of optional procs, shouldn't they just be optional instead? I just see these now loaded in two different places which feels wrong to me.

skallweitNV (commented):
Human review :)

jvepsalainen-nv:
> Human review :)

This one was still on a "workaround it to pipeclean the prototype working"-shape. Fair comments - I'll rework this a bit.

src/vulkan/vk-backend.cpp
jvepsalainen-nv: I think there was no reason for this path to diverge from the existing style. Reverted this.

src/vulkan/vk-backend.cpp
jvepsalainen-nv: This also looks not needed.

skallweitNV (approved):
Thanks for the cleanup!

PR #753, opened 2026-05-22
Add bounds assertions to CPU backend buffer operations

expipiplus1:
## Summary
- Add `SLANG_RHI_ASSERT` bounds checks to `cmdCopyBuffer` and `cmdClearBuffer` in the CPU backend
- These are internal invariant checks; if the offsets/sizes are out of bounds at this point it indicates a bug in the command recording layer

## Test plan
- [ ] Verify CPU backend tests pass in both Debug and Release

### DIFF
--- a/src/cpu/cpu-command.cpp
+++ b/src/cpu/cpu-command.cpp
@@ -97,6 +97,8 @@ void CommandExecutor::cmdCopyBuffer(const commands::CopyBuffer& cmd)
 {
     BufferImpl* dst = checked_cast<BufferImpl*>(cmd.dst);
     BufferImpl* src = checked_cast<BufferImpl*>(cmd.src);
+    SLANG_RHI_ASSERT(cmd.dstOffset + cmd.size <= dst->m_desc.size);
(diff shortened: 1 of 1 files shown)


src/cpu/cpu-command.cpp
```diff
@@ -97,6 +97,8 @@ void CommandExecutor::cmdCopyBuffer(const commands::CopyBuffer& cmd)
 {
     BufferImpl* dst = checked_cast<BufferImpl*>(cmd.dst);
     BufferImpl* src = checked_cast<BufferImpl*>(cmd.src);
+    SLANG_RHI_ASSERT(cmd.dstOffset + cmd.size <= dst->m_desc.size);
```
skallweitNV, line 100: not sure these make sense. the debug layer should catch these. why do we need these asserts?

expipiplus1: Argus generated a "security report" which was upset here, I'll close this PR as false positive since we have the debug layer for this.

PR #754, opened 2026-05-22
Validate input element count in D3D11 createInputLayout

expipiplus1:
## Summary
- The `inputElements` array is fixed at 16 entries but `desc.inputElementCount` is not validated, allowing a stack buffer overflow with more than 16 elements
- Add an early-out check and replace the magic number with a named constant

Note: the original bug report also flagged `sprintf` usage, but that was already fixed in 70118ca0 ("Refactor shader object #166") which changed all calls to `snprintf` with proper size limits.

## Test plan
- [ ] Verify D3D11 backend tests pass

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(gh api:*)"
+    ],
+    "deny": [],
+    "ask": []
+  }
+}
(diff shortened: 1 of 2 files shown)


skallweitNV (dismissed):
LGTM but needs a pre-commit run to fix the formatting

skallweitNV:
we could add a validation check for d3d11 in the debug layer to get out a better error message. but i personally don't care too much about d3d11.

PR #756, opened 2026-05-23
[Metal] Fix root acceleration structure bindings

fknfilewalker:
Previously, the Metal backend silently ignored root-level acceleration structures because it falsely assumed they would only ever be bound indirectly via argument buffers. Slang compiles these structures into MSL as standard [[buffer(N)]] parameters, which cannot be bound via setBuffer and require a dedicated method. This PR resolves the mismatch by explicitly tracking root-level acceleration structures and correctly binding them to their respective buffer slots using the required setAccelerationStructure API.


fknfilewalker:
Sorry, I will fix the format later.

src/metal/metal-command.cpp
```diff
@@ -812,6 +812,32 @@ void CommandRecorder::cmdSetComputeState(const commands::SetComputeState& cmd)
...
+                {
+                    if (as)
+                        validAS.push_back(as);
+                }
+                if (!validAS.empty())
+                {
+                    encoder->useResources(validAS.data(), validAS.size(), MTL::ResourceUsageRead);
+                }
+            }
+        }
```
skallweitNV: While this code only runs on older GPUs without residency, creating a temporary `validAS` vector is not great. Can we use `Device::getAccelerationStructureArray()` instead which has lazy-evaluation?

src/metal/metal-shader-object.cpp
```diff
@@ -267,7 +274,43 @@ Result BindingDataBuilder::bindAsValue(
             }
             break;
         case slang::BindingType::RayTracingAccelerationStructure:
-            // Acceleration structures are bound via argument buffers, not directly
+            for (uint32_t i = 0; i < count; ++i)
+            {
+                const ResourceSlot& slot = shaderObject->m_slots[slotIndex + i];
+                AccelerationStructureImpl* accelerationStructure =
+                    checked_cast<AccelerationStructureImpl*>(slot.resource.get());
+                if (accelerationStructure)
```
skallweitNV: Should the `nullptr` case clear the acceleration structure in `rootAccelerationStructures`/`rootAccelerationStructureSlots`?

skallweitNV (commented):
Thanks for the contribution. Two questions from my side, otherwise LGTM.

src/metal/metal-command.cpp
fknfilewalker: Two problems. The first is that useResources requires a c array and ns::array doesn't expose a contiguous pointer to its internal storage. Second is that the resource list should not contain nullptrs, which could happen with the array.

skallweitNV: In that case, we probably want to maintain the vector on device and lazy evaluate when needed?!

src/metal/metal-device.cpp
```diff
@@ -407,6 +407,12 @@ NS::Array* DeviceImpl::getAccelerationStructureArray()
 {
     if (m_accelerationStructures.dirty)
     {
+        m_accelerationStructures.validResources.clear();
```
skallweitNV: We should move this to a separate function like `getValidAccelerationStructureResources()` and have a separate dirty flag. It's not great to have to call `getAccelerationStructureArray()` and not use the result just to update the valid resources.

skallweitNV (approved):
LGTM

PR #759, opened 2026-05-28
Cleanup metal acceleration structure registry

skallweitNV:
Minor cleanup for handling Metal acceleration structure registry.


skallweitNV:
@fknfilewalker some minor cleanup on your recent PR, just FYI

PR #760, opened 2026-05-28
vulkan: gate NVX and ray tracing extensions behind opt-out flags

wlewNV:
## Summary

Adds two new opt-out flags to `DeviceDesc`, both default `true` (no behaviour change for existing consumers):

- **`enableCUDALaunchFromGfx`** — gates `VK_NVX_binary_import` + `VK_NVX_image_view_handle` (exposes `vkCmdCuLaunchKernelNVX`).
- **`enableRayTracing`** — gates `VK_KHR_acceleration_structure` and dependents (`ray_tracing_pipeline`, `ray_query`, `ray_tracing_position_fetch`, `VK_NV_ray_tracing_linear_swept_spheres`, `VK_NV_cluster_acceleration_structure`).

## Why

On NVIDIA Linux + Blackwell sm_120 + driver 595.x, enabling **either** of these extension groups at `vkCreateDevice` brings up a dual-purpose CUDA primary context (or a CUDA-adjacent BVH/RT execution path) inside the driver. This perturbs cuDNN's MHA dispatcher engine catalog: torch's `_scaled_dot_product_cudnn_attention` raises `RuntimeError: Expected mha_graph.execute(...).is_good() to be true` on the first attention call in the same process.

slang-rhi today eagerly enables every extension the driver advertises, so any slang-rhi/slangpy consumer that also uses cuDNN MHA hits this on first-gen Blackwell hardware. Applications that don't use `vkCmdCuLaunchKernelNVX` or RT can now opt out and avoid it.

## Verified

Minimal repro: created Vulkan instance + device via raw `vkCreate*` with selectable extension sets, then exercised `torch.ops.aten._scaled_dot_product_cudnn_attention` with the model shapes used by our world-model pipeline.

- baseline (no NVX, no RT) → works
- `+VK_NVX_binary_import` → `is_good()=false`
- `+VK_KHR_acceleration_structure` family → `is_good()=false`
- everything *except* those two groups → works

Tested on RTX PRO 6000 Blackwell + Ubuntu 26.04 + NVIDIA driver 595.71.05 + cuDNN 9.20 + torch 2.12. WSL2 boxes are unaffected (DXVK doesn't expose the offending extensions in the first place).

### DIFF
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -3258,24 +3258,18 @@ struct DeviceDesc
     bool enableCompilationReports = false;
 
     /// Enable launching CUDA kernels from inside graphics command buffers
-    /// (Vulkan only, via VK_NVX_binary_import + VK_NVX_image_view_handle).
-    /// On by default to preserve historical behaviour. Set to false on
-    /// NVIDIA Linux + Blackwell sm_120 (driver 595.x) if you don't need
-    /// vkCmdCuLaunchKernelNVX: enabling those two extensions forces the
-    /// driver to bring up a dual-purpose CUDA primary context, which has
-    /// been observed to perturb cuDNN's MHA dispatcher engine catalog
-    /// (torch's _scaled_dot_product_cudnn_attention starts returning
-    /// mha_graph.execute().is_good() == false).
+    /// (Vulkan only, via VK_NVX_binary_import). On by default. Set to
+    /// false if the application doesn't need vkCmdCuLaunchKernelNVX;
+    /// enabling this extension has been observed to interfere with
+    /// concurrent cuDNN usage on some driver/GPU pairs.
     bool enableCUDALaunchFromGfx = true;
 
-    /// Enable Vulkan ray tracing extensions (VK_KHR_acceleration_structure,
-    /// VK_KHR_ray_tracing_pipeline, VK_KHR_ray_query, VK_KHR_ray_tracing_position_fetch,
-    /// VK_NV_ray_tracing_linear_swept_spheres, VK_NV_cluster_acceleration_structure).
-    /// On by default to preserve historical behaviour. Set to false on
-    /// NVIDIA Linux + Blackwell sm_120 (driver 595.x) if your application
-    /// doesn't use ray tracing: enabling these extensions initializes a
-    /// CUDA-adjacent BVH/RT execution path inside the driver that, like
-    /// CUDA-launch-from-gfx, can perturb cuDNN's MHA dispatcher.
+    /// Enable Vulkan ray tracing extensions (acceleration_structure,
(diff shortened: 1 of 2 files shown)


include/slang-rhi.h
```diff
@@ -3257,6 +3257,27 @@ struct DeviceDesc
     /// Enable reporting of shader compilation timings.
     bool enableCompilationReports = false;
 
+    /// Enable launching CUDA kernels from inside graphics command buffers
```
csyonghe, line 3260: the comments here can be simplified, do not need to mentioned the detailed case of linux +_ blackwell setup and torch calls, that is too specific.

src/vulkan/vk-device.cpp
```diff
@@ -1096,11 +1098,14 @@ Result DeviceImpl::initVulkanDevice(
...
         }
-        if (extensionNames.count(VK_NVX_BINARY_IMPORT_EXTENSION_NAME))
+        // See DeviceDesc::enableCUDALaunchFromGfx for rationale.
+        if (desc.enableCUDALaunchFromGfx &&
+            extensionNames.count(VK_NVX_BINARY_IMPORT_EXTENSION_NAME))
         {
             deviceExtensions.push_back(VK_NVX_BINARY_IMPORT_EXTENSION_NAME);
         }
-        if (extensionNames.count(VK_NVX_IMAGE_VIEW_HANDLE_EXTENSION_NAME))
+        if (desc.enableCUDALaunchFromGfx &&
```
csyonghe: why is image view handle extension protected too?

wlewNV: That was an artifact from my triage, removed the VK_NVX_IMAGE_VIEW_HANDLE_EXTENSION_NAME protection.

csyonghe (approved):

skallweitNV:
@wlewNV This change feels rushed.

First, `enableCUDALaunchFromGfx` just controls adding `VK_NVX_BINARY_IMPORT_EXTENSION_NAME`. If we need to gate adding that extension, maybe spelling it out is better than a generic sounding name that doesn't really do anything else.

Second, `enableRayTracing` only affects the Vulkan backend. Again, it should probably spell that out in the name.

However, both of these are just work-arounds for very specific things. I wonder if they really should be in the main `DeviceDesc`? Is this a temporary issue that will resolve with future driver versions? The thing that worries me is that this also was added to slangpy. If this is a temporary issue then maybe a env variable would have been the better solution instead of threading it through the public API.

If we truly want to be able to control features, this should be implemented in a more thought out way so that it works across all backends.

wlewNV:
Apologies! This was a rushed solution to support an urgent customer product.
I will take responsibility on either reverting or coming up with the proper solution.

I have opened the issue in the slangpy repository: https://github.com/shader-slang/slangpy/issues/1011

PR #779, opened 2026-06-25
CUDA: route hoisted entry-point uniforms to SLANG_globalParams at dispatch

szihs:
## Summary (DRAFT — host half of shader-slang/slang#11747)

Companion to the Slang compiler pass in **shader-slang/slang#11747**, which hoists a CUDA compute kernel's runtime-indexed resource-array entry-point uniforms into a `__constant__ SLANG_globalParams` block (to avoid the slow serial `.param` `ld.param` chain). Because that hoist is a **late emit pass**, it is invisible to reflection — so slang-rhi otherwise binds the uniforms into the `cuLaunchKernel` argument buffer while the emitted kernel reads `__constant__` → `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES`.

This adds the dispatch-time routing in `CommandExecutor::cmdDispatchCompute`: when a compute kernel's reflection reports entry-point uniforms (`entryPointData.size > 0`) **and** the module exposes a `SLANG_globalParams` `__constant__` symbol (`m_globalParams` valid) **and** there is no separate global-param data (`globalParamsSize == 0`) — i.e. the compiler hoisted the entry-point uniforms — copy the entry-point uniform blob into `SLANG_globalParams` (`cuMemcpyHtoDAsync`) and launch with an empty parameter buffer. Non-hoisted kernels are unaffected.

**Validated** with slang#11747 on an L40S (SlangPy `test_tensor_sum_indirect`): 0.103 / 0.744 / 2.585 ms → 0.026 / 0.031 / 0.208 ms (4× / 24× / 12×), results correct, no `globalParamsSize` mismatch (byte layouts match).

⚠️ Inert without shader-slang/slang#11747 (the routing condition only arises when that pass hoists). Draft until the pair is co-reviewed.

<sub>🤖 Generated by an automated coworker — may be inaccurate. A human maintainer should verify.</sub>

### DIFF
--- a/src/cuda/cuda-command.cpp
+++ b/src/cuda/cuda-command.cpp
@@ -640,38 +640,64 @@ void CommandExecutor::cmdDispatchCompute(const commands::DispatchCompute& cmd)
     SLANG_RHI_ASSERT(computePipeline->m_entryPointIndex < bindingData->entryPointCount);
     const auto& entryPointData = bindingData->entryPoints[computePipeline->m_entryPointIndex];
 
+    // The argument data for the entry-point parameters is stored in host memory, normally passed
+    // to cuLaunchKernel as the launch parameter buffer.
+    const void* launchParamData = entryPointData.data;
+    size_t launchParamSize = entryPointData.size;
+
     // Copy global parameter data to the `SLANG_globalParams` symbol.
     if (computePipeline->m_globalParams)
     {
+        // A late CUDA IR pass can hoist a compute kernel's entry-point uniform parameters into the
+        // `SLANG_globalParams` __constant__ block. In that case reflection still reports them as
+        // entry-point uniforms (so their data lands in `entryPointData` and `globalParams` is
+        // empty), but the emitted kernel reads `__constant__` and takes no launch arguments. Detect
+        // that and route the entry-point uniform blob (host memory) into the symbol, then launch
+        // with no argument buffer.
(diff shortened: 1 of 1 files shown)


skallweitNV:
@szihs is this still in draft state?

szihs:
No, it's not in draft — the PR's draft flag is already off; the `[Draft]` was just a stale prefix left in the title, which I've now removed. It's ready for review.

For context: this is the slang-rhi half of a pair with shader-slang/slang#11747 (a CUDA-only Slang pass that hoists a compute entry point's runtime-indexed fixed-size resource-array uniforms into a `ConstantBuffer<GlobalParams>`, emitted as `__constant__ SLANG_globalParams`). #779 routes the entry-point uniform data to that `SLANG_globalParams` symbol at dispatch, since reflection still reports the original per-parameter entry-point layout. It's backward-compatible (only takes effect when the hoisted global is present), so it's intended to land first — the Slang side then picks it up via a slang-rhi submodule bump.

Happy to address any review feedback.

<sub>🤖 Generated by an automated SlangPy coworker — may be inaccurate. A human maintainer should verify.</sub>

szihs:
> @szihs is this still in draft state?

No. I would like to get it to review. Let me add you to a thread for full context

szihs:
Since this is doing HtoD cuda copy every dispatch. This is not the right approach.

PR #780, opened 2026-06-26
Update Dawn WebGPU binaries

### LINKED ISSUE #11732: SPIR-V: groupshared (Workgroup) array variables lack explicit layout — rejected by newer Vulkan validators (VUID-StandaloneSpirv-None-10684)
_Split out from #11720_ — that report ("slang-test fails on Windows when ran without test-server") bundled several distinct SPIR-V validation errors; this issue tracks the **groupshared (Workgroup) array explicit-layout** one. (The atomic VUID is #11731; the test-isolation problem stays in #11720.)

## Summary
Newer Vulkan validation layers (SDK 1.4.328, via `-enable-debug-layers`) reject groupshared (`Workgroup` storage class) array variables that lack explicit layout decorations:

```
Error: VUID-StandaloneSpirv-None-10684: vkCreateShaderModule(): spirv-val produced an error:
Invalid explicit layout decorations on type for operand '7[%7]'
  %6 = OpVariable %_ptr_Workgroup__arr_uint_uint_6 Workgroup %10
Command to reproduce: spirv-val <input.spv> --relax-block-layout --target-env vulkan1.3
```

## Affected tests (from #11720's run)
- `tests/bugs/array-size-groupshared.slang` (`_arr_uint_uint_6`)
- `tests/bugs/generic-groupshared.slang` (`_arr_uint_uint_4`)
- `tests/compute/groupshared.slang` (`_arr_int_uint_4`)
- `tests/front-end/static-var-in-generic-type.slang` (`_arr_int_uint_2`)
- `tests/metal/groupshared-threadlocal-same-parameter.slang` (`_arr_uint_uint_2`)

## Findings (HEAD `e21cdfafa`)
- Slang emits Workgroup array variables with **no `ArrayStride`** decoration: `getPointerArrayStrideValue` (`slang-emit-spirv.cpp:1993`) only emits a stride for `StorageBuffer`/`PhysicalStorageBuffer` (`:1998-1999`), so Workgroup arrays get none. Reproduced standalone: `groupshared uint a[6]` → `%v = OpVariable %_ptr_Workgroup__arr_uint_int_6 Workgroup` with no `ArrayStride`.
- Slang declares **no** `WorkgroupMemoryExplicitLayoutKHR` capability / `SPV_KHR_workgroup_memory_explicit_layout` extension anywhere (whole-source grep, empty).
- Crucially, Slang's **embedded** spirv-tools (v2026.2) **accepts** this SPIR-V (`SLANG_RUN_SPIRV_VALIDATION=1` → rc 0). Only the newer system Vulkan layer (SDK 1.4.328) rejects it. This is the opposite of the atomic VUID (#11731), which the embedded validator *does* reject.

## Open question (needs a maintainer / spec call) — no fix proposed yet
This looks like validator version-skew. VUID-10684 normally requires explicit layout only when `WorkgroupMemoryExplicitLayoutKHR` is declared (which Slang does not). Either:
- the newer layer now enforces explicit layout on plain `Workgroup` arrays → Slang should emit `ArrayStride` (and declare the capability) for groupshared arrays; or
- the layer is over-strict for this case → pin/track the validator version rather than change emission.

Tracking issue — holding for that decision before any fix.

<sub>🤖 Generated by an automated Slang coworker — may be inaccurate. A human maintainer should verify.</sub>

jkwak-work:
Fixes shader-slang/slang#11732

## Motivation

The Slang `wgpu` execution path sends WGSL to Dawn, and Dawn uses Tint internally to lower
that WGSL to the native backend shader format. With the older Dawn package, several Slang
workgroup-array tests still run successfully but emit Vulkan validation-layer diagnostics when
`VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation` is enabled.

For example, `tests/bugs/array-size-groupshared.slang -api wgpu` can report
`VUID-StandaloneSpirv-None-10684` from the SPIR-V that Dawn/Tint produces for Vulkan. Slang's
direct SPIR-V path is not the producer in that scenario, so updating the WebGPU runtime/compiler
package is the right layer to test first.

## Proposed solution

Update the packaged Dawn WebGPU binaries from `135.0.7049.80` to `138.0.7204.168` and refresh the
hashes for each supported host package.

The newer Dawn headers also changed the surface acquisition status API. Native Dawn no longer
defines `WGPUSurfaceGetCurrentTextureStatus_Success`; successful acquisition is now represented by
`WGPUSurfaceGetCurrentTextureStatus_SuccessOptimal` or
`WGPUSurfaceGetCurrentTextureStatus_SuccessSuboptimal`. The WASM path already accepted that split
success form, so `SurfaceImpl::acquireNextImage` now uses the same check for native builds.

## Change summary

`CMakeLists.txt` updates `DAWN_VERSION` and the SHA256 hashes for the prebuilt Dawn packages.

`src/wgpu/wgpu-surface.cpp` updates `SurfaceImpl::acquireNextImage` to accept Dawn v138's two
successful `WGPUSurfaceGetCurrentTextureStatus` values.

## Concepts and vocabulary

Dawn is the WebGPU runtime/backend implementation that owns devices, adapters, queues, surfaces,
and native backend selection.

Tint is the shader translator Dawn uses internally for WGSL-to-backend shader lowering.

## Process report

The motivating failure reaches Slang RHI through the `wgpu` backend. Slang emits WGSL, the RHI
passes that WGSL to Dawn with `wgpuDeviceCreateShaderModule`, and Dawn/Tint produces the backend
shader module used by Vulkan. Because the invalid SPIR-V diagnostic is produced after Dawn/Tint
translation, changing Slang's direct SPIR-V emitter would not address the actual producer.

Updating Dawn to `138.0.7204.168` removes the reproduced VUID-10684 diagnostics for the affected
workgroup-array tests. The surface status change is a compile-time compatibility update required
by the v138 header: the old native-only `Success` enumerant is gone, and both `SuccessOptimal` and
`SuccessSuboptimal` still represent a valid acquired surface texture. Failure statuses continue to
return `SLANG_FAIL`.

Validation run from the Slang superproject:

```text
cmake.exe --build --preset debug --target slangc slang-test
VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation build/Debug/bin/slang-test.exe -disable-retries -api wgpu tests/bugs/array-size-groupshared.slang tests/bugs/generic-groupshared.slang tests/compute/groupshared.slang tests/front-end/static-var-in-generic-type.slang tests/metal/groupshared-threadlocal-same-parameter.slang
```

The focused run passed with `100% of tests passed (5/5), 15 tests ignored` and no VUID-10684
messages.


jkwak-work:
@coderabbitai review

skallweitNV (approved):

PR #792, opened 2026-07-03
Resolves an issue where `vkGetPipelineKeyKHR` sometimes return 0

tdavidovicNV:
We see vkGetPipelineKeyKHR returning 0 for clearly different pipelines. This means we have invalid "hash collisions" from getPipelineCacheKey.

This PR adds Slang's program hashes into the key (it hashes the hash, this is a known tradeoff of correctness, speed and simplicity) to make sure getPipelineCacheKey returns different keys even if vkGetPipelineKeyKHR fails to return meaningfully disciminating value.

### DIFF
--- a/src/vulkan/vk-pipeline.cpp
+++ b/src/vulkan/vk-pipeline.cpp
@@ -47,9 +47,57 @@ struct PipelineCacheBinaryHeader
     uint32_t dataOffset;
 };
 
+Result hashSlangEntryPointHash(
+    SHA1& sha1,
+    slang::IComponentType* componentType,
+    uint32_t entryPointIndex,
+    uint32_t targetIndex
+)
+{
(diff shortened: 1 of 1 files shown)


skallweitNV:
Unfortunately I don't think this workaround is good enough. Hashing the program contents alone is not enough, because the pipeline has a lot of additional state baked in (rasterizer state, rt pipeline options etc.) that are missed now. This means if we create multiple pipelines with the same program but different pipeline settings, they all incorrectly map to the same key now. I think the only sensible and correct thing we can do is not cache a pipeline for which we get a zero sized key back. We should probably log that case as a warning on the device, so that we can monitor how often and for what cases this happens. Maybe that will shine some more light on the issue.

tdavidovicNV:
Closing this, as it does not actually address the whole issue.

PR #793, opened 2026-07-03
Fix Vulkan pipeline cache for RT pipelines

skallweitNV:
- Add persistent pipeline-cache coverage for Vulkan ray-tracing pipelines.
- Work around drivers returning unusable all-zero RT pipeline keys by generating a stable fallback hash from:
  - emitted SPIR-V and shader-stage metadata;
  - specialization data;
  - hit-group configuration;
  - relevant RT pipeline settings.
- Disable caching with a warning when an all-zero key is returned and no safe fallback is available.
- Log when pipeline creation from cached binaries fails and compilation is retried.
- Expand compute, render, and RT cache tests to cover distinct programs and pipeline configurations, including cold misses, hot hits, and output validation.

### DIFF
--- a/src/vulkan/vk-pipeline.cpp
+++ b/src/vulkan/vk-pipeline.cpp
@@ -47,22 +47,6 @@ struct PipelineCacheBinaryHeader
     uint32_t dataOffset;
 };
 
-template<typename T>
-void hashValue(SHA1& sha1, const T& value)
-{
-    sha1.update(&value, sizeof(value));
-}
-
-inline void hashString(SHA1& sha1, const char* value)
-{
-    const uint64_t size = value ? strlen(value) : UINT64_MAX;
-    hashValue(sha1, size);
-    if (value)
-    {
-        sha1.update(value, size);
-    }
-}
-
 // Generate a stable ray tracing pipeline key from the inputs owned by slang-rhi. This is used only when
 // vkGetPipelineKeyKHR returns an unusable all-zero key. Keep the version and hashed fields in sync with
 // RayTracingPipelineDesc and the Vulkan ray tracing pipeline creation code.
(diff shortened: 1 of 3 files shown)


src/vulkan/vk-pipeline.cpp
```diff
@@ -47,9 +47,119 @@ struct PipelineCacheBinaryHeader
     uint32_t dataOffset;
 };
 
+template<typename T>
+void hashValue(SHA1& sha1, const T& value)
+{
+    sha1.update(&value, sizeof(value));
+}
+
+inline void hashString(SHA1& sha1, const char* value)
```
tdavidovicNV: This doesn't look really like it should be named hashString. There is already update for a string_view, but this also hashes null string (differently than empty string)

src/vulkan/vk-pipeline.cpp
```diff
@@ -47,9 +47,119 @@ struct PipelineCacheBinaryHeader
...
+        return SLANG_FAIL;
+    hashValue(sha1, moduleCount);
+    for (uint32_t i = 0; i < moduleCount; ++i)
+    {
+        const ShaderProgramImpl::Module& module = program->m_modules[i];
+        const VkPipelineShaderStageCreateInfo& stage = program->m_stageCreateInfos[i];
+        if (!module.code || stage.pNext)
+            return SLANG_E_NOT_AVAILABLE;
+
+        hashValue(sha1, stage.flags);
```
tdavidovicNV: I'd be much happier updating sha1 to have `update<T>(..)` than this thing

src/vulkan/vk-pipeline.cpp
```diff
@@ -47,9 +47,119 @@ struct PipelineCacheBinaryHeader
...
+{
+    if (key.keySize == 0)
+        return true;
+    for (uint32_t i = 0; i < key.keySize; ++i)
+    {
+        if (key.key[i] != 0)
+            return false;
+    }
+    return true;
+}
```
tdavidovicNV, line 147: Funny, I had identical helper with an identical name.

src/vulkan/vk-pipeline.cpp
```diff
@@ -74,7 +184,22 @@ Result getPipelineCacheKey(DeviceImpl* device, void* createInfo, ISlangBlob** ou
...
+        {
+            if (!fallbackPipelineKey)
+            {
+                device->printWarning(
+                    "vkGetPipelineKeyKHR returned an all-zero pipeline key and application didn't provide a fallback "
+                    "key, disabling caching for this pipeline."
+                );
+                return SLANG_E_NOT_AVAILABLE;
+            }
+            sha1.update(fallbackPipelineKey->data(), fallbackPipelineKey->size());
```
tdavidovicNV: aside from hashing a hash (yay), it is slightly awkward that we pay for the fallback key always, even when not using it. Not sure how expensive it ends up being, but I'd rather have it as an actual fallback. Maybe a callback?

ccummingsNV (approved):

PR #797, opened 2026-07-16
Improve D3D12 timestamp query resolving

skallweitNV:


skallweitNV:
this needs more work (after my vacation)

PR #801, opened 2026-07-23
Implement native Metal buffer import

fknfilewalker:
This PR adds buffer imports to the Metal backend, which is required to support importing MPS-based PyTorch tensors in slangpy. Multiple references to the same Buffer are supported.


jkwak-work (commented):
Looks good to me but I will wait for @skallweitNV to review.

src/metal/metal-buffer-address-map.h
```diff
@@ -54,8 +66,8 @@ class BufferAddressMap
 private:
     struct Entry
     {
-        BufferImpl* buffer;
-        DeviceAddress size;
+        std::vector<BufferImpl*> buffers;
```
skallweitNV: This feels kinda expensive, adding lots of heap allocations where previously there were none.
You should discuss with @jhelferty-nv who originally authored this code.

skallweitNV (changes requested):
LGTM apart from the mentioned overhead which would be nice to optimize.

src/metal/metal-buffer-address-map.h
fknfilewalker: I changed it to a chain (linked list), otherwise a small_vector could be used too.

skallweitNV (approved):
LGTM

PR #811, opened 2026-08-05
Do not cache shader object layouts keyed on a caller-supplied type layout

jvepsalainen-nv:
## Motivation

`Device::m_shaderObjectLayoutCache` (`src/device.h`) is keyed on a raw `slang::TypeLayoutReflection*` and lives as long as the `Device`. Two entry points reach it:

| entry point | key origin | actual owner | reference held by the entry | safe |
| --- | --- | --- | --- | --- |
| `createShaderObject(session, type, container)` | `session->getTypeLayout(type)` | `Linkage` / `ISession` | `ComPtr<ISession>` | yes |
| `createShaderObjectFromTypeLayout(typeLayout)` | caller-supplied | `TargetProgram` / `ComponentType` | `ComPtr<ISession>` of the *device's* session — the wrong object | **no** |

On the second path the key belongs to an object slang-rhi holds no reference to and knows nothing about. The entry outlives the layout as soon as the caller releases its program, and once an unrelated allocation lands on the recycled address the next lookup returns a `ShaderObjectLayout` whose `m_elementTypeLayout` dangles. `ShaderObject::init` then dereferences it.

That is shader-slang/slang#10893, which the Slang nightly sanitizer job hits intermittently:

```
SUMMARY: AddressSanitizer: heap-buffer-overflow source/core/slang-list.h:95:25 in Slang::List<Slang::TypeLayout::ResourceInfo, Slang::StandardAllocator>::begin()
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior source/slang/slang-reflection-api.cpp:1401:29
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior source/slang/slang-type-layout.h:772:25
```

Concretely, `tools/render-test/render-test-main.cpp` takes this path whenever a test input gives no explicit type name, e.g. `//TEST_INPUT:set gParams = new{ values: [ new ImplA{ v: 3 }, new ImplB{ v: 5 } ] }` in `tests/language-feature/dynamic-dispatch/constantbuffer-struct-interface-array.slang`. Devices persist across tests via a process-lifetime cache, so entries accumulate for the life of the process, and the CPU backend never clears `m_shaderObjectLayoutCache` at all (only the d3d12, vulkan and wgpu destructors do).

## Proposed solution

Build the layout directly on that path instead of going through the cache.

This is correct by construction rather than by argument: slang-rhi cannot vouch for the lifetime of a pointer a caller handed it, so it must not retain one past the call. It also keeps every remaining cached key session-owned, which promotes the existing `ComPtr<ISession>` from incidental to load-bearing.

The alternative — extending the public API so the caller passes the owning `IComponentType*` and holding a reference to *that* — would preserve caching, but it is an ABI break, and the measurements below show the cache was providing no reuse on this path to preserve.

Two options were considered and rejected:

- **Keying on a stable identity such as `ShaderComponentID`.** `ShaderCache::getComponentId` builds its key from `type->getName()` plus specialization argument names, so distinct types sharing a name across modules, and anonymous types, collide. A collision returns the *wrong* layout rather than crashing — worse than the present bug.
- **Validating or evicting entries on lookup.** There is no way to test a raw pointer for staleness, and Slang exposes no program- or session-destruction hook to subscribe to.

## Change summary

| file | change |
| --- | --- |
| `src/device.cpp` | `createShaderObjectFromTypeLayout` calls `createShaderObjectLayout` directly; adds the `testing::getShaderObjectLayoutCacheSize` implementation |
| `src/device.h` | declares `testing::getShaderObjectLayoutCacheSize`, alongside the existing `gDebugDisableStateTracking` test hook |
| `tests/test-shader-object-from-type-layout.cpp` | regression test asserting the invariant |
| `tests/test-shader-object-from-type-layout.slang` | minimal shader with a `ConstantBuffer<Params>` global |
| `CMakeLists.txt` | registers the new test |

## Concepts and vocabulary

- **`TargetProgram` vs `Linkage` ownership.** `Linkage` owns `List<RefPtr<TargetRequest>> targets`, and each `TargetRequest` owns `Dictionary<TypeLayoutKey, RefPtr<TypeLayout>> typeLayouts`. So a layout obtained via `ISession::getTypeLayout` lives exactly as long as the session. A layout reached through `IComponentType::getLayout()` instead belongs to the `TargetProgram`, which is tied to a `ComponentType` — releasing the program frees those `TypeLayout` objects however long the session lives. This distinction is the whole of the bug.
- **Dangling entry vs stale lookup.** A cache entry whose key has been freed is *dangling*; that alone is harmless until an unrelated allocation reuses the address and a lookup returns the entry — the *stale lookup*, which is the crash. The former happens deterministically, the latter only when the allocator cooperates. The measurements below rely on that separation.

## Process report

**Why this layer.** The defect is that a cache entry outlives the object owning its key. slang-rhi is where the entry is created, and it is the only place that can know it has no claim on the key's lifetime, so the fix belongs here rather than in Slang. Nothing in Slang is malformed: `TypeLayoutReflection` is a valid pointer that its owner is entitled to free.

**Is the input shape correct?** Yes — a caller passing a program-owned type layout to `createShaderObjectFromTypeLayout` is using the API exactly as documented. The API takes a `slang::TypeLayoutReflection*` and says nothing about ownership, so a program-owned layout is legitimate input. What was wrong was the callee's assumption that it could retain the pointer. That is why the fix changes the retention, not the caller and not the accepted shape.

**Why not fix the caller instead.** `tools/render-test/render-test-main.cpp` in Slang could be changed to keep its program alive, but that only patches the one in-tree caller; `createShaderObjectFromTypeLayout` is public API, and any downstream caller doing the same legitimate thing would still corrupt the device's cache.

**Evidence that the split above is real, not just argued.** Instrumenting `Device::getShaderObjectLayout` to record every insert with its entry point, then testing recorded keys with `__asan_region_is_poisoned` while the cache is in use, over the full Slang test suite at `ff45b15ed3`:

```
createShaderObject                 inserts=49 dangling=0
createShaderObjectFromTypeLayout   inserts=38 dangling=36
```

Zero of 49 on the session-keyed path; 36 of 38 on the caller-supplied path. The remaining 2 are the final inserts of the run, freed after the last sweep. After this change the caller-supplied path no longer appears in the cache at all, and the Slang suite is unchanged at 6374 passed / 0 failed.

**Cost of losing the cache on this path.** Counting calls and distinct keys over the same suite:

```
createShaderObjectFromTypeLayout: calls=38 distinct-keys=38
 -> 0 lookup(s) would have hit the cache.
```

Every call presented a distinct key, so the cache never served a hit here. This measures render-test's usage only; a downstream caller could exercise the entry point far more heavily, in which case `createShaderObjectLayout` cost would be worth revisiting.

**Why the test asserts the invariant rather than the crash.** The use-after-free is only observable when the allocator recycles a freed address, so a test for the symptom is really a test for "the sanitizer happened to stay quiet" — which is how these findings were previously written off as fixed, and why the issue then resurfaced. The test instead checks that `createShaderObjectFromTypeLayout` leaves the cache size unchanged, which fails deterministically on any platform, with or without a sanitizer. Verified both ways on the CPU backend: `PASSED` with the fix, `CHECK_EQ( 1, 0 )` / `FAILED` with it reverted.

**Why a test hook.** The invariant is about internal state with no public accessor. `src/device.h` already exposes `gDebugDisableStateTracking` and `gResourceCount` in `namespace testing` for the same reason, so `getShaderObjectLayoutCacheSize` follows the established pattern rather than widening the public API.

**Note for reviewers.** `tests/testing.cpp:902` blanket-skips the CPU backend on Linux ("Known issues with CPU backend on linux"), so this test only runs there on the other backends. The pass/fail verification above was done by lifting that gate locally; that change is deliberately not included here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1067,6 +1067,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED AND NOT EMSCRIPTEN)
         tests/test-sampler.cpp
         tests/test-sha1.cpp
         tests/test-shader-cache.cpp
+        tests/test-shader-object-from-type-layout.cpp
         tests/test-shader-object-large.cpp
         tests/test-shader-object-resource-tracking.cpp
         tests/test-ring-queue.cpp
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -12,6 +12,11 @@ namespace rhi {
 namespace testing {
 bool gDebugDisableStateTracking = false;
 std::atomic<uint64_t> gResourceCount{0};
+
+size_t getShaderObjectLayoutCacheSize(IDevice* device)
+{
+    return checked_cast<Device*>(device)->m_shaderObjectLayoutCache.size();
+}
 } // namespace testing
 
 // ----------------------------------------------------------------------------
@@ -784,8 +789,25 @@ Result Device::createShaderObject(
 
 Result Device::createShaderObjectFromTypeLayout(slang::TypeLayoutReflection* typeLayout, IShaderObject** outObject)
 {
+    // Build the layout directly rather than going through m_shaderObjectLayoutCache.
+    //
+    // That cache is keyed on a raw `slang::TypeLayoutReflection*` and lives as
+    // long as the `Device`. Every other way in supplies a key obtained from
+    // `ISession::getTypeLayout`, which the `Linkage` owns, so the entry's
+    // `ComPtr<slang::ISession>` genuinely covers the key's lifetime. Here the key
+    // comes from the caller, and in practice it is a layout owned by a
+    // `TargetProgram` -- an object slang-rhi holds no reference to and knows
+    // nothing about. Caching it leaves an entry that outlives the layout as soon
+    // as the caller releases its program, and a later allocation landing on the
+    // recycled address turns the next lookup into a use-after-free
+    // (shader-slang/slang#10893).
+    //
+    // slang-rhi cannot vouch for the lifetime of a pointer it was handed, so it
+    // must not retain one past the call. Skipping the cache here keeps every
+    // cached key session-owned, which makes the existing session reference
+    // load-bearing rather than incidental.
     RefPtr<ShaderObjectLayout> shaderObjectLayout;
-    SLANG_RETURN_ON_FAIL(getShaderObjectLayout(m_slangContext.session, typeLayout, shaderObjectLayout.writeRef()));
+    SLANG_RETURN_ON_FAIL(createShaderObjectLayout(m_slangContext.session, typeLayout, shaderObjectLayout.writeRef()));
     RefPtr<ShaderObject> shaderObject;
     SLANG_RETURN_ON_FAIL(createShaderObject(shaderObjectLayout, shaderObject.writeRef()));
     returnComPtr(outObject, shaderObject);
--- a/src/device.cpp
+++ b/src/device.cpp
@@ -3,6 +3,7 @@
 #include "rhi-shared.h"
 #include "shader.h"
 #include "heap.h"
+#include "debug-layer/debug-device.h"
 
 #include <algorithm>
 #include <cstdarg>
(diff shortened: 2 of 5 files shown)


jvepsalainen-nv:
Removed review requests for now. Work in progress.

skallweitNV (approved):

skallweitNV:
@jvepsalainen-nv just noticed now that this PR added a bunch of very extensive comment blocks. Can you see if they can be cleaned up a bit?

PR #813, opened 2026-08-05
Apply fixupBufferDesc when importing buffers from native handles (vulkan/d3d)

fknfilewalker:
createBufferFromNativeHandle skipped fixupBufferDesc on the Vulkan and D3D12 backends, so an imported buffer kept defaultState == Undefined while every other buffer got a real state from determineDefaultResourceState(usage). Metal and WGPU already fix up the desc on import.


jkwak-work (commented):
Looks good to me.
But I want @skallweitNV to review, because he is more familiar with this code lines.

skallweitNV (approved):

PR #817, opened 2026-08-09
Derive Vulkan surface supportedUsage from surface capabilities

fknfilewalker:
supportedUsage was a hardcoded set plus UnorderedAccess inferred from the preferred format alone. The preferred format is always one of the *_SRGB formats, which never support storage, so UnorderedAccess was never advertised and the debug layer rejected storage usage even for formats that genuinely support it (e.g. RGBA8Unorm).

Query `VkSurfaceCapabilitiesKHR::supportedUsageFlags` instead, the format-independent surface constraint. Format-dependent restrictions remain validated against the chosen format in configure().

### DIFF
--- a/src/vulkan/vk-surface.cpp
+++ b/src/vulkan/vk-surface.cpp
@@ -390,7 +390,8 @@ Result SurfaceImpl::configure(const SurfaceConfig& config)
         // format (e.g. *_SRGB), tripping VUID-VkSwapchainCreateInfoKHR-imageFormat-01778. Apps
         // that need storage on the swapchain must request it explicitly; configure() then
         // validates it against the format below.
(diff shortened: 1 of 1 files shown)


jhelferty-nv:

Auto-assigned @jkwak-work as shepherd for this Community PR.

FYI for maintainers: committer signal on the changed files is higher for skallweitNV than for the auto-requested reviewer (jkwak-work). They were not auto-requested; a human may optionally add them as a reviewer.

skallweitNV (approved):
LGTM, I will follow this up with some additional cleanup across the various backeds and validation layer.

fknfilewalker:
> LGTM, I will follow this up with some additional cleanup across the various backeds and validation layer.

Sounds good. Yes, maybe the api needs a bit of a redesign.

PR #823, opened 2026-08-10
Compile deferred pipeline shaders in parallel

skallweitNV:
## Summary

This is the third change in the parallel pipeline compilation series, following #822.

It adds task-parallel entry-point code generation when resolving a batch of deferred pipelines. Backend shader-module installation, pipeline object creation, cache publication, and command patching remain serialized.

## Changes

- Add `PipelineCompilationMode`:
  - `Serial` - use sequential pipeline resolution (default)
  - `Parallel`, allow deferred pipeline resolution to use available task parallelism (experimental)
- Replace the single-pass resolver with a phased pipeline resolver:
  1. Collect and deduplicate pipeline requests.
  2. Resolve specialized programs and prepare entry-point compilation.
  3. Generate prepared entry-point code concurrently.
  4. Install backend shader modules serially.
  5. Create backend pipeline objects serially.
  6. Publish caches and patch command-list references serially.
- Deduplicate:
  - repeated uses of the same pipeline;
  - equivalent specialized pipeline requests;
  - multiple pipelines sharing the same shader program.
- Acquire each program’s compilation lock before preparation and hold it through backend module installation.
- Serialize resolution and cache publication across command encoders on the same device.
- Wait for all submitted compilation tasks before reporting diagnostics or returning a compilation failure.
- Keep the CPU backend on the serial path.
- Avoid task-pool overhead when a batch contains only one entry point.
- Add testing support for selecting the pipeline compilation mode.

## Compilation reporting

Slang currently exposes cumulative global compiler timing. Those values cannot be attributed accurately to individual entry points compiled concurrently.

Parallel entry-point compilation therefore continues to report diagnostics, cache status, output size, and wall-clock timestamps, but does not report per-entry-point compiler and downstream times.

## Scope

This PR parallelizes shader code generation only.

Backend pipeline object creation remains serial and will be enabled selectively in a follow-up PR after each backend’s concurrency requirements have been reviewed.

## Tests

The new deferred-pipeline test exercises:

- Multiple independent compute programs.
- Multiple entry points from one linked graphics program.
- Repeated use of the same pipeline.
- Multiple pipelines sharing one shader program.
- The explicit serial compatibility mode.
- Result validation after command submission.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1047,6 +1047,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED AND NOT EMSCRIPTEN)
         tests/test-nvrtc.cpp
         tests/test-nvapi.cpp
         tests/test-offset-allocator.cpp
+        tests/test-parallel-pipeline-creation.cpp
         tests/test-pipeline-cache.cpp
         # tests/test-precompiled-module-cache.cpp
         tests/test-precompiled-module.cpp
--- a/include/slang-rhi.h
+++ b/include/slang-rhi.h
@@ -3266,6 +3266,15 @@ enum class AftermathFlags
 };
 SLANG_RHI_ENUM_CLASS_OPERATORS(AftermathFlags);
 
+enum class PipelineCompilationMode
+{
+    /// Use sequential pipeline resolution.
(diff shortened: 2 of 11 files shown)


skallweitNV:
@coderabbitai review

PR #841, opened 2026-08-15
Add ResourceHeap API for memory aliasing

EvanLuo42:
## Summary

Adds a ResourceHeap API so applications can place buffers and textures into shared backing memory and alias overlapping resources across D3D12, Vulkan, Metal, and CUDA.

## Motivation

Today every buffer and texture allocates its own committed memory. That works, but it wastes GPU memory when resource lifetimes do not overlap — for example transient render targets, scratch buffers, and upload/readback staging that are only live for part of a frame.

Native APIs already expose this:

- D3D12 placed resources on an `ID3D12Heap`
- Vulkan `vkBindBufferMemory` / `vkBindImageMemory`
- Metal placement heaps
- CUDA suballocation from a device or host-pinned block

slang-rhi had no corresponding API, so higher-level users (SlangPy / SGL allocators, custom engines) could not do explicit placement or memory aliasing without dropping to native handles. This PR adds that missing layer.

The intended workflow is:

1. Create an `IResourceHeap` of a given size, memory type, and kind.
2. Query `getBufferMemoryRequirements` / `getTextureMemoryRequirements` for size and alignment.
3. Create the resource with a `ResourcePlacementDesc` chained onto `BufferDesc::next` / `TextureDesc::next`.
4. When two placed resources overlap, record `ICommandEncoder::aliasResources(before, after)` before using the newly activated resource.

`Feature::MemoryAliasing` reports whether the device supports this path.


jhelferty-nv:

Auto-assigned @jkiviluoto-nv as shepherd for this Community PR.

FYI for maintainers: committer signal on the changed files is highest for skallweitNV among collaborators other than the assignee. They were not auto-requested; a human may optionally add them as a reviewer.

src/d3d12/d3d12-command.cpp
```diff
@@ -1569,6 +1571,37 @@ void CommandRecorder::cmdSetTextureState(const commands::SetTextureState& cmd)
...
+        m_stateTracking.setBufferState(buffer, buffer->m_desc.defaultState);
+    if (Texture* texture = asTexture(cmd.before))
+        m_stateTracking.setTextureState(texture, kEntireTexture, texture->m_desc.defaultState);
+    commitBarriers();
+
+    auto getResource = [](IResource* resource) -> ID3D12Resource*
+    {
+        if (Buffer* buffer = asBuffer(resource))
+            return checked_cast<BufferImpl*>(buffer)->m_resource.getResource();
+        if (Texture* texture = asTexture(resource))
```
jkwak-work: Would it make sense to have `else if`?
```
if (Buffer* buffer = asBuffer(resource))
  ...;
else if (Texture* texture = asTexture(resource))
```

src/vulkan/vk-texture.cpp
```diff
@@ -29,7 +30,8 @@ TextureImpl::~TextureImpl()
     }
     if (m_shouldDestroyImage)
     {
-        api.vkFreeMemory(api.m_device, m_imageMemory, nullptr);
+        if (m_ownsMemory)
+            api.vkFreeMemory(api.m_device, m_imageMemory, nullptr);
```
jkwak-work, line 34: It looks like this memory will never be freed once m_ownsMemory is set to false for any reason.
It may leak the memory as it is.

jkwak-work (commented):
I think the implementation is mostly fine.
But I am missing what this PR is for.

@EvanLuo42 , can you update the PR description to provide some background information about what this PR is about?

src/vulkan/vk-texture.cpp
EvanLuo42: This should be okay. Heap-placed images set `m_ownsMemory` to false because `m_resourceHeap` keeps the `ResourceHeapImpl` alive, and the heap’s destructor is responsible for freeing the `VkDeviceMemory`.

EvanLuo42:
> @EvanLuo42 , can you update the PR description to provide some background information about what this PR is about?

@jkwak-work Sure! Updated.

src/d3d12/d3d12-device.cpp
```diff
@@ -907,7 +908,10 @@ Result DeviceImpl::initialize(const DeviceDesc& desc, BackendImpl* backend)
             {
                 addFeature(Feature::Bindless);
             }
+
+            m_resourceHeapTier = options.ResourceHeapTier;
         }
+        addFeature(Feature::MemoryAliasing);
```
jkwak-work, line 914: Shouldn't we add this feature only when tier-2 is supported?
```
if (options.resourceHeapTier >= D3D12_RESOURCE_HEAP_TIER_2)
    addFeature(Feature::MemoryAliasing);
```

src/vulkan/vk-texture.cpp
jkwak-work: @nv-slang-bot , can you investigate if this change can cause a memory leak.
I don't see how m_imageMemory can be released once m_ownsMemory turns false.

src/d3d12/d3d12-device.cpp
EvanLuo42: My understanding is that Resource Heap Tier 2 only affects whether mixed buffer/texture heaps are supported (https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_heap_tier). On Tier 1, same-kind heaps, placed resources, and aliasing are still valid.

Since `ResourceHeapKind::All` is already gated separately, I think it makes sense for `Feature::MemoryAliasing` to continue representing the availability of the heap + placement + aliasing path. Otherwise, Tier 1 devices would lose support for working same-kind aliasing.

So I’d prefer to keep these two layers separate.

jkwak-work:
This looks good to me as far as I can tell.
But I will ask @skallweitNV to review and approve.

skallweitNV (changes requested):
@EvanLuo42  thanks for this contribution/ Explicit resource heaps and aliasing are valuable additions, and the implementation is meaningfully cross-backend. Before merging, we need to tighten the public contract and address the following correctness issues.

A lot of this analysis was written by AI and quickly reviewed by me. There might requests for changes in here that are worth discussing further before acting on them.

## Required API changes

1. **Make `ResourceMemoryRequirements` extensible.**

   Add `structType` and `next`. This structure will likely need additional allocation constraints in the future, and changing its layout later would be difficult.

   Replace `requiresDedicatedAllocation` with a flags field, initially containing `RequiresDedicatedAllocation`, leaving room for constraints such as `PrefersDedicatedAllocation`.

2. **Represent heap alignment separately from placement alignment.**

   Add `heapAlignment` to `ResourceMemoryRequirements`, and either:

   - Add `alignment` to `ResourceHeapDesc`, or
   - Derive the heap alignment from requirements supplied during heap creation.

   Resource-offset alignment and native heap alignment are not always identical; D3D12 MSAA resources require 4 MiB heap alignment.

3. **Do not treat `ResourceHeapKind` as the complete compatibility contract.**

   The buffer/non-RT/RT classification describes allowed resource categories, but it does not capture Vulkan memory-type compatibility.

   We propose:

   - Rename it to `ResourceHeapUsage`.
   - Make it a bitmask (`Buffers`, `NonRtDsTextures`, `RtDsTextures`).
   - Carry separate opaque, device-local compatibility information in `ResourceMemoryRequirements`.
   - Make heap creation consume one or more requirements, or provide another mechanism that lets the backend choose a heap compatible with the intended resources.
   - Add an `isResourceHeapCompatible(heap, requirements)` query for persistent heap pools.

   Required invariant: if a heap is created for a set of requirements, every corresponding resource must be placeable in that heap at a valid aligned offset on every supported backend.

4. **Clarify placement descriptor lifetime.**

   `ResourcePlacementDesc` should be documented as a creation-only extension. At minimum, strip it from the descriptor stored by the created resource.

   Currently `Buffer` and `Texture` copy `desc.next`, which can leave `getDesc().next` pointing to a caller-owned stack object after creation returns. The created resource should retain the heap itself, but not the caller’s chained structure.

5. **Clarify capability reporting.**

   A single `Feature::MemoryAliasing` flag covers substantially different support levels. Either split resource-heap support from aliasing support, or add resource-heap capability properties describing:

   - Supported resource usage classes.
   - Supported logical memory types.
   - Mixed buffer/texture support.
   - Required buffer/image granularity.

   Per-resource requirements queries should continue returning `SLANG_E_NOT_AVAILABLE` when that resource cannot be placed.

## Required correctness fixes

6. **Fix D3D12 MSAA heap creation.**

   [`d3d12-resource-heap.cpp`](/C:/src/slang-rhi/src/d3d12/d3d12-resource-heap.cpp:51) always creates 64 KiB-aligned heaps. MSAA heaps require 4 MiB alignment. Add MSAA coverage for sampled, render-target, and `All` heaps.

7. **Fix D3D12 placed-buffer initialization.**

   [`CopyBufferRegion`](/C:/src/slang-rhi/src/d3d12/d3d12-device.cpp:1700) currently writes to the placed buffer without ensuring that it is in `COPY_DEST`.

   Create or transition the resource to `COPY_DEST`, perform the upload, and then transition it into its requested/default initial state. Run this under native D3D12 validation.

8. **Use the exact same Vulkan resource description for querying and creation.**

   Factor `VkImageCreateInfo` construction into one helper used by both paths. The requirements query currently differs from actual creation for:

   - Typeless/mutable-format textures.
   - 3D render targets requiring 2D-array-compatible views.
   - Usage flags added because initial data is present.

   A query must describe the exact native resource that will later be created.

9. **Handle Vulkan memory compatibility and granularity correctly.**

   Do not derive heap compatibility solely from a few representative buffer/image objects. Heap creation must select a memory type compatible with the actual supplied requirements.

   Also account for `bufferImageGranularity` when buffers and optimal images are placed at adjacent, non-overlapping offsets.

## Aliasing semantics and validation

10. **Specify the synchronization contract for `aliasResources`.**

    Document that:

    - Earlier accesses to `before` happen before later accesses to `after` on the same queue.
    - `before` becomes inactive and its tracked state is discarded.
    - `after` becomes active with undefined contents.
    - Cross-queue synchronization remains the caller’s responsibility.
    - The permitted meanings of null `before` and null `after` are explicit.

11. **Strengthen debug-layer validation.**

    When both resources are provided, validate that they:

    - Belong to the same device.
    - Are placed resources.
    - Refer to the same heap.
    - Have overlapping memory ranges.
    - Have mutually compatible allocation requirements.

    Invalid calls should produce a clear validation error rather than relying on backend behavior.

12. **Avoid redundant D3D12 transitions.**

    The current alias test produces a native warning because the same subresource is transitioned multiple times in one barrier batch. Commit or coalesce the pending transition before emitting the aliasing transition/barrier.

## Required tests

Please add tests for:

- MSAA heap requirements and placement.
- D3D12 placed-buffer initialization into CopySource and UAV-default buffers.
- Vulkan typeless textures.
- Vulkan 3D render targets.
- Vulkan texture initial data when `CopyDestination` was not explicitly requested.
- Adjacent buffer and texture allocations respecting buffer/image granularity.
- Invalid alias calls: different devices, different heaps, committed resources, and non-overlapping ranges.
- A synchronization sequence in one command buffer with no host wait:

  `write A → alias A to B → write/read B`

- The same sequence across submissions without a host wait, using only queue ordering.
- Heap lifetime after the caller releases its heap reference.

Batching multiple alias barriers can be a follow-up, but the compatibility model, extensibility, ownership rules, synchronization contract, and correctness fixes should be resolved before this API becomes public.

skallweitNV:
@EvanLuo42 thanks for the changes. The API and cross-backend consistency look substantially better. I think the remaining merge blockers are focused implementation fixes:

1. **Stop tracking the deactivated Vulkan alias.** `cmdAliasResources()` currently resets both resources to `Undefined`, leaving both in the state tracker. At command-buffer close, `requireDefaultStates()` consequently transitions the inactive resource too. I confirmed this sequence: write A → alias A to B → write B → closing the command buffer transitions B, then A. A layout transition on A can modify memory now owned by B. Please forget the state of `before`, while resetting `after` to `Undefined`. Add a texture regression covering this sequence without intermediate host waits, and verify that no transition is emitted for A after its deactivation.

2. **Avoid touching placed Vulkan texture memory during creation without initial data.** `createTexture()` currently performs an initial layout transition even for a placed texture with no initial data. Creating B over memory currently used by A can therefore affect A before the application activates B. Please defer that transition until activation/first use, keeping the state tracker consistent with the actual layout. Also document how creation with initial data interacts with alias activation. Add a regression that writes A, creates overlapping B without initial data, and verifies A remains intact before the alias switch.

3. **Honor or reject explicit heap alignment.** CUDA currently accepts and reports the requested heap alignment without ensuring the allocation satisfies it. I confirmed that a heap requesting 1 GiB alignment succeeds with a native address that is not aligned accordingly. Please either provide the requested alignment or reject unsupported values. Audit the equivalent Metal and Vulkan paths as well, and add coverage for an explicit alignment beyond the allocator’s default guarantee.

With these fixes and regression tests in place, I don’t currently see a need for another broad public-API redesign.

EvanLuo42:
@skallweitNV Thanks, these should be addressed in the latest commit.

skallweitNV:
@EvanLuo42  AI review did surface another issue and I think it's correct. The resource's reference to the ResourceHeap introduces a cyclic reference. I haven't fully thought through this, so take this with a grain of salt. Anyway, here goes the analysis:

### Resource heaps introduce a reference cycle during deferred deletion

Placed resources retain their `ResourceHeap`, and the heap retains its creating device. When a resource is released, its destruction is deferred through the device’s command queue.

The existing deferred-deletion path breaks the resource’s **direct** strong reference to the device. However, the new heap reference preserves an indirect reference:

```text
Device
  → Command queue
    → Resource awaiting deferred deletion
      → ResourceHeap
        → Device
```

Consequently, dropping all application references may leave the device, queue, resource, heap, and backing allocation alive indefinitely. Device destruction would drain the queue, but the heap’s device reference prevents destruction from starting.

Relevant code:

- [Resource retains its placement heap](/C:/src/slang-rhi/src/rhi-shared.h:86)
- [Heap device-reference hooks](/C:/src/slang-rhi/src/resource-heap.h:28)
- [D3D12 deferred deletion breaks only the direct device reference](/C:/src/slang-rhi/src/d3d12/d3d12-device.cpp:2507)

### Reproduction

No GPU submission is necessary:

1. Create a device.
2. Create a resource heap and a placed buffer.
3. Release the buffer, heap, and device.
4. Check whether the objects were destroyed.

Repeated runs retained **14 objects per iteration on D3D12, 8 on Vulkan, and 10 on CUDA**. The equivalent committed-buffer control returned to the baseline after initial setup.

The normal tests can mask this because later queue activity drains deferred resources.

### Potential solution

The ownership model needs to distinguish **keeping the heap allocation alive** from **keeping the device alive**.

One approach is to integrate resource-to-heap references into the existing internal/external reference model:

- An application-held heap must keep its device alive.
- A live resource already keeps its device alive.
- A resource awaiting deferred deletion must keep the heap’s backing allocation alive until GPU completion, without indirectly retaining the device.

`ResourceHeap::makeInternal()` and `makeExternal()` already provide hooks, but ordinary `RefPtr` placement references do not establish the bookkeeping needed to break this cycle. Any such bookkeeping must handle multiple resources and concurrent reference changes correctly.

Another approach is to separate the native heap allocation into a backing object retained by resources, with device lifetime managed separately by the public heap and resource objects.

**Simply dropping the heap reference when deferring the resource is unsafe:** it could free backing memory while GPU work still uses it.

### Regression coverage

Add a test using a fresh, uncached device that releases a placed resource, heap, and device **without a subsequent submission or explicit cleanup call**, then verifies destruction. Also cover:

- Different release orders.
- Multiple resources sharing a heap.
- Releasing application references while GPU work is pending.

The fix must eliminate the ownership cycle while preserving backing memory until GPU completion.

EvanLuo42:
@skallweitNV I have confirmed the bug and fixed it. Since the fix involve editing `RefObject` so I also added tests for `RefObject`. I also found another cyclic reference bug and opened an issue #876.

skallweitNV:
@EvanLuo42 I don't have time to review this properly today. How much of this code was written by AI?

EvanLuo42:
@skallweitNV I wrote the regression tests and the new `RefObject` APIs. AI was used to wire the new APIs across backends.

PR #843, opened 2026-08-26
Fix: handle PushConstant binding ranges in bindAsValue

konakona418:
`bindAsValue` iterates over every binding range of a shader object, and its two switches have no case for `SlangBindingType::PushConstant`. Any shader with a `[vk::push_constant]` parameter therefore falls through to the default branch and trips `SLANG_RHI_ASSERT_FAILURE`, both in the descriptor write loop and in the sub-object loop.

Push constant data is carried in the shader object's uniform data and written via vkCmdPushConstants, so no descriptor work is needed here, the same as ConstantBuffer and ParameterBlock.

### DIFF
--- a/src/vulkan/vk-shader-object.cpp
+++ b/src/vulkan/vk-shader-object.cpp
@@ -449,6 +449,7 @@ Result BindingDataBuilder::bindAsValue(
         case slang::BindingType::ConstantBuffer:
         case slang::BindingType::ParameterBlock:
         case slang::BindingType::ExistentialValue:
+        case slang::BindingType::PushConstant:
             break;
 
         case slang::BindingType::Texture:
@@ -656,7 +657,9 @@ Result BindingDataBuilder::bindAsValue(
             break;
         case slang::BindingType::RawBuffer:
         case slang::BindingType::MutableRawBuffer:
-            // No action needed for sub-objects bound though a `StructuredBuffer`.
+        case slang::BindingType::PushConstant:
+            // No action needed for sub-objects bound though a `StructuredBuffer`
+            // or a push constant.
             break;
         default:
             SLANG_RHI_ASSERT_FAILURE("Unsupported sub-object type");


jhelferty-nv:

Auto-assigned @jkiviluoto-nv as shepherd for this Community PR.

FYI for maintainers: committer signal on the changed files is higher for skallweitNV than for the auto-requested reviewer (jkiviluoto-nv). They were not auto-requested; a human may optionally add them as a reviewer.

jkwak-work (commented):
It looks good to me.
But I like to ask @skallweitNV to approve

skallweitNV (changes requested):
Handling `PushConstant` in the descriptor-range switch is correct, but treating it as a no-op in the sub-object switch is not.

`[[vk::push_constant]] ConstantBuffer<T>` is an explicitly supported Slang construct. The [[Slang documentation](https://docs.shader-slang.org/en/stable/parameter-blocks.html#push-constant-ranges)](https://docs.shader-slang.org/en/stable/parameter-blocks.html#push-constant-ranges) states that it reflects as a `BindingType::PushConstant` sub-object range. Upstream compiler tests also use this exact form, including the [[SPIR-V layout test](https://github.com/shader-slang/slang/blob/master/tests/spirv/push-constant-layout.slang#L10-L18)](https://github.com/shader-slang/slang/blob/master/tests/spirv/push-constant-layout.slang#L10-L18).

For such an explicit global push constant, the data lives in the child shader object’s `m_data`. It does not pass through `bindAsEntryPoint()`, which is currently the only path that appends data to `pushConstantData` for `vkCmdPushConstants`. Therefore, the new sub-object `break` silently skips the object containing the push-constant bytes.

There is also an earlier layout issue that this PR does not address: `addBindingRanges()` does not count `PushConstant` ranges in `m_subObjectCount`. A focused runtime test using the documented explicit form still crashes on this PR during root shader-object construction because reflection reports one push-constant sub-object while the Vulkan layout allocates zero sub-object slots.

The relationship to [[slang#11147](https://github.com/shader-slang/slang/issues/11147)](https://github.com/shader-slang/slang/issues/11147) is also unclear. That issue says the offending `BindingType` was unknown, and the affected shader contains no explicit push constant or `ConstantBuffer`. Please provide a reproducer or captured `"Unsupported binding type: 14"` output demonstrating that this patch addresses that failure.

I think the first addition is directionally correct because push constants do not require descriptor writes. The sub-object case needs complete handling instead:

- count and instantiate `PushConstant` sub-objects;
- copy their ordinary data into the corresponding `VkPushConstantRange`;
- recurse into contained bindings;
- add an explicit Vulkan runtime test.

As written, this suppresses an assertion without correctly binding the documented explicit push-constant form.

skallweitNV:
Opened https://github.com/shader-slang/slang-rhi/pull/848 to fix this plus the push constant support for constant buffers.

skallweitNV:
@konakona418 let me know if think there is still open follow-up after #848 merged, otherwise lets close this.

konakona418:
Thank you for the explanation. Closing the PR.

PR #846, opened 2026-08-27
Share constant buffer pages across command buffers

### LINKED ISSUE #844: Per-command-buffer 4 MiB constant-buffer pools cause multi-GiB retention and D3D12 binding failures
## Summary

Each D3D12 command buffer owns a `ConstantBufferPool` whose first allocation creates a 4 MiB upload buffer. Retired command buffers are cached by `CommandQueueImpl`, and `ConstantBufferPool::reset()` resets only the page index/offset; it does not release the allocated pages.

Workloads that submit many small command buffers can therefore retain several GiB of upload buffers, command allocators, and command lists. On D3D12 this eventually produces repeated:

```text
[ERROR] (rhi) layer: Failed to get binding data
```

and subsequent device/readback failures.

This was originally reported as shader-slang/slangpy#665.

## Reproduction

A headless two-file SlangPy reproduction is available here:

https://gist.github.com/kaizhangNV/b817c7ce7aa84bf2810acbbe4bfcfc60

It has no window, swapchain, or image dependencies. With a current SlangPy build and NumPy installed:

```powershell
python issue665-headless-repro.py
```

The training loop performs 250 iterations per frame. Each iteration submits one differentiation call plus seven optimizer calls and seven tensor clears, for approximately 3,750 command-buffer submissions per frame. About 2,000 of those command buffers allocate ordinary uniform data and therefore instantiate a constant-buffer page.

## Observed result

On the tested D3D12 system, frames 0-3 complete. During frame 4 the process begins emitting `Failed to get binding data` repeatedly and exits with a device failure.

The original reproduction also reports growing live-object counts for:

- `ID3D12CommandAllocator`
- `ID3D12GraphicsCommandList`
- `ID3D12Resource`

## Source analysis

- D3D12 `ConstantBufferPool::kPageSize` is 4 MiB.
- A pool is owned by every `CommandBufferImpl`.
- `CommandQueueImpl` retains retired command buffers in `m_commandBuffersPool` until queue shutdown.
- `CommandBufferImpl::reset()` calls `m_constantBufferPool.reset()`.
- `ConstantBufferPool::reset()` preserves `m_pages`; it only resets `m_currentPage` and `m_currentOffset`.

For the reproduction, roughly 2,000 pages x 4 MiB gives a theoretical high-water allocation of about 7.8 GiB for constant-buffer pages alone, before accounting for the command allocators, command lists, descriptors, tensors, and other resources.

The generic binding error does not identify the failed allocation. `getBindingData()` can propagate failures from both descriptor allocation and `ConstantBufferPool::allocate()`/buffer creation, so the current log message is insufficient to distinguish them.

## Cross-backend observation

Vulkan currently has the same 4 MiB per-command-buffer pool and retains its pages on reset. The reproduction completed three frames on the tested Linux/Vulkan machine, but one frame (3,755 submissions in the original sample) reached a maximum RSS of 9,298,704 KiB (~8.87 GiB).

D3D11 uses 64 KiB pages. CUDA uses 16 KiB pages and returns its page allocations during reset.

## A/B validation

As a diagnostic experiment, changing only the D3D12 page size from 4 MiB to 64 KiB made the original Windows reproduction complete ten frames without binding errors. The baseline failed during frame 2 on the same machine/build.

This demonstrates that constant-buffer page pressure is causal, but a global one-line page-size change may not be the complete production design. Possible fixes include:

- use a small initial page and grow only when a command buffer actually records more uniform data;
- trim large/excess pages when command buffers retire;
- use a queue/device-level fence-aware upload arena;
- bound the number of cached/in-flight command buffers;
- propagate the actual allocation error instead of logging only `Failed to get binding data`.

The Vulkan pool should be audited/fixed alongside D3D12.

## Tested configuration

- SlangPy: `2c8afea2bbdcaf8c87db0797a2504c2fb15fe866`
- slang-rhi: `1a9768741246b1f356116d5dfa305933034ce049`
- OS: Windows 10 Pro 22H2, build 19045
- GPU: NVIDIA RTX 3500 Ada Generation Laptop GPU
- NVIDIA driver: 582.16 (`32.0.15.8216`)
- Python: 3.12.10

Related SlangPy issue: https://github.com/shader-slang/slangpy/issues/665

kaizhangNV:
## Summary

- Replace the private 4 MiB D3D12/Vulkan constant-buffer page in every command buffer with a queue-owned shared `StagingHeap`.
- Start shared constant-buffer pages at 64 KiB and grow geometrically up to 4 MiB only when recorded uniform-data demand requires it.
- Keep suballocation handles on each command buffer until its submission fence retires, so ranges cannot be reused while the GPU may still read them.
- Release empty grown pages immediately and retain at most one warm 64 KiB page for reuse.
- Add native allocator and D3D12/Vulkan regressions covering page growth, trimming, fence-scoped lifetime, page reuse, and GPU result validation.

## Validation

- GCC Debug library/test build on Linux.
- Focused Vulkan regressions: 2 test cases and 1,343 assertions passed.
- All Vulkan staging-heap tests: 11 test cases and 113 assertions passed.
- Broader Vulkan suite with environment-dependent `surface-*` tests excluded: 864 test cases and 75,138,985 assertions passed.
- MSVC Debug `slang-rhi-tests` build on Windows 10 / D3D12 (`192.168.0.96`).
- D3D12 page-growth test: 26 assertions passed.
- D3D12 fence/retirement integration test: 1,318 assertions passed.
- `pre-commit run --all-files` passed.

Fixes #844

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1067,6 +1067,7 @@ if(SLANG_RHI_BUILD_TESTS AND NOT SLANG_RHI_BUILD_SHARED AND NOT EMSCRIPTEN)
         tests/test-caching-allocator.cpp
         tests/test-command-encoder.cpp
         tests/test-compilation-report.cpp
+        tests/test-constant-buffer-pool.cpp
         tests/test-cmd-clear-buffer.cpp
         tests/test-cmd-clear-texture.cpp
         tests/test-cmd-copy-buffer.cpp
--- a/src/command-buffer.h
+++ b/src/command-buffer.h
@@ -11,6 +11,7 @@
 #include "reference.h"
 #include "command-list.h"
 #include "device-child.h"
+#include "staging-heap.h"
 
 #include "rhi-shared-fwd.h"
 
@@ -52,6 +53,7 @@ public:
 
 public:
     QueueType m_type;
+    StagingHeap m_constantBufferHeap;
 };
 
 class RenderPassEncoder : public IRenderPassEncoder
--- a/src/d3d12/d3d12-command.cpp
+++ b/src/d3d12/d3d12-command.cpp
@@ -1919,7 +1919,8 @@ Result CommandQueueImpl::init(uint32_t queueIndex)
         CreateEventEx(nullptr, nullptr, CREATE_EVENT_INITIAL_SET | CREATE_EVENT_MANUAL_RESET, EVENT_ALL_ACCESS);
 
     StagingHeap::Config constantBufferHeapConfig;
-    constantBufferHeapConfig.pageSize = 4 * 1024 * 1024;
+    constantBufferHeapConfig.pageSize = 64 * 1024;
+    constantBufferHeapConfig.maxPageSize = 4 * 1024 * 1024;
     constantBufferHeapConfig.memoryType = MemoryType::Upload;
     constantBufferHeapConfig.usage = BufferUsage::ConstantBuffer;
     constantBufferHeapConfig.defaultState = ResourceState::ConstantBuffer;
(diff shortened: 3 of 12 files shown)


skallweitNV (commented):
I've arrived at another solution to this. See https://github.com/shader-slang/slang-rhi/pull/849 

Please review and report back if this solves the issue.

kaizhangNV:
Close this PR as the alternative is merged.

PR #856, opened 2026-09-12
Add application data to shader table records

kaizhangNV:
## Motivation

A ray-tracing shader can declare per-record data such as:

```slang
struct MaterialRecord
{
    float3 albedo;
}
```

The application must place one `MaterialRecord` after each native miss, hit-group, or callable shader identifier. `ShaderTableDesc` currently exposes only `ShaderRecordOverwrite`, whose fixed eight-byte payload is intended for patching native record bytes and cannot represent an arbitrary record. As a result, a client cannot construct the local-root/SBT-record data required by the structural ray-tracing API.

## Proposed solution

Add `ShaderRecordData`, a pointer-and-size view whose bytes are copied immediately after the native record header. `ShaderTableDesc` accepts one optional array for each miss, hit-group, and callable section. Shader-table creation owns a copy because D3D12, Vulkan, and OptiX materialize pipeline-specific tables lazily.

Each backend derives a common stride from the largest record in its section, applies its native alignment and maximum-stride rules, writes the native identifier first, then the application bytes, and finally applies the legacy overwrite so existing overwrite precedence remains unchanged.

## Change summary

- Extend the public descriptor with miss, hit-group, and callable record-data arrays.
- Store owned record bytes in the shared shader-table implementation.
- Validate pointer, arithmetic, alignment, and backend stride limits before taking ownership.
- Pack the data for D3D12, Vulkan, CUDA/OptiX, and their existing dispatch descriptors.
- Add shared unit coverage plus Vulkan validation coverage and exercise record data in the ray-tracing tests.

## Concepts and vocabulary

- **Native record header**: the D3D12 shader identifier, Vulkan shader-group handle, or OptiX SBT header at the beginning of a record.
- **Application record data**: opaque bytes following that header, interpreted by the target shader ABI.
- **Section stride**: the common aligned record size used by every record in one miss, hit-group, or callable table section.

## Process report

Consider two hit records using the same shader but different material values. The caller supplies two `ShaderRecordData` elements. `ShaderTable::ShaderTable` copies each byte range into an `OwnedRecord`; this is the source of truth retained until a backend-specific table is requested. The backend obtains the section size from `getMaxRecordSize`, aligns that value using its native limits, writes every record at that stride, and copies each record's bytes after the native header. This preserves the native requirement that one section has a uniform stride while allowing records to contain different amounts of data.

The previous backend code calculated section sizes in `uint32_t`. A large application record could therefore wrap before allocation or before a native stride field was populated. The new checked `Size` helpers validate addition, multiplication, and alignment first. `DeviceImpl::createShaderTable` supplies the actual backend header, alignment, and maximum stride; invalid descriptors fail before `OwnedRecord` attempts to copy their pointers. Narrowing occurs only after validation proves the value fits the native field.

`ShaderRecordOverwrite` remains separate from the owned application bytes. Materialization writes the identifier, then the application range, then the overwrite. That ordering preserves the prior behavior for clients that intentionally patch an absolute native-record offset.

Focused Linux validation:

- Built the parent Slang `slang-rhi` target.
- Built the standalone `slang-rhi-tests` target with eight-job/CPU limits.
- Ran `shader-table-*`: 6 test cases and 613 assertions passed, including the Vulkan validation path; CUDA was unavailable locally.

All CI build jobs pass across Windows, Linux, macOS, and Emscripten.


kaizhangNV:
The D3D12 structural Record materialization that consumes this ShaderRecordData contract is now available as the stacked draft #875.

PR #864, opened 2026-09-14
Fix OptiX callable entry-point lookup

### LINKED ISSUE #857: OptiX callable program groups use the wrong Slang symbol prefix
## Problem

Slang emits an OptiX callable entry point named `__direct_callable__<entryPoint>` in `source/slang/slang-emit-cuda.cpp`, matching the OptiX direct-callable naming contract. The slang-rhi OptiX backend instead asks `optixProgramGroupCreate` for `__callable__<entryPoint>` in `src/cuda/optix-api-impl.cpp`.

For example, a Slang callable named `callableWriteAttribute` is emitted as:

```cpp
extern "C" __device__ void __direct_callable__callableWriteAttribute(...)
```

but slang-rhi sets `entryFunctionNameDC` to `__callable__callableWriteAttribute`. OptiX therefore cannot create the callable program group.

This is independent of the structural ray-tracing API: the existing legacy callable runtime test is currently excluded from CUDA/OptiX, so the mismatch is not covered.

## Expected fix

Use the `__direct_callable__` prefix for `entryFunctionNameDC`, then enable OptiX coverage for the existing callable runtime test. Keep continuation callables out of scope.

kaizhangNV:
## Motivation

Consider a Slang callable entry point named `callableWriteAttribute`. Slang emits the OptiX symbol
`__direct_callable__callableWriteAttribute`, but slang-rhi asked OptiX to load
`__callable__callableWriteAttribute`. The requested symbol does not exist, so
`optixProgramGroupCreate` fails before the callable can run. This is tracked by #857.

## Proposed solution

Use Slang's `__direct_callable__` entry-point prefix when slang-rhi creates an OptiX direct-callable
program group. Enable CUDA on the existing callable runtime test so the compiler and backend naming
contracts stay synchronized. The fetched Slang release is updated to 2026.17.1 because that is the
first release containing CUDA `CallShader` support needed by the test.

## Change summary

- Request the emitted `__direct_callable__<source name>` symbol from OptiX.
- Run the existing `ray-tracing-intrinsics-call-shader` test on CUDA.
- Remove the shader's CUDA exclusion.
- Update the fetched Slang release and its official archive hashes to 2026.17.1.

## Concepts and vocabulary

- **Direct callable**: an OptiX callable program invoked with `optixDirectCall`; Slang gives these
  entry points the `__direct_callable__` native prefix.
- **Source entry-point name**: the unmangled Slang name stored by the ray-tracing pipeline descriptor.

## Process report

The pipeline descriptor reaches `OptixApiImpl::createRayTracingPipeline` with the correct callable
stage and source entry-point name. The backend then owns the target-specific native prefix before it
passes `entryFunctionNameDC` to `optixProgramGroupCreate`. No alternate AST, IR, or reflection shape
is involved: the only broken invariant was that this backend-local prefix disagreed with
`CUDASourceEmitter::generateEntryPointNameImpl()`. Updating that one mapping fixes the producer and
consumer contract without adding a fallback or a second lookup path.

Validation used a capped Release build against Slang 2026.17.1 and ran:

```text
slang-rhi-tests -tc=ray-tracing-intrinsics-call-shader.cuda \
    -select-devices=cuda -require-devices=cuda
```

The test passed on OptiX and verified the callable's returned payload values.

Closes #857.


skallweitNV (approved):
LGTM

PR #865, opened 2026-09-14
Configure OptiX callable stack sizes

### LINKED ISSUE #858: OptiX pipelines do not configure direct-callable stack sizes
## Problem

The OptiX backend creates and links ray-tracing pipelines in `src/cuda/optix-api-impl.cpp`, but it never computes program-group stack requirements or calls `optixPipelineSetStackSize` (or `optixPipelineSetStackSizeFromCallDepths`).

OptiX documents that its internal default is correct only when there are no direct-callable programs or motion transforms. Once a pipeline contains a Slang `[shader("callable")]` entry invoked through `CallShader`, the backend cannot rely on that default.

This is separate from #857. Correcting the emitted entry-point prefix lets OptiX create the callable program group, but does not establish a valid runtime stack contract for non-trivial direct callables.

## Reproduction shape

A ray-generation, miss, or closest-hit shader calls a direct callable whose body has a non-zero direct stack requirement. slang-rhi links the callable program group but launches the pipeline without configuring `directCallableStackSizeFromState`. An intersection or any-hit caller has the analogous `directCallableStackSizeFromTraversal` requirement.

The existing callable runtime test is a leaf callable and may happen to use no measurable direct stack, so it is not sufficient coverage for this defect.

## Expected direction

- Accumulate stack sizes for every linked OptiX program group.
- Provide or derive maximum direct-call depth separately for calls from traversal (`intersection`/`anyhit`) and state (`raygeneration`/`miss`/`closesthit`).
- Configure the pipeline stack after linking.
- Add a callable runtime test whose generated direct-callable body requires stack.

The public `RayTracingPipelineDesc` currently exposes only `maxRecursion`; it has no direct-callable-depth or traversable-graph-depth contract, so the API/default policy needs to be decided as part of the fix.

kaizhangNV:
## Motivation

Consider this call tree:

```slang
[shader("raygeneration")]
void rayGen()
{
    CallShader(0, payload);
}

[shader("callable")]
void callableA(inout Payload payload)
{
    CallShader(1, payload);
}
```

The ray-generation shader activates callable A, which activates callable B. OptiX requires stack space for two simultaneously active direct-callable frames, but slang-rhi linked the program groups without computing or setting any pipeline stack sizes. OptiX's internal default is only valid when a pipeline has no direct callables or motion transforms. This is tracked by #858.

## Proposed solution

Add an ABI-safe `OptixRayTracingPipelineDesc` chained through `RayTracingPipelineDesc::next`. It gives the application explicit maximum direct-callable depths for call trees originating from state shaders and traversal shaders. A state depth of one preserves existing leaf-callable behavior; nested or recursive call trees opt into their actual finite bound.

After linking, the OptiX backend queries every program group through the selected SDK wrapper, accumulates the requirements locally, and uses the official stack-size formula to compute the state, traversal, and continuation sizes. It then configures the pipeline for the existing backend IAS-to-GAS traversable graph depth of two.

## Change summary

- Add `OptixRayTracingPipelineDesc` with documented state/traversal depth semantics and a concrete depth-two example.
- Query each program-group stack size through its version-local OptiX wrapper and accumulate the maxima.
- Compute split direct-callable sizes with `optixUtilComputeStackSizesDCSplit` and apply them with `optixPipelineSetStackSize`.
- Add a portable nested `CallShader` runtime test on D3D12, Vulkan, and CUDA; CUDA chains a state-call depth of two.
- Keep a value live across the nested call so the regression exercises a non-trivial direct-callable frame.

## Concepts and vocabulary

- **State shader**: OptiX's ray-generation, miss, or closest-hit execution state.
- **Traversal shader**: OptiX's intersection or any-hit execution state.
- **Direct-call depth**: the maximum number of simultaneously active direct-callable frames in one call tree. Ray generation -> callable A -> callable B has depth two.
- **Traversable graph depth**: the acceleration-structure graph depth, independent of shader-call depth. slang-rhi currently accepts IAS -> GAS, whose depth is two.

## Process report

The valid native input to stack calculation is the vector of `OptixProgramGroup` objects already created by `ContextImpl::createPipeline`. The implementation keeps that canonical representation: it does not reconstruct an AST, IR call graph, or source-level entry-point map. Each program group is passed directly to `optixProgramGroupGetStackSize` together with the linked pipeline, which also lets OptiX account for external functions. The maxima are accumulated in a translation-unit-local helper rather than `optixUtilAccumulateStackSizes`: the OptiX 8.0 inline helper has external linkage and can be coalesced across the 8.0, 8.1, and 9.0 wrapper translation units in debug builds. That made a newer wrapper dispatch through the uninitialized 8.0 function table. The local helper preserves the SDK accumulation formula while binding the query to the selected SDK version.

`OptixStackSizes` exposes one conservative maximum direct-callable requirement rather than reachability split by caller category. The backend therefore uses that same safe upper bound for each enabled category, while `OptixRayTracingPipelineDesc` supplies the separate finite call-tree depths. Deriving those depths from shader-table count would be incorrect because a dynamic callable index can form cycles or data-dependent recursion, and current Slang reflection exposes no finite call-graph bound.

The pipeline takes ownership of its native modules and program groups before the new stack queries. If a query or stack configuration fails, normal `RefPtr` destruction releases every partially configured OptiX object; no fallback or masked error path is introduced. Continuation callables remain unsupported and use depth zero. Motion transforms remain unsupported, so `maxTraversableGraphDepth` stays two, matching the existing single-level-instancing compile flag.

Validation used the fetched Slang 2026.17.1 dependency and a capped Clang Debug build (eight jobs maximum):

- Full Linux x86_64 CUDA/Vulkan/WebGPU device sweep: 884/884 test cases passed.
- CUDA ray-tracing suite: 66/66 test cases passed under each of OptiX 8.0, 8.1, and 9.0.
- Both CI crash reproducers, `opacity-micromap-trace.cuda` and `ray-tracing-intrinsics-accept-hit-and-end-search.cuda`, pass in the local OptiX suites.
- Nested callable runtime passed on Vulkan.

Closes #858.


include/slang-rhi.h
```diff
@@ -2186,6 +2188,31 @@ struct RayTracingPipelineDesc
...
 };
 
+/// OptiX-specific stack limits for a ray-tracing pipeline.
+///
+/// Chain this structure through `RayTracingPipelineDesc::next` when an OptiX pipeline needs limits
+/// other than the defaults below. OptiX requires the maximum callable depth before it can allocate
+/// the pipeline stack. Slang reflection does not expose a finite bound for dynamic `CallShader`
+/// indices, so the application supplies that bound explicitly. Other backends ignore this
+/// descriptor.
+struct OptixRayTracingPipelineDesc
```
skallweitNV, line 2198: I assume the same max depth configuration is not available in D3D12, Vulkan and Metal? Otherwise this should probably go into the main pipeline descriptor.

kaizhangNV: Correct. The same configuration is not available on the other implemented ray-tracing backends. D3D12 and Vulkan expose a total pipeline stack size in bytes rather than the separate OptiX maximum direct-callable depths from state and traversal shaders. Their API-defined default stack calculations already include two callable stack contributions, which covers the depth-two test added here; slang-rhi currently relies on those defaults. Metal ray-tracing pipeline creation is not implemented yet. Since the split depth values map directly to optixUtilComputeStackSizesDCSplit and have no direct cross-backend representation, I kept them in the OptiX extension rather than adding fields that the other backends would ignore.

skallweitNV: that makes sense

PR #878, opened 2026-09-21
Expose CUDA GPU capability and Slang-selected NVRTC architectures

tdavidovicNV:
Applications need both the GPU's compute capability and the architectures accepted
by Slang's NVRTC to choose a CUDA target. Hardware capability names alone cannot
provide that intersection, especially when the GPU is newer than the compiler.

Expose the exact driver-reported capability in `DeviceInfo` and add
`getCUDACompilerInfo()` for the Slang-selected NVRTC's path, version and complete
architecture list. The query is lazy and cached; returned pointers remain valid
for the device lifetime. Missing architecture-query support is reported explicitly.
The application retains control over target selection.

Validated on Windows with an RTX 5090 and NVRTC 13.3: direct driver/compiler
comparison, missing-query and error handling, and non-CUDA behavior.


tdavidovicNV:
Closing this because the remaining work is moving to SlangPy. RHI already reports the GPU's CUDA capabilities through its existing capability set. SlangPy owns the Slang global session, so it can discover the selected NVRTC and combine its supported architectures with those GPU capabilities without adding another RHI metadata interface or DeviceInfo field. CUDA target selection and the temporary compiler workarounds will be implemented there.

PR #879, opened 2026-09-21
Allow callers to supplement persistent shader-cache keys

tdavidovicNV:
Slang can return the same entry-point hash for programs with different link-time
downstream options ([Slang #13197](https://github.com/shader-slang/slang/issues/13197)).
Reusing that hash as the persistent cache key can return code compiled with the
wrong options.

Add optional `ShaderProgramDesc::cacheKeySalt` so callers can include those settings
in the key. RHI copies the string, carries it into specialized programs, and uses
the supplemented key consistently for cache reads, writes and compilation reports.
Null or empty salt preserves existing keys. This lets callers work around the Slang
issue without changing the compiler hash API.

Validated cold/warm reuse, salt ownership, specialization inheritance and report-key
agreement on D3D12 and Vulkan.


tdavidovicNV:
Closing this in favor of keeping the temporary workaround for shader-slang/slang#13197 local to SlangPy's program-linking code. A public RHI cache-salt API adds a lasting interface and ownership contract for a compiler bug; the workaround should instead be isolated and easy to delete once our minimum Slang version contains the fix. The Slang issue remains open, and the independent CUDA metadata PR #878 is unaffected.

